PR #1, opened 2016-11-16
Setup Rspec as the unit testing framework

leenagupte:
We have chosen to use Rspec as the testing framework rather than Minitest (the default).


pmanrubia:
👍

PR #5, opened 2016-11-24
Preparations to use gov.uk style frontend

selfthinker:
This adds all the necessary frontend gems and prepares the CSS and layout to produce gov.uk styled pages.
After this has been merged, you can use simple HTML in views or more complex HTML copied and pasted from [Elements](http://govuk-elements.herokuapp.com/).

What still needs to be done:

* At some point we need to swap `govuk_template` for the `slimmer` gem. I tried to do that but was running into issues, so decided to go the simplest route first. Slimmer would be needed to a) automatically update all the assets and b) to get the gov.uk content in the header, footer, breadcrumbs, etc. and c) to use Gov.uk Components. All of that is not important in the beginning but will get more important when we plan to make this customer-facing.
* The general layout can be extended (e.g. with a title or a sidebar) but that should be done when and if it is needed. Look into [govuk_template](https://github.com/alphagov/govuk_template) to see all the possible hooks etc.

I tested everything in a test page which I removed again. It looked like this:
com-prototype-frontend

### DIFF
--- a/app/assets/stylesheets/base.scss
+++ b/app/assets/stylesheets/base.scss
@@ -1,0 +1,55 @@
+/* sensible default styles for main elements */
+
+main {
+
+  h1 {
+    @extend .heading-xlarge !optional;
+  }
+
+  h2 {
+    @extend .heading-large !optional;
+  }
+
+  h3 {
+    @extend .heading-medium !optional;
+  }
+
+  h4 {
+    @extend .heading-small !optional;
+  }
+
+  ul,
+  ol {
+    @extend .list !optional;
+  }
+
+  ul {
+    @extend .list-bullet !optional;
+  }
+
+  ol {
+    @extend .list-number !optional;
+  }
+
+  strong {
+    @extend .bold !optional;
+  }
+
+  /* forms */
+
+  input,
+  select,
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/base.scss
```diff
@@ -0,0 +1,47 @@
+main {
+
+  h1 {
+    @extend .heading-xlarge !optional;
```
selfthinker: This bug report might be related: https://github.com/sass/sass/issues/2177
I'll chase that up when I'm back from my holiday.

chao-xian: In the `Gemfile.lock` we are using sass 3.4.22 and it is the exact error message.

pmanrubia:
Thank you for your work on this @selfthinker & @chao-xian 
Happy to merge as this is a temporary step to move our alpha-version forward, and we have created a [Trello card in our backlog](https://trello.com/c/h0q7VuNs/456-convert-prototype-app-to-use-better-gov-uk-styling-configuration) to track the work related to upgrading to [smiller](https://github.com/alphagov/slimmer) - as mentioned above.

PR #6, opened 2016-11-28
Endpoint for Organisations and ContentItems per organisation

pmanrubia:
Covers [this story][1] that exercises the first components of our application. Organisations, Content Items, Controllers and all the testing stack.

## Includes:

1\. Exposes an **endpoint** at `/organisations/index` and return a JSON with the organisations and the number of items of each organisation

```
[{
  organisation_id: 8392378947298374,
  number_of_pages: 12.323
},{
  organisation_id: 8392378947298374,
  number_of_pages: 123
}]
```

2\. Add a **`rake task`** to import all the Content Items for an Organisation provided the Organisation slug

```bash
$ rails import:organisation[hm-revenue-customs]
```

## Notes

1. We temporarily rely on [HTTParty][5] to query SearchAPI; this is a temporary solution as we will use GDS  Adapters as a way to communicate with other applications. The task to upgrade to use GDS adapters is tracked in a [Trello Card][6]
2. It is not ideal the way we are mocking the HTTP requests, as we are not `webmok-ing` the `get` request only the library request. This is something that [will go away in a few weeks][6], so no need to work any further.
3. We are using a very simple mechanism to serialise to JSON the organisations . This is because most of this code will go away once we expose an API to other apps. We will to rely on [Api Serializers][3] or using libraries like [Grape][4]. No need to complicate things further right now.
4. As a result of the development of this PR, I have created a [Trello card][2] to automate Lint when running all tests in each commit so we prevent this situation from happening again. This way we don't need to write a final commit with all the fixes.

## Aditional notes

We are not doing the Heroku work as part of this story because there is some pending work related to use GDS Heroku account and I don't want this task to block this card in our Trello board.
I have created a [Trello card][7] to track the work to deploy to Heroku.

[1]: https://trello.com/c/JP2lwuod/69-as-a-content-designer-i-would-like-to-access-the-number-of-pages-i-that-belong-to-my-organisation
[2]: https://trello.com/c/O6wNOS9N/79-run-lint-validator-when-running-rspec-tests
[3]: https://github.com/rails-api/active_model_serializers
[4]: https://github.com/ruby-grape/grape
[5]: https://github.com/jnunemaker/httparty
[6]: https://trello.com/c/mn8JPHYR/78-use-gds-adapter-to-retrieve-content-items-for-an-organisation
[7]: https://trello.com/c/1Hc4Q5Np/81-deploy-to-the-common-heroku-account-of-gds


spec/controllers/organisations_controller_spec.rb
```diff
@@ -12,8 +12,5 @@
       get :index
       expect(response.content_type).to eq('application/json')
     end
-
-
   end
-
 end
```
leenagupte: This is an odd place to remove lines in this file.

db/schema.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  create_table "organisations", force: :cascade do |t|
+    t.string   "organisation_id"
+    t.datetime "created_at",      null: false
+    t.datetime "updated_at",      null: false
+  end
+
+  add_foreign_key "content_items", "organisations"
+end
```
leenagupte: I would move the creation of the schema file to commit 66da862d271577732a621318b33f4165eba73c14

Gemfile.lock
```diff
@@ -222,4 +225,4 @@ DEPENDENCIES
   web-console
 
 BUNDLED WITH
-   1.13.1
+   1.13.6
```
leenagupte: Why has the version changed? Is it a requirement of the shoulda-matchers gem?

Gemfile.lock
```diff
@@ -230,5 +234,8 @@ DEPENDENCIES
   uglifier (>= 1.3.0)
   web-console
 
+RUBY VERSION
+   ruby 2.3.1p112
+
```
leenagupte: Do we need this? Is this something that should have been added when the app was created?

leenagupte:
In commit aa0254427c508ef0bbd41cf85829754442056cbc can you change the commit message to `filters all organisations by ContentItems`.

`filters out` means to remove the content items.

spec/models/importers/organisation_spec.rb
```diff
@@ -63,4 +63,11 @@
     content_ids = organisation.content_items.pluck(:content_id)
     expect(content_ids).to eq(%w(number-1 number-2))
   end
+
+  it 'raises an exception with an organisation that does not exist' do
```
leenagupte: Should change this to `raises an exception with an organisation that does not have any content items`

spec/models/importers/organisation_spec.rb
```diff
@@ -1,28 +1,19 @@
...
   context 'Organisation' do
     it 'imports an organisation with the provided slug' do
-      response = double(body: {
-        results: [
-          {
-            some_field: 'some_value',
-          },
-        ]}.to_json)
-      allow(HTTParty).to receive(:get).and_return(response)
+      expect(HTTParty).to receive(:get).and_return(one_content_item_response)
```
leenagupte: Why did this change from `allow` to `expect`?

app/models/importers/organisation.rb
```diff
@@ -11,19 +12,36 @@ def initialize(slug, batch: 10, start: 0)
...
     organisation = ::Organisation.create!(slug: slug)
     loop do
-      response = HTTParty.get "https://www.gov.uk/api/search.json?filter_organisations=#{slug}&count=#{batch}&fields=content_id&start=#{start}"
-      results = JSON.parse(response.body).fetch('results')
-
-      raise "No result for slug" if results.empty? && organisation.content_items.empty?
-
-      results.each do |result|
-        attributes = result.slice('content_id')
+      result = search_content_items_for_organisation
```
leenagupte: I thought it wasn't very Rubyist to use verbs in method names? Should this be `content_items_for_organisation`?

app/models/importers/organisation.rb
```diff
@@ -11,19 +12,36 @@ def initialize(slug, batch: 10, start: 0)
...
+  end
+
+  private
+
+  def last_page?(results)
+    results.length < batch
+  end
+
+  def next_page!
+    self.start += batch
```
leenagupte: Doesn't `self` mean that you are using the class variable? Isn't `start` an instance variable now? Shouldn't `start` be passed to `next_page!` instead? Or, use `@start`?

app/models/importers/organisation.rb
```diff
@@ -25,7 +24,7 @@ def run
     raise 'No result for slug' if organisation.content_items.empty?
   end
 
-  private
+private
```
leenagupte: You could fix this up to the previous commit!

spec/controllers/organisations_controller_spec.rb
```diff
@@ -1,14 +1,13 @@
...
-  describe "GET #index" do
-    it "returns http success" do
+  describe 'GET #index' do
+    it 'returns http success' do
       get :index
       expect(response).to have_http_status(:success)
     end
 
-    it "returns JSON content type" do
+    it 'returns JSON as content type' do
```
leenagupte: I think this reads better as `returns a JSON content type`

spec/lib/tasks/import_spec.rb
```diff
@@ -20,7 +20,7 @@
       Rake::Task['import:organisation'].invoke('a_slug')
     end
 
-    it 'raise an error if a slug is not provided' do
+    it 'raise an error if a slug parameter is not present' do
       expect { Rake::Task['import:organisation'].invoke }.to raise_error('Missing slug parameter')
```
leenagupte: Nitpick: should be `raises`.

leenagupte:
@pmanrubia As a whole, the code looks good. I'm missing the story though. 

In commit 5979c3a, an empty `run` method is created to test the new rake task, but this `run` method is ignored for three commits until aa0254427c508ef0bbd41cf85829754442056cbc. 

I thought the plan was to only add something right before it's needed?

Gemfile.lock
pmanrubia: I would get rid of it, but it was added in https://github.com/alphagov/com-prototype/pull/2 in https://github.com/alphagov/com-prototype/pull/2/commits/6e0691a054991bad257db1794f86cd15620787a8. 
Not a big deal, though. It might be easier for the first sprints to rely only on `.ruby-version`

Gemfile.lock
pmanrubia: I had a more modern Bundler version. This number never goes backwards.

spec/controllers/organisations_controller_spec.rb
pmanrubia: Yes, during the live-coding-sessions we didn't put that much attention to spaces and blank lines. This is why I have [fixed all lint errors in the latest commit](https://github.com/alphagov/com-prototype/pull/6/commits/db720509f1e2011fdb17b434536cacffa19d73ce), and I have [create a Trello card](https://trello.com/c/O6wNOS9N/79-run-lint-validator-when-running-rspec-tests) to prevent this situation from happening again.

db/schema.rb
pmanrubia: thanks!, fixed in https://github.com/alphagov/com-prototype/pull/7/commits/58ecd31901d1fb9f8670011f08a287731acb2bd9

pmanrubia:
> In commit aa02544 can you change the commit message to filters all organisations by ContentItems. > filters out means to remove the content items.

Done in #7

spec/models/importers/organisation_spec.rb
pmanrubia: We are not mocking the get request, just the library. So, I just wanted to be sure that we are actually using the mock. As commented in the description the library (HTTParty) along with the way we are testing it will go away in a few weeks, so feels good enough for the time being.

app/models/importers/organisation.rb
pmanrubia: My understanding is that `nouns` are good for class names, and `verbs` for methods. The opposite, using verbs for class names is the antipattern.
Have a look at a controller or at any ActiveRecord API. Most of them are verbs.

spec/models/importers/organisation_spec.rb
pmanrubia: I have updated it. Thank you

app/models/importers/organisation.rb
pmanrubia: I just fixed all lint errors in one commit. I prefer not to put time into fixing up this line. I hope this is ok with you.

spec/controllers/organisations_controller_spec.rb
pmanrubia: I have updated the copy of the test. Thank you

spec/lib/tasks/import_spec.rb
pmanrubia: I have updated the copy of the test, thank you

pmanrubia:
Thank you for your review, I have addressed your comments in #7 

Regarding the blocks of commits... I have move 5979c3a down 3 commits to be close to aa02544. I hope this is more clear now.

Closing this PR, as this has been superseeded by #7

Gemfile.lock
leenagupte: Ok.

app/models/importers/organisation.rb
leenagupte: I shall point @ikennaokpala to this comment :)

app/models/importers/organisation.rb
leenagupte: Thanks for the explanation. I'm never quite sure when `self` refers to a class variable or the existing instance variable.

app/models/importers/organisation.rb
leenagupte: That's fine.

PR #9, opened 2016-11-29
Show first (25) Content Items for an Organisation

ikennaokpala:
[Trello card](https://trello.com/c/XwH3123y/60-story-show-first-25-content-items-for-an-organisation)

## Description 

This PR covers this [card](https://trello.com/c/XwH3123y/60-story-show-first-25-content-items-for-an-organisation), with the intention to fetch an organisation's content items (i.e given the organisation’s id) and renders the content items in a HTML table.

1. Defines `/organisations/organisation_id/content_items` endpoint
2. Gets the first 25 content items belonging to an organisation.
3. Renders the content items ids in a html table.

### NB: 

The following gems have been added as required.

- rails-controller-testing: Providing methods/matchers like `assigns`, `render_template ` etc
- capybara: Providing matchers like have_selector that help inspect the rendered page

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,6 +42,7 @@ group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platform: :mri
   gem 'rspec-rails', '~>3.5'
+  gem 'rails-controller-testing', '~>1.0.1'
   gem 'shoulda-matchers'
   gem 'factory_girl_rails'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,7 @@ group :development, :test do
   gem 'byebug', platform: :mri
   gem 'rspec-rails', '~>3.5'
   gem 'rails-controller-testing', '~>1.0.1'
+  gem 'capybara'
   gem 'shoulda-matchers'
   gem 'factory_girl_rails'
 end
--- a/app/views/content_items/index.html.erb
+++ b/app/views/content_items/index.html.erb
@@ -1,0 +1,15 @@
+<h1>List Content Items</h1>
+<table>
+  <thead>
+    <tr>
+      <th>Content Ids</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @content_items.each do |content_item| %>
+      <tr>
+        <td><%= content_item.content_id %></td>
+      </tr>
+    <% end %>
+  </tbody>
+</table>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,7 @@
 Rails.application.routes.draw do
   root to: 'organisations#index'
 
-  resources :organisations, only: %w(index)
+  resources :organisations, only: %w(index) do
+    resources :content_items, only: %w(index)
+  end
 end
--- a/spec/controllers/content_items_controller_spec.rb
+++ b/spec/controllers/content_items_controller_spec.rb
@@ -2,9 +2,18 @@ require 'rails_helper'
 
 RSpec.describe ContentItemsController, type: :controller do
   describe "GET #index" do
+    let(:organisation) { FactoryGirl.create(:organisation_with_content_items) }
+
+    before do
+      get :index, params: { organisation_id: organisation }
+    end
+
     it "returns http success" do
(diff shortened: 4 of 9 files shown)


spec/controllers/content_items_controller_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    before do
+      get :index, params: { organisation_id: organisation }
+    end
+
+    it "returns http success" do
+      expect(response).to have_http_status(:success)
+    end
+
+    it "assigns content_items" do
```
pmanrubia: Nice controller testing.

I would ensure that the correct list of content items are assigned as an instance variable. We would suggest going with something like this.

```ruby
    it "assigns content_items" do
      content_items = FactoryGirl.build_list(:content_items, 1)
      allow(organisation).to receive(:content_items).and_return(content_items)

      expect(assigns(:content_items)).to eq(:some_content_items)
    end
```

Gemfile
```diff
@@ -42,6 +42,8 @@ group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platform: :mri
   gem 'rspec-rails', '~>3.5'
+  gem 'rails-controller-testing'
```
pmanrubia: Ikenna, could you please update the commit message explaining why this gem was added?
Would be enough with the error message we get when running the tests without it. In the long run it is very helpful to understand why we have included every gem that is not a default in Rails.

```bash
  1) ContentItemsController GET #index renders the :index template
     Failure/Error: expect(subject).to render_template(:index)

     NoMethodError:
       assert_template has been extracted to a gem. To continue using it,
               add `gem 'rails-controller-testing'` to your Gemfile.
```

config/routes.rb
```diff
@@ -1,5 +1,7 @@
 Rails.application.routes.draw do
-  root to: 'organisations#index'
+  root :to => 'organisations#index'
```
ikennaokpala: Ahh this must have creeped in when I rebased against previous changes.. will fix it.

app/views/content_items/index.html.erb
```diff
@@ -0,0 +1,15 @@
+<h1>List Content Items</h1>
+<table>
+  <thead>
+    <tr>
+      <th>Content Ids</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @content_items.each do |content_item| %>
```
ikennaokpala: Either way is fine.. I have just setup a basic HTML table..

spec/views/content_items/index.html.erb_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
+
+RSpec.describe "content_items/index.html.erb", type: :view do
+  let(:organisation) { FactoryGirl.create(:organisation_with_content_items) }
```
ikennaokpala: Make sense.. and more explicit

spec/controllers/content_items_controller_spec.rb
ikennaokpala: I can see your point for stubbing this out this way.. in the organisation's factory girl definition 

we already cater for this.. 

```
let(:organisation) { FactoryGirl.create(:organisation_with_content_items, content_items_count: 1) }
```

From here the only difference is that create would return a saved instance

pmanrubia: Fair enough @ikennaokpala, I am fine with it as well

pmanrubia:
@ikennaokpala It might be a better approach to stick to your first commit:

```ruby
    let(:organisation) { FactoryGirl.create(:organisation_with_content_items) }

    it "assigns list of content items" do
      expect(assigns(:content_items)).to eq(organisation.content_items)
    end
```

But, I am not sure if you were using the factory `:organisation_with_content_items` or `organisation`. I think my comment came because of this reason, but not sure abou it

ikennaokpala:
@pmanrubia it was your observation that led be think / review / remember the organisation's factory girl definition

pmanrubia:
Good work @ikennaokpala!

ikennaokpala:
@pmanrubia cheers for reviewing and point out some areas for thought..

PR #10, opened 2016-11-29
Add rake tasks to deploy to Heroku

pmanrubia:
Adds three rake tasks:

```bash
$ rake -T | grep heroku                                                                                                                                      
rake heroku:db:migrate  # Run pending migrations in Heroku
rake heroku:deploy         # Deploys the application to Heroku
rake heroku:prepare       # Configure the environment to do Heroku deployments
```

## Note

Deploying to Heroku is the temporary solution we have chosen while
a more robust solution is set up for our app.

I have created an application in Heroku: `content-performance-manager` 
and I have given permissions to all developers. 

In another story, we will reuse the GDS heroku account.

### DIFF
--- a/lib/tasks/heroku.rake
+++ b/lib/tasks/heroku.rake
@@ -3,4 +3,9 @@ namespace :heroku do
   task :prepare do |_task|
     Kernel.system 'git remote add heroku https://git.heroku.com/content-performance-manager.git'
   end
+
+  desc 'Deploys the application to Heroku'
+  task :deploy do |_task|
+    Kernel.system 'git push heroku master'
+  end
 end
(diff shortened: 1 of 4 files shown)


pmanrubia:
@chao-xian this is a good point. This is the first sprint, so I just aimed for having some place to deploy our stories and prove that they are done. 
I am very happy for the team to play another story to set up the staging environment.

chao-xian:
@pmanrubia I guess we don't need those yet, so we can just have a story in the icebox for later down the line when we actually do. I'm happy to approve.

pmanrubia:
Thank you for your review @chao-xian.
I have force pushed in preparation for merging.

PR #11, opened 2016-11-30
Run the database migration in same rake task as the deployment to Heroku

pmanrubia:
It feels more convenient for a developer not to have to run the 
database migration each time he needs to do a deployment.

For this reason, I am including the migration task within the deployment
task.

As a side note, I had to add the `Bundler.with_clean_env` to run the 
migrations because Heroku was picking up the wrong version of ruby.
See suggested solution in StackOverflow[1] and explanation in:

[1]: http://stackoverflow.com/questions/12995271/rails-server-error-ruby-version-is-1-8-7-but-your-gemfile-specified-1-9-3

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,3 @@
-### Important
-
-The name of this repo is temporary, and will change in a few weeks.
-
-
-
 ### App development
 
 * [GOVUK-LINT-RUBY](doc/govuk-lint.md)
--- a/README.md
+++ b/README.md
@@ -1,3 +1,18 @@
+
+### Deployment to Heroku
+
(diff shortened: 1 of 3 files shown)


leenagupte:
How do heroku database migrations work if you (I mean me) run the app in the Development VM, but deploy to heroku from my base mac?

Do I need to do anything extra? Would it be better if I installed the heroku client in my VM?

pmanrubia:
The migration process runs in Heroku, so the quick answer is that it is not relevant from where you are running the command.

pmanrubia:
@chao-xian I have updated the link. Thank you 👍 
I am force-pushing now in preparation for merging.

PR #13, opened 2016-11-30
Add last-update-time to the content items of an organisation

ikennaokpala:
[Trello card](https://trello.com/c/7swQPTrR/63-story-add-last-update-time-to-the-content-items-of-an-organisation)

## Description 
For every content item, content designers want to see the last time a
content item was updated for an organisation.

Here a new column/attribute (i.e `last_updated_at`) has been added to the content_item(s).

The `last_updated_at` attribute’s value is obtained via from the publishing api [1] via
the gds-api-adapters gem.

The gds-api-adapters gem provides access to the publishing_api endpoints

For obtaining the last time a piece of content_item was updated, this endpoint [1] is 
queried with the content_id supplied.

All organisation importer tests have been affected and a before hook introduced to
stub calls to get_content.

***NB:***

Kindly take note that publishing api can only be access via the dev VM [2]. 

[1] /v2/content/:content_id
[2] http://publishing-api.dev.gov.uk/v2/content/:content_id

PR #14, opened 2016-12-01
Show URL for a departments content item

ikennaokpala:
Related #13 

[Trello card](https://trello.com/c/SgGrXnx9/71-collect-url-for-a-departments-content)

## Description 

For each department’s content item, content designers want to visit the url of a
content item.

The link column/attribute has been added to the content_item(s).
This attribute’s value is obtained via this request to the search api.

A new method (i.e url) has been introduced that concatenates https://gov.uk and
the content item’s path.

This is then used to populate a Content URL column.

### DIFF
--- a/spec/models/content_item_spec.rb
+++ b/spec/models/content_item_spec.rb
@@ -2,4 +2,11 @@ require 'rails_helper'
 
 RSpec.describe ContentItem, type: :model do
   it { should belong_to(:organisation) }
+
+  describe '#url' do
+    it 'returns a url to a content item on gov.uk' do
+      content_item = build(:content_item, link: '/api/content/item/path/1')
+      expect(content_item.url).to eq('https://gov.uk/api/content/item/path/1')
+    end
+  end
 end
--- a/spec/models/importers/organisation_spec.rb
+++ b/spec/models/importers/organisation_spec.rb
@@ -31,28 +31,22 @@ RSpec.describe Importers::Organisation do
   end
 
   context 'Content Items' do
-    it 'imports all content items for the organisation' do
-      allow(HTTParty).to receive(:get).and_return(two_content_items_response)
-      Importers::Organisation.new('a-slug').run
-      organisation = Organisation.find_by(slug: 'a-slug')
+    before { allow(HTTParty).to receive(:get).and_return(two_content_items_response) }
+    let(:organisation) { Organisation.find_by(slug: 'a-slug') }
 
+    it 'imports all content items for the organisation' do
+      Importers::Organisation.new('a-slug').run
       expect(organisation.content_items.count).to eq(2)
     end
 
     it 'imports a `content_id` for every content item' do
-      allow(HTTParty).to receive(:get).and_return(two_content_items_response)
       Importers::Organisation.new('a-slug').run
-      organisation = Organisation.find_by(slug: 'a-slug')
-
       content_ids = organisation.content_items.pluck(:content_id)
       expect(content_ids).to eq(%w(content-id-1 content-id-2))
     end
 
     it 'imports a `link` for every content item' do
-      allow(HTTParty).to receive(:get).and_return(two_content_items_response)
       Importers::Organisation.new('a-slug').run
-      organisation = Organisation.find_by(slug: 'a-slug')
-
       links = organisation.content_items.pluck(:link)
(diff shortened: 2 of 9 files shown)


spec/models/importers/organisation_spec.rb
```diff
@@ -31,29 +31,22 @@
...
 
   context 'Content Items' do
-    it 'imports all content items for the organisation' do
+    before do
       allow(HTTParty).to receive(:get).and_return(two_content_items_response)
       Importers::Organisation.new('a-slug').run
-      organisation = Organisation.find_by(slug: 'a-slug')
-
-      expect(organisation.content_items.count).to eq(2)
+      @organisation = Organisation.find_by(slug: 'a-slug')
```
pmanrubia: 1. Duplication in the code that exercises the code is accepted if it increases readability. I think this is the case
2. It is also easier to add extra setup code for the test (if we use a before block we can't do it per test)

I would go with something like this:

```ruby
before { allow(HTTParty).to receive(:get).and_return(two_content_items_response) } 
let(:organisation) { organisation = Organisation.find_by(slug: 'a-slug') }

it 'imports a `content_id` for every content item' do
  Importers::Organisation.new('a-slug').run

  content_ids = organisation.content_items.pluck(:content_id)
  expect(content_ids).to eq(%w(content-id-1 content-id-2))
end
```

spec/views/content_items/index.html.erb_spec.rb
```diff
@@ -5,24 +5,31 @@
...
-    expect(rendered).to have_selector('table thead', text: 'Content Ids')
+    expect(rendered).to have_selector('table thead tr:first-child', text: 'Content Ids')
+    expect(rendered).to have_selector('table thead tr:nth(1)', text: 'Content URL')
   end
 
   it 'renders a row per Content Item' do
     render
 
-    expect(rendered).to have_selector('table tbody tr', count: 2)
+    expect(rendered).to have_selector('table tbody tr', count: 4)
```
pmanrubia: Why are we expecting to have 4 rows?

spec/views/content_items/index.html.erb_spec.rb
```diff
@@ -5,24 +5,31 @@
...
       render
 
-      expect(rendered).to have_selector('table tbody tr td', text: 'a-content-id')
+      expect(rendered).to have_selector('table tbody tr:first-child td', text: 'a-content-id')
+    end
+
+    it 'contains the content\'s url in the second column' do
+      render
+
+      expect(rendered).to have_selector('table tbody tr td:nth(1) a', text: 'https://gov.uk/content/item/path')
```
pmanrubia: This expectation is a bit hidden (we are testing a value that is hardcoded in the factory). It feels better to be explicit as per line 23.

spec/views/content_items/index.html.erb_spec.rb
ikennaokpala: This is meant be 4 rows.. becuase in the template I created an extra table row have.. I have changed it to only two table rows and 2 cells per row

spec/views/content_items/index.html.erb_spec.rb
ikennaokpala: Yeah, I see your point.. or we can generate like:

```ruby 
sequence(:link) { |index|  "/content/#{index}/path" }
```

This way it will be based on the index..  `build(:content_item)` `link` will be `/content/1/path`

spec/models/importers/organisation_spec.rb
ikennaokpala: I have updated the commits..

spec/models/content_item_spec.rb
```diff
@@ -2,4 +2,11 @@
 
 RSpec.describe ContentItem, type: :model do
   it { should belong_to(:organisation) }
+
+  describe '#url' do
+    it 'returns a url to a content item on gov.uk' do
+      content_item = build(:content_item)
```
pmanrubia: Sorry, Ikenna, would you mind being more explicit here?

ikennaokpala:
@pmanrubia thanks I have updated the link tests to reflect 

`content_item = build(:content_item, link: '/api/content/item/path/1')`

pmanrubia:
LGTM!

PR #18, opened 2016-12-07
Remove logging configuration from importing file

pmanrubia:
This PR intends to remove environment-dependent code

### Background

The rake task outputs content to STDOUT, but if we use it from `Rails console` we would like to use the defaults configured in `config/environments/development.rb`.
Likewise, if we are in  production, we would like to use the configuration specified in `config/environments/production.rb`.

To summarise it, we want to get rid of any configuration code related to the the current environment.

We can configure logging per environment in the files:
- config/environments/test.rb
- config/environments/development.rb
- config/environments/production.rb

## Developer note:

If the logging information is valuable for the script user we will store it in a more reliable place (Database, filesytem...) We have chosen a quick way to get feedback from the rake task
until we have a better picture about all the issues we have when parsing the data.

### DIFF
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -2,9 +2,9 @@ namespace :import do
   desc 'Import an organisation\'s details and content items'
   task :organisation, [:slug] => :environment do |_, args|
     raise 'Missing slug parameter' unless args.slug
-    Rails.logger = Logger.new(STDOUT)
-    Rails.logger.level = Logger::INFO
+    logger = Logger.new(STDOUT)
+    logger.level = Logger::INFO
 
-    Importers::Organisation.new(args.slug).run
+    Importers::Organisation.new(args.slug, logger: logger).run
   end
 end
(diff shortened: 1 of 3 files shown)


lib/tasks/import.rake
```diff
@@ -2,6 +2,8 @@ namespace :import do
   desc 'Import an organisation\'s details and content items'
   task :organisation, [:slug] => :environment do |_, args|
     raise 'Missing slug parameter' unless args.slug
+    Rails.logger = Logger.new(STDOUT)
```
chao-xian: This need to be wrapped in a:
```ruby
unless Rails.env.test?
...
end
```
as I'm getting log messages to the STDOUT when I run `rspec`.

pmanrubia:
Hi @chao-xian, I have updated the PR so we don't have any environment related code. Please, have a look at the commits and let me know if this is ok with you. Cheers

pmanrubia:
Closing the PR, as it is not needed after #38

PR #19, opened 2016-12-08
Improve the information displayed in the content items table

pmanrubia:
Include a few improvements for our first story:

1. Removes the column with the content item id
2. Uses the content item title for the link text
3. Open the Content Item link in a new Tab
4. Another minor refactoring

### Before
image

### After
image

### DIFF
--- a/app/views/content_items/_content_item.html.erb
+++ b/app/views/content_items/_content_item.html.erb
@@ -1,0 +1,4 @@
+<tr>
+  <td><%= link_to content_item.title, content_item.url %></td>
+  <td><%= time_ago_in_words(content_item.public_updated_at) %> ago</td>
+</tr>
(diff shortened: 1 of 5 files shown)


app/views/content_items/_content_item.html.erb
```diff
@@ -0,0 +1,4 @@
+<tr>
+  <td><%= link_to content_item.title, content_item.url, target: '_blank' %></td>
```
selfthinker: Opening links in a new window is a usability issue (and as I've just learned also a [security issue](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/)). If we're going to use it (and I don't see the need here), then it should be made clear (e.g. with a title "opens in new window").

pmanrubia: Ok, fair enough. I will revert it.

PR #20, opened 2016-12-08
Displays the slug of the Organisation

pmanrubia:
Although we would need to display the name of the organisation, it
feels good enough for the first sprint to display the slug.

### DIFF
--- a/app/views/content_items/index.html.erb
+++ b/app/views/content_items/index.html.erb
@@ -1,4 +1,4 @@
-<h1>List Content Items</h1>
+<h1><%= @organisation.name %></h1>
 <table>
   <thead>
     <tr>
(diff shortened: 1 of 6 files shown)


app/views/content_items/index.html.erb
```diff
@@ -1,4 +1,4 @@
-<h1>List Content Items</h1>
+<h1><%= @organisation.slug %></h1>
```
chao-xian: @pmanrubia Is it worth "cheating" for now and doing `.gsub('-', ' ').humanize` here, abstracting later to the model?

pmanrubia:
Hi @chao-xian, thank you for your review. I have addressed your comments.

PR #21, opened 2016-12-08
Handle ordering of content items by attribute in ascending or descending order

ikennaokpala:
[Trello card](https://trello.com/c/jGICn84K/65-story-list-the-pages-that-have-not-been-updated-in-order-bigger-gap-first)

## Description 

As a content designer, when I click on the last_update_at column I change the ordering of the column (ascending and descending).

When a user clicks on the last update table column the ordering of the column changes to
ascending or descending.

Here two links have been included to help with sorting the column by the last updated time.

We cater for the order and sort params that come through and are used to sort the content items by `public_updated_at` in ascending and descending order.

The ordering format is as follows `order('<attribute> (ASC|DESC)')`.

This has not been hard coded and can be reused by any other content item
attribute.

Following the conversations below.. we have agreed to park 6f501c34d3dd3505f8bbb134e069a8a90385a2dd to [story/handle-content-ordering-cntd](https://github.com/alphagov/content-performance-manager/commits/story/handle-content-ordering-cntd), to be revisited in the coming sprints.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -1,6 +1,6 @@
 class ContentItemsController < ApplicationController
   def index
     @organisation = Organisation.find(params[:organisation_id])
-    @content_items = @organisation.content_items.first(25)
+    @content_items = @organisation.content_items.order("#{params[:sort]} #{params[:order]}").first(25)
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,2 +1,9 @@
 module ApplicationHelper
+  def order_link(heading, content_item_attribute)
+    link_to heading, organisation_content_items_path(
+      organisation_id: @organisation.id,
+      sort: content_item_attribute,
+      order: params.present? && params[:order] == "asc" ? :desc : :asc
+    )
+  end
 end
--- a/app/views/content_items/index.html.erb
+++ b/app/views/content_items/index.html.erb
@@ -3,7 +3,7 @@
   <thead>
     <tr>
       <th>Title</th>
-      <th>Last Updated</th>
+      <th><%= order_link "Last Updated", :public_updated_at %></th>
     </tr>
   </thead>
   <tbody>
--- a/spec/controllers/content_items_controller_spec.rb
+++ b/spec/controllers/content_items_controller_spec.rb
@@ -2,26 +2,85 @@ require 'rails_helper'
 
 RSpec.describe ContentItemsController, type: :controller do
   describe "GET #index" do
-    let(:organisation) { create(:organisation_with_content_items) }
+    context "find by organisation" do
+      let(:organisation) { create(:organisation_with_content_items, content_items_count: 2) }
 
-    before do
-      get :index, params: { organisation_id: organisation }
+      before do
+        get :index, params: { organisation_id: organisation }
+      end
+
+      it "returns http success" do
+        expect(response).to have_http_status(:success)
+      end
+
+      it "assigns current organisation" do
+        expect(assigns(:organisation)).to eq(organisation)
+      end
+
+      it "assigns list of content items" do
+        expect(assigns(:content_items)).to eq(organisation.content_items)
+      end
+
+      it "renders the :index template" do
+        expect(subject).to render_template(:index)
+      end
     end
 
-    it "returns http success" do
-      expect(response).to have_http_status(:success)
-    end
+    context "sorts with parameters" do
+      let(:content_items) do
+        [
+          build(:content_item, id: 1, public_updated_at: Time.parse("2016-12-09")),
+          build(:content_item, id: 2, public_updated_at: Time.parse("2015-12-09"))
+        ]
+      end
+      let(:organisation) { create(:organisation, content_items: content_items) }
 
-    it "assigns list of content items" do
-      expect(assigns(:content_items)).to eq(organisation.content_items)
-    end
+      context "in ascending order" do
+        before do
+          get :index, params: { organisation_id: organisation, order: :asc, sort: :public_updated_at }
+        end
 
-    it "assigns the organisation" do
-      expect(assigns(:organisation)).to eq(organisation)
-    end
+        it "uses order, sort, organisation_id, controller and action params" do
+          expect(request.params.keys).to eq(%w(order sort organisation_id controller action))
+        end
 
-    it "renders the :index template" do
-      expect(subject).to render_template(:index)
+        it "uses asc as the order" do
+          expect(request.params[:order]).to eq("asc")
+        end
+
+        it "sorts by a valid content item attribute" do
+          content_item = build(:content_item)
+          expect(content_item.attributes).to have_key(request.params[:sort])
+        end
+
+        it "returns content_items based on supplied params" do
+          expect(assigns(:content_items).pluck(:id)).to eq([2, 1])
+        end
+      end
+
+      context "in descending order" do
+        before do
+          get :index, params: { organisation_id: organisation, order: :desc, sort: :public_updated_at }
+        end
+
+        it "uses order, sort, organisation_id, controller and action params" do
+          expect(request.params.keys).to eq(%w(order sort organisation_id controller action))
+          expect(request.params[:order]).to eq("desc")
+        end
+
+        it "uses desc as the order" do
+          expect(request.params[:order]).to eq("desc")
+        end
+
+        it "sorts by a valid content item attribute" do
+          content_item = build(:content_item)
+          expect(content_item.attributes).to have_key(request.params[:sort])
+        end
+
+        it "returns content_items based on supplied params" do
+          expect(assigns(:content_items).pluck(:id)).to eq([1, 2])
+        end
+      end
     end
   end
 end
--- a/spec/factories/content_items_factory.rb
+++ b/spec/factories/content_items_factory.rb
@@ -2,6 +2,6 @@ FactoryGirl.define do
   factory :content_item do
     sequence(:content_id) { |index| "7776ddf3-f918-5f32-bf18-dc1ced2eeeb#{index}" }
     link "api/content/item/path"
-    public_updated_at Time.parse("2016-11-01 11:20:45.481868")
+    public_updated_at { Time.now }
   end
 end
--- a/spec/factories/organisations_factory.rb
+++ b/spec/factories/organisations_factory.rb
@@ -1,5 +1,6 @@
 FactoryGirl.define do
   factory :organisation do
+    sequence(:id) { |index| index }
     sequence(:slug) { |index| "slug-#{index}" }
 
     factory :organisation_with_content_items do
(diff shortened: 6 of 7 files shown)


db/schema.rb
```diff
@@ -10,18 +10,19 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20161130153206) do
```
pmanrubia: Not sure why we are getting this change here. It seems that it is only changing the schema.

spec/factories/organisations_factory.rb
```diff
@@ -1,5 +1,6 @@
 FactoryGirl.define do
   factory :organisation do
+    sequence(:id) { |index| index }
```
pmanrubia: I am not sure we need the id here. If you remove it all tests pass. 

Would you mind re-wording the start of your second paragraph in the commit notes.

app/controllers/content_items_controller.rb
```diff
@@ -1,6 +1,6 @@
 class ContentItemsController < ApplicationController
   def index
     @organisation = Organisation.find(params[:organisation_id])
-    @content_items = @organisation.content_items.first(25)
+    @content_items = @organisation.content_items.order("#{params[:sort]} #{params[:order]}").first(25)
```
pmanrubia: Good work on not hardcoding the attribute

pmanrubia: It might be a good idea to use [strong parameters](https://github.com/rails/strong_parameters)

spec/controllers/content_items_controller_spec.rb
```diff
@@ -2,26 +2,77 @@
...
-      expect(assigns(:content_items)).to eq(organisation.content_items)
+      it "renders the :index template" do
+        expect(subject).to render_template(:index)
+      end
     end
 
-    it "renders the :index template" do
-      expect(subject).to render_template(:index)
+    context "sorts with parameters" do
+      context "in ascending order" do
```
pmanrubia: It seems that we are using the autogenerated `public_updated_at` for ordering. Might be a bit hidden?. I would be explicit about the `public_updated_at` in this test.

pmanrubia: Also, by setting the number of `content_items_count` to four we are slowing down the test as the before clause applies to every test.

spec/factories/content_items_factory.rb
```diff
@@ -2,6 +2,6 @@
   factory :content_item do
     sequence(:content_id) { |index| "7776ddf3-f918-5f32-bf18-dc1ced2eeeb#{index}" }
     link "api/content/item/path"
-    public_updated_at Time.parse("2016-11-01 11:20:45.481868")
+    public_updated_at { Time.at(0 + rand * (Time.now.to_f - 0.to_f)) }
```
pmanrubia: I would be more than happy with something like:

```ruby
public_updated_at { Time.now }
```

If we need a particular value, we just need to set up the test.

spec/views/content_items/index.html.erb_spec.rb
```diff
@@ -1,23 +1,27 @@
 require 'rails_helper'
 
 RSpec.describe 'content_items/index.html.erb', type: :view do
-  let(:content_items) { build_list(:content_item, 2) }
+  let(:organisation) { create(:organisation_with_content_items, content_items_count: 4) }
```
pmanrubia: Not sure we need 4 items here.

app/views/content_items/index.html.erb
```diff
@@ -5,7 +5,7 @@
       <th>Content Ids</th>
       <th>Content URL</th>
       <th>Title</th>
-      <th>Last Updated</th>
+      <th>Last Updated <%= link_to "Desc", organisation_content_items_path(organisation_id: @organisation.id,sort: :public_updated_at, order: :desc) %>&nbsp;&nbsp;<%= link_to "Asc", organisation_content_items_path(organisation_id: @organisation.id, sort: :public_updated_at, order: :asc) %></th>
```
pmanrubia: What about using a helper method?

app/controllers/concerns/content_items_concern.rb
```diff
@@ -0,0 +1,19 @@
+module ContentItemsConcern
```
pmanrubia: @ikennaokpala I am happy with extracting a concern if the pattern applies multiple times 👍 , but would you mind keeping this commit for the future (I am not sure about the name, for example) The same sorting pattern might apply to other lists, for example.

pmanrubia:
@ikennaokpala sorry for the early review, I thought this card was in Ready for Code Review 😊

spec/views/content_items/index.html.erb_spec.rb
ikennaokpala: Yeah i increased it to get a bigger sample size for testing, especially for sorting. This was a personal choice

app/controllers/concerns/content_items_concern.rb
ikennaokpala: According to the card this isn't meant to be hardcoded and should apply to all attributes.. Also for clarity, the merits of this extract to me is of immediate benefit.. Also I prefer this to having a somewhat fat controller.

spec/controllers/content_items_controller_spec.rb
ikennaokpala: Sure I can reduce the `content_items_count`. I prefer to be implicit and randomly generate the`public_updated_at`..

spec/factories/content_items_factory.rb
ikennaokpala: I thought about this, but the range diff was small and somewhat confusing.. rather than do this I prefer being a bit more explicit.

spec/factories/organisations_factory.rb
ikennaokpala: yeah sorry for its inclusion here.. it is actually required in later commits..

db/schema.rb
ikennaokpala: Yeah it is weird one. it kept re-occuring. I just checked it with the intention to remove it later on.

ikennaokpala:
@pmanrubia no worries.. your thought and observations are much appreciated.  :)

app/controllers/content_items_controller.rb
ikennaokpala: Cheers for you suggestions.. Will look into this

spec/controllers/content_items_controller_spec.rb
pmanrubia: If you set up the test in the factory (by being implicit) you are basically coupling a factory with a test. This is not a very good practice, in my opinion. You won’t be able, for example, to follow the same approach to test other stuff related to content items and `public_updated_at`.
By being explicit when setting up your test you:
1. Improve readability. You don’t need to go to the factory to understand how the ordering is being set
2. The factory will be simpler. This is a goal by itself.
3. The test will be easier to change.
4. If you remove the test, for example, you don’t need to track the logic that was set in the factory.

spec/factories/content_items_factory.rb
pmanrubia: As of today, a `ContentItem` only need a `public_updated_at` timestamp. If the test need some value in particular, then that value could be set in the test by being explicit.

spec/views/content_items/index.html.erb_spec.rb
pmanrubia: If we increase the number of items when setting up a test, we need a reason for it. Adding two more items in each test would imply creating more records in the database in each test, hence, we are slowing our tests.

app/controllers/concerns/content_items_concern.rb
pmanrubia: A few comments here:

1. I don’t think this is a fat controller for sorting the content items. Indeed the code is something like:

```ruby
   @content_items.order(params[:sort] => params[:order])
```
This does not look to me like a fat controller for our first test.

2. If we extract a concern for sorting, it does not need to be related to ContentItems, and in my opinion, it does not need to be related to a controller. It feels more close to a model, or to a QueryObject, in case you want to follow that path.

spec/factories/organisations_factory.rb
ikennaokpala: I have moved it to this commit 309d54644f4a1b9ce1e29b0ed0b62f0759a47c13 where it is used

spec/controllers/content_items_controller_spec.rb
ikennaokpala: Fine.. I believe this is worthy to be properly expanded in a collab session.

app/controllers/concerns/content_items_concern.rb
ikennaokpala: But then your suggestion seeks to associate/define these  concrete implementation i.e the methods withing the controller.. At what point would it be acceptable to introduce a data access or query object ??

pmanrubia:
@ikennaokpala feel free to supersede this PR, it is a bit difficult to follow up all our discussions.

ikennaokpala:
@pmanrubia will create another PR for this.. Cheers

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,9 @@
 module ApplicationHelper
+  def order_link(heading, content_item_attribute)
```
ikennaokpala: These have become redundant because of the removal of the concern

spec/controllers/content_items_controller_spec.rb
```diff
@@ -2,26 +2,85 @@
...
+        ]
+      end
+      let(:organisation) { create(:organisation, content_items: content_items) }
+
+      context "in ascending order" do
+        before do
+          get :index, params: { organisation_id: organisation, order: :asc, sort: :public_updated_at }
+        end
+
+        it "uses order, sort, organisation_id, controller and action params" do
```
ikennaokpala: These have become redundant because of the removal of the concern

app/helpers/application_helper.rb
ikennaokpala: Was omitted in error on Firday will push it shortly

spec/controllers/content_items_controller_spec.rb
```diff
@@ -2,26 +2,85 @@
...
+          expect(request.params.keys).to eq(%w(order sort organisation_id controller action))
+        end
+
+        it "uses asc as the order" do
+          expect(request.params[:order]).to eq("asc")
```
ikennaokpala: These have become redundant because of the removal of the concern

spec/controllers/content_items_controller_spec.rb
```diff
@@ -2,26 +2,85 @@
...
+        end
+
+        it "sorts by a valid content item attribute" do
```
ikennaokpala: These have become redundant because of the removal of the concern

spec/controllers/content_items_controller_spec.rb
```diff
@@ -2,26 +2,85 @@
...
+          content_item = build(:content_item)
+          expect(content_item.attributes).to have_key(request.params[:sort])
+        end
+
+        it "returns content_items based on supplied params" do
+          expect(assigns(:content_items).pluck(:id)).to eq([2, 1])
```
ikennaokpala: See #22

spec/controllers/content_items_controller_spec.rb
```diff
@@ -2,26 +2,85 @@
...
+        end
+      end
+
+      context "in descending order" do
```
ikennaokpala: These have become redundant because of the removal of the concern

PR #22, opened 2016-12-12
Handle ordering of content items by attribute in ascending or descending order

ikennaokpala:
Supersedes #21 

[Trello card](https://trello.com/c/jGICn84K/65-story-list-the-pages-that-have-not-been-updated-in-order-bigger-gap-first)

## Description 

As a content designer, when I click on the last_update_at column I change the ordering of the column (ascending and descending).

When a user clicks on the last update table column the ordering of the column changes to
ascending or descending.

Here two links have been included to help with sorting the column by the last updated time.

We cater for the order and sort params that come through and are used to sort the content items by `public_updated_at` in ascending and descending order.

The ordering format is as follows `order('<attribute> (ASC|DESC)')`.

This has not been hard coded and can be reused by any other content item
attribute.

Also helper method `order_link` is used to updated the ordering links. When no ordering is provided, the link defaults to descending, when descending is supplied the link produced is for ascending, vice versa.

These have been included to help with sorting the column by the
last updated time.

Following the conversations from #21 .. we have agreed to park 6f501c34d3dd3505f8bbb134e069a8a90385a2dd to [story/handle-content-ordering-cntd](https://github.com/alphagov/content-performance-manager/commits/story/handle-content-ordering-cntd), to be revisited in the coming sprints.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,2 +1,9 @@
 module ApplicationHelper
+  def order_link(heading, content_item_attribute, order)
+    link_to heading, organisation_content_items_path(
+      organisation_id: @organisation,
+      sort: content_item_attribute,
+      order: order.present? && order == "asc" ? :desc : :asc
+    )
+  end
 end
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -1,6 +1,6 @@
 class ContentItemsController < ApplicationController
   def index
     @organisation = Organisation.find(params[:organisation_id])
-    @content_items = @organisation.content_items.first(25)
+    @content_items = @organisation.content_items.order("#{params[:sort]} #{params[:order]}").first(25)
   end
 end
--- a/app/views/content_items/index.html.erb
+++ b/app/views/content_items/index.html.erb
@@ -3,7 +3,7 @@
   <thead>
     <tr>
       <th>Title</th>
-      <th>Last Updated</th>
+      <th><%= order_link "Last Updated", :public_updated_at, params[:order] %></th>
     </tr>
   </thead>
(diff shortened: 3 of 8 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,2 +1,9 @@
 module ApplicationHelper
+  def order_link(heading, content_item_attribute, order)
+    link_to heading, organisation_content_items_path(
```
selfthinker: This doesn't need to be addressed in this PR, but just a heads up of what changes we will need to make to the markup:

We will need to add a class and a title (or otherwise hidden copy), some of it potentially on the th and some on the link. (We might not need the class at all as we could use `[aria-sort] {}` for styling?)
This is some example markup (subject to change, would like to discuss with @theleebriggs):

```
<th class="sort none" aria-sort="none"><a href="..." title="Sort column ascending">Last Updated</a></th>
```
would be the default, and...

```
<th class="sort ascending" aria-sort="asc"><a href="..." title="Sort column descending">Last Updated</a></th>
```
...would be when it's sorted ascending. Sorted descending would be like the latter but the opposite.

I wonder if that means that the `th` should ideally be part of `order_link()`?

ikennaokpala: Thanks Anika.. Much appreciated.

pmanrubia:
Thank you @ikennaokpala for your work on this PR
LGTM

app/helpers/application_helper.rb
pmanrubia: Thank you @selfthinker, I am merging this PR now, but we are more than happy to address this changes in future PRs.

ikennaokpala:
@pmanrubia thanks

PR #23, opened 2016-12-12
Update an organisation or a content item if they already exist

leenagupte:
Trello card: https://trello.com/c/j7qkEq3Q

## Motivation

The first go at the import task just create new organisations and content items every time it ran. This meant that you could end up with the same organisation being stored in the database multiple times. 

If we load the same organisation twice, then we should have no duplications in the database. Instead, what we would need to do is to update the attributes of a Content Item if it already exists.

## Expected changes

Each organisation should only appear once on the `organisations` page, regardless of how many times it has been imported.

### DIFF
--- a/app/models/importers/organisation.rb
+++ b/app/models/importers/organisation.rb
@@ -9,7 +9,10 @@ class Importers::Organisation
   end
 
   def run
-    @organisation = ::Organisation.create!(slug: slug)
+    @organisation = ::Organisation.find_by(slug: slug)
+    if @organisation.blank?
+      @organisation = ::Organisation.create!(slug: slug)
+    end
 
     loop do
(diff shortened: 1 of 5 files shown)


app/models/importers/organisation.rb
```diff
@@ -9,7 +9,11 @@ def initialize(slug, batch: 10, start: 0)
   end
 
   def run
-    organisation = ::Organisation.create!(slug: slug)
+    organisation = ::Organisation.find_by(slug: slug)
+    if organisation.blank?
+      organisation = ::Organisation.create!(slug: slug)
+    end
+
```
pmanrubia: Could you please use `find_or_create_by`?

PR #31, opened 2016-12-21
Add content item details page

leenagupte:
Trello card: https://trello.com/c/dZNJBYlc

## Details

Add a new page that shows all of the details for a single content item.

Currently that is limited to the information that already appears in the content item list:

* Title
* URL on GOV.UK
* Document type
* Last updated date
* Organisation name

### Before

screen shot 2016-12-21 at 12 13 44

### After
contentperformancemanager

screen shot 2016-12-21 at 12 18 19

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -6,5 +6,6 @@ class ContentItemsController < ApplicationController
 
   def show
     @content_item = ContentItem.find(params[:id])
+    @organisation = Organisation.find(params[:organisation_id])
   end
 end
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -5,6 +5,6 @@ class ContentItemsController < ApplicationController
   end
 
   def show
-    head :ok
+    @content_item = ContentItem.find(params[:id])
   end
 end
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -3,4 +3,8 @@ class ContentItemsController < ApplicationController
     @organisation = Organisation.find(params[:organisation_id])
     @content_items = @organisation.content_items.order("#{params[:sort]} #{params[:order]}").page(params[:page])
   end
+
+  def show
+    head :ok
+  end
 end
--- a/spec/controllers/content_items_controller_spec.rb
+++ b/spec/controllers/content_items_controller_spec.rb
@@ -56,13 +56,22 @@ RSpec.describe ContentItemsController, type: :controller do
 
   describe "GET #show" do
     context "find by content item" do
-      let(:organisation) { create(:organisation_with_content_items, content_items_count: 2) }
-      let(:content_item) { build(:content_item, id: 1) }
+      let(:organisation) { create(:organisation_with_content_items, content_items_count: 1) }
+
+      before do
+        get :show, params: { organisation_id: organisation, id: organisation.content_items.first.id }
+      end
 
       it "returns http success" do
-        get :show, params: { organisation_id: organisation, id: content_item.id }
-
         expect(response).to have_http_status(:success)
+      end
+
+      it "assigns current content item" do
+        expect(assigns(:content_item)).to eq(organisation.content_items.first)
+      end
+
+      it "renders the :show template" do
+        expect(subject).to render_template(:show)
       end
     end
   end
--- a/spec/controllers/content_items_controller_spec.rb
+++ b/spec/controllers/content_items_controller_spec.rb
@@ -53,4 +53,17 @@ RSpec.describe ContentItemsController, type: :controller do
       end
     end
   end
+
+  describe "GET #show" do
+    context "find by content item" do
+      let(:organisation) { create(:organisation_with_content_items, content_items_count: 2) }
+      let(:content_item) { build(:content_item, id: 1) }
+
+      it "returns http success" do
+        get :show, params: { organisation_id: organisation, id: content_item.id }
+
+        expect(response).to have_http_status(:success)
+      end
+    end
+  end
 end
--- a/spec/controllers/content_items_controller_spec.rb
+++ b/spec/controllers/content_items_controller_spec.rb
@@ -70,6 +70,10 @@ RSpec.describe ContentItemsController, type: :controller do
         expect(assigns(:content_item)).to eq(organisation.content_items.first)
       end
 
+      it "assigns current organisation" do
+        expect(assigns(:organisation)).to eq(organisation)
+      end
+
       it "renders the :show template" do
         expect(subject).to render_template(:show)
       end
--- a/spec/features/content_item_spec.rb
+++ b/spec/features/content_item_spec.rb
@@ -3,7 +3,7 @@ require 'rails_helper'
 RSpec.feature "Content Item Details", type: :feature do
   let(:organisation) { create(:organisation_with_content_items, content_items_count: 1) }
 
-  xscenario "the user clicks on the view content item link and is redirected to the content item show page" do
+  scenario "the user clicks on the view content item link and is redirected to the content item show page" do
     visit "organisations/#{organisation.id}/content_items"
     click_on "View detail"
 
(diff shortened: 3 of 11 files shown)


spec/routes/content_item_spec.rb
```diff
@@ -11,5 +11,16 @@
           organisation_id: '1'
         )
     end
+
+    it 'routes to #show' do
+      expect(get: organisation_content_item_path(organisation_id: 1, id: 1)).to be_routable
```
pmanrubia: `be_routable` is already tested in line 17. Why are you testing this?

spec/views/content_items/index.html.erb_spec.rb
```diff
@@ -68,5 +68,13 @@
 
       expect(rendered).to have_link('Last Updated', href: href)
     end
+
+    it 'has a link to the content item details page' do
+      render
+      expect(rendered).to have_link('View detail', href: organisation_content_item_path(
```
pmanrubia: Very minor: Could you please leave an empty line between the `render` and the expectation?

spec/views/content_items/show.html.erb_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  let(:content_item) { build(:content_item) }
+
+  it 'renders the title of the content item' do
+    content_item.title = 'A Title'
+    assign(:content_item, content_item)
+    render
+
+    expect(rendered).to have_selector('h1', text: 'A Title')
+  end
+
```
pmanrubia: Could you please render the organisation name?

```ruby
xit 'renders the name of the organisation`
```

app/controllers/content_items_controller.rb
```diff
@@ -5,6 +5,6 @@ def index
   end
 
   def show
-    head :ok
```
pmanrubia: What about a controller test?

spec/views/content_items/show.html.erb_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  xit 'renders the url of the content item'
+
+  xit 'renders the document type of the content item'
+
+  xit 'renders the last updated date of the content item'
+end
```
pmanrubia: What about removing `of the content item` from each test?

```ruby
Randomized with seed 41922

content_items/show.html.erb
  renders the title of the content item
  renders the last updated date of the content item
  renders the document type of the content item
  renders the url of the content item

Finished in 0.1599 seconds (files took 2.33 seconds to load)
4 examples, 0 failures
```

spec/views/content_items/show.html.erb_spec.rb
```diff
@@ -28,5 +28,13 @@
     expect(rendered).to have_text('guidance')
   end
 
-  xit 'renders the last updated date of the content item'
+  it 'renders the last updated date of the content item' do
+    Timecop.freeze('2016-3-20')
```
pmanrubia: It would be better if you use a block:

```ruby
Timecop.freeze do
# sfuff with freezed time
end
```

spec/views/content_items/show.html.erb_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    assign(:content_item, content_item)
+    render
+
+    expect(rendered).to have_text('Page on GOV.UK')
+    expect(rendered).to have_link('https://gov.uk/content/1/path', href: 'https://gov.uk/content/1/path')
+  end
+
+  it 'renders the document type of the content item' do
+    content_item.document_type = 'guidance'
+    assign(:content_item, content_item)
```
pmanrubia: what about extracting `assign(:content_item, content_item)` to a new block?

spec/features/content_item_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+
+RSpec.feature "Content Item Details", type: :feature do
```
pmanrubia: Well done on the feature test.  Two comments:

1. Could you please add this test to the first commit? No need to fail, you can mark it as pending
2. I would change the expectation test. It is better not to test the current path, but something in the view that implies that it is currently working fine. For example, the content item title.

spec/routes/content_item_spec.rb
leenagupte: Doesn't line 16 check that you've added the route in config/routes, and line 17 check that the controller has been setup correctly? That's what I was going for.

spec/views/content_items/show.html.erb_spec.rb
leenagupte: What do you mean?

spec/views/content_items/show.html.erb_spec.rb
leenagupte: I was going to do that, then completely forgot. Will do it now.

spec/features/content_item_spec.rb
leenagupte: Wouldn't testing the title make it the same as a view test, rather than an integration test?
I was trying not to repeat the `title` test in `show.html.erb_spec.rb`.

spec/views/content_items/show.html.erb_spec.rb
leenagupte: Ok.

spec/views/content_items/show.html.erb_spec.rb
leenagupte: I can change it, I'm not bothered either way.

app/controllers/content_items_controller.rb
leenagupte: Oh! I didn't think about those. I assume I add test to check the status when the `show` method is added, and then the other tests when create the `@content_item` instance variable?

spec/features/content_item_spec.rb
leenagupte: I've added a pending version of the test as first commit.

leenagupte:
@pmanrubia I've made most the changes you suggested and pushed the code.

The two comments I have questions about are:

> I would change the expectation test. It is better not to test the current path, but something in the view that implies that it is currently working fine. For example, the content item title.

> Could you please render the organisation name?

`xit 'renders the name of the organisation`

app/controllers/content_items_controller.rb
pmanrubia: Yep, at least we would need it when the variable is assigned.

spec/features/content_item_spec.rb
pmanrubia: Integration tests exercise all the stack without being exhaustive in the tests; so my feeling is that it would be ok to check for the title of the content item to be ok. 
I like the `Then I am on the content item detail page.`. But I would not rely on the browser URL as that would also pass with a blank page. But, feel free to follow the approach you think is better.

spec/views/content_items/show.html.erb_spec.rb
pmanrubia: Sorry if this wasn't clear on the card. I just thought it would be useful to get the organisation title when rendering a content item.

spec/controllers/content_items_controller_spec.rb
```diff
@@ -53,4 +53,17 @@
       end
     end
   end
+
+  describe "GET #show" do
+    context "find by content item" do
+      let(:organisation) { create(:organisation_with_content_items, content_items_count: 2) }
+      let(:content_item) { build(:content_item, id: 1, public_updated_at: Time.parse("2016-12-09")) }
```
pmanrubia: We are not using `public_updated_at` here. What about removing it?

pmanrubia:
@leenagupte LGTM

spec/features/content_item_spec.rb
leenagupte: I think I'll leave this one as is. The content is being tested in the view test. This test is checking that you are in the right place, the right page, whether it's blank or not.

spec/controllers/content_items_controller_spec.rb
leenagupte: Yeah ok. I'll do that this afternoon.

spec/views/content_items/show.html.erb_spec.rb
leenagupte: Oh. I'll see what I can do. I might just tack on a commit to do this at the end, rather rebasing again.

PR #33, opened 2016-12-21
Table sorting with gov.uk styling

theleebriggs:
# Trello
https://trello.com/c/86AooFW7/99-make-table-sorting-use-gov-uk-styling-and-conventions

# Description
Added a new rails helper to generate sortable table headers on the content items list view - adds screen reader text and style hooks to show up/down arrows when the table is sorted by that column.

Includes a new sass partial _tables.scss with light weight table styles.

# Preview

screen shot 2016-12-21 at 15 30 58


app/helpers/table_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+      sort_link heading, next_sort[:label], content_item_attribute, next_sort[:param]
+    end
+  end
+
+private
+
+  def sort_link(heading, description, content_item_attribute, order)
+    link_content = heading + " " + (content_tag :span, description, class: "visually-hidden")
+    link_content = link_content.html_safe
+    link_to link_content, organisation_content_items_path(
```
pmanrubia: You can use a [block in link_to](http://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to) to define the link content.

app/helpers/table_helper.rb
```diff
@@ -0,0 +1,42 @@
+module TableHelper
+  def sort_table_header(heading, content_item_attribute, order, sort)
+    # Determine if this column is currently sorted on
```
pmanrubia: I would love to find a way to get rid of lines 3-8

1. Lines 3-5 are comments, and if possible, we shouldn't add them within the code
2. Lines 5-8 could be replaced with method calls within lines 9-10

To do so, you might need to split method `next_sort` into two (one to get the label and other to get the param)

In general, the name of the methods and variable are not obvious to me. `next sort`, `current_sort`, for example, are not revealing what they return. You need to navigate to the implementation of the method to find out what they do.

What about pairing with another developer and try to find better names? Try to pair with @chao-xian for example.

spec/helpers/table_helper_spec.rb
```diff
@@ -0,0 +1,82 @@
+require "rails_helper"
+
+RSpec.describe TableHelper, type: :helper do
+  describe "#sort_table_header", :sort_table_header do
+    let(:organisation) { build(:organisation) }
+    let(:header_label) { "Last Updated" }
```
pmanrubia: it would be good to use the same names as in the helper. If you use `heading` keeping it consistent in the test will make it easier to read.
Same applies to `header_content_attribute`

spec/helpers/table_helper_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    before do
+      assign(:organisation, organisation)
+    end
+
+    describe 'Accessibility' do
+      context 'When the column is unsorted' do
+        it "sets aria-sort to none " do
+          header = helper.sort_table_header(header_label, header_content_attribute, nil, nil)
```
pmanrubia: The two extra parameters (`nil`, `nil`) would go away if we don't pass them as parameters and use `params` directly.

app/views/content_items/index.html.erb
```diff
@@ -2,9 +2,9 @@
 <table>
   <thead>
     <tr>
-      <th>Title</th>
-      <th>Type of Document</th>
-      <th><%= order_link "Last Updated", :public_updated_at, params[:order] %></th>
+      <%= sort_table_header "Title", "title", params[:order], params[:sort] %>
```
theleebriggs: I've updated the template and the helper to use params[:key] now.

app/helpers/table_helper.rb
```diff
@@ -0,0 +1,42 @@
+module TableHelper
+  def sort_table_header(heading, content_item_attribute, order, sort)
```
theleebriggs: Renamed helper and spec references to use `attribute_name`.

app/helpers/table_helper.rb
theleebriggs: I have removed the comments and split the methods as suggested - also addressed the naming while pairing with @chao-xian. Looked into using a `class` but it started to get quite complicated giving the class access to the `ActionView` helpers such as `content_tag`, `params[]` especially in the tests and as such might not be worth it - more than willing to pair with you if you have a vision for how it could work.

app/helpers/table_helper.rb
theleebriggs: Update as per suggestion.

app/helpers/table_helper.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def current_sort(currently_sorted, heading, order)
+    aria_attr = currently_sorted ? readable_sort_term(order) : "none"
+    label = currently_sorted ? "sorted by #{heading}, #{aria_attr}" : ""
+    { aria_attr: aria_attr, label: label }
+  end
+
+  def readable_sort_term(term)
+    order_term_map = { "asc" => "ascending", "desc" => "descending" }
```
theleebriggs: Updated as per second suggestion. `map[term] || 'ascending'`

spec/helpers/table_helper_spec.rb
theleebriggs: Names used in test now match helper names.

spec/helpers/table_helper_spec.rb
theleebriggs: Resolved as helper now uses `params[]`.

pmanrubia:
@theleebriggs @chao-xian thank you for your work on this.

Regarding using a class for the helper... what about this?

```ruby
module TableHelper
  def sort_table_header(heading, attribute_name)
    SortTable.new(self, heading, attribute_name).render
  end

  class SortTable < Struct.new(:view, :heading, :attribute_name)
    delegate :content_tag, :params, to: :view

    def render
      content_tag :th, 'aria-sort' => aria_label(attribute_name) do
        link heading, text_label(heading, attribute_name), attribute_name, order_param(attribute_name)
      end
    end

  private

    def link(heading, label, attribute_name, order)
      view.link_to view.organisation_content_items_path(organisation_id: view.instance_variable_get(:@organisation), sort: attribute_name, order: order) do
        "#{heading} #{content_tag :span, label, class: 'visually-hidden'}".html_safe
      end
    end
   
   # ...

  end
end
```

app/assets/stylesheets/_tables.scss
```diff
@@ -0,0 +1,36 @@
+[aria-sort] > * {
+  position: relative;
+
+  &:after {
+    font-size: em(10, 19);
```
selfthinker: GOV.UK uses pixels for font sizes everywhere. (I personally prefer `em` as well but it makes sense to stay consistent.) When you use [the typography mixins from the frontend toolkit](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_typography.scss) the sizes will automatically be in `px`.

In this case I'd actually prefer a bigger arrow, but maybe double check with the design community first before changing that.

app/assets/stylesheets/_tables.scss
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  &:hover {
+    color: $link-hover-colour;
+  }
+}
+
+[aria-sort=descending],
+[aria-sort=ascending] {
+    padding-right: em(30, 19);
```
selfthinker: Minor: Indentation has two spaces too many.

app/assets/stylesheets/_tables.scss
```diff
@@ -0,0 +1,36 @@
+[aria-sort] > * {
+  position: relative;
+
+  &:after {
+    font-size: em(10, 19);
+    position: absolute;
+    right: -15px;
```
selfthinker: Minor: If there are any "magic numbers" like here (or below on the `margin-top`), I would try to make them relate to some other size (e.g. a fraction of the size of the arrow) and then explain that in a comment if it isn't obvious.

app/assets/stylesheets/_tables.scss
```diff
@@ -0,0 +1,36 @@
...
+}
+
+[aria-sort=descending],
+[aria-sort=ascending] {
+    padding-right: em(30, 19);
+}
+
+[aria-sort=descending] > * {
+  &:after {
+    content: '\00a0\00a0\25B2';
```
selfthinker: The two spaces (to the left of the arrow) don't seem to be necessary because the arrow is absolutely positioned to the right.
As it's not obvious what the characters are, I would add a comment, saying that it is an "up (or down) arrow/triangle".

app/helpers/table_helper.rb
```diff
@@ -0,0 +1,61 @@
...
+      end
+    end
+
+    def currently_sorted_by?
+      params[:order].present? && params[:sort] == attribute_name
+    end
+
+    def text_label
+      param = order_param
+      "sort by #{heading}, #{readable_sort_term(param)}".downcase
```
selfthinker: What's hidden in the `th` would be read out by screen readers in every `td` underneath. Therefore I would try to keep this as short as possible and not duplicate anything. I would suggest something like "sort column descending" or just "sort descending".

app/assets/stylesheets/_tables.scss
```diff
@@ -0,0 +1,36 @@
+[aria-sort] > * {
+  position: relative;
```
selfthinker: This together with the `.visually-hidden` span is causing a weird effect in Firefox on focus:
table-sorting-focus

I tried to play around to find some solution but nothing I tried was ideal. I don't think it's a big issue. So, maybe just spend 10 minutes with it and ignore it if you cannot find a solution within that time.

app/assets/stylesheets/_tables.scss
theleebriggs: Updated to use pixel sizing - didn't use mixin as it was clearly to read it with plain pixel values .

app/assets/stylesheets/_tables.scss
theleebriggs: Corrected now.

app/assets/stylesheets/_tables.scss
theleebriggs: Commented numbers appropriately now - all relative to arrow sizing.

app/assets/stylesheets/_tables.scss
theleebriggs: Removed unnecessary space characters as elements is positioned by it's right hand side so these are useless.

app/helpers/table_helper.rb
theleebriggs: Added comma and simplified message.

app/assets/stylesheets/_tables.scss
theleebriggs: I have a potential solution stashed locally which I can run you through.

selfthinker:
I'm happy with the frontend changes. :+1:
@pmanrubia, what about the backend?

pmanrubia:
@theleebriggs Great job 👍

PR #35, opened 2016-12-22
GOV.UK styled pagination 

theleebriggs:
# trello

https://trello.com/c/bwNfIegw/98-3-make-pagination-use-gov-uk-styling-and-conventions

# Description

To allow the user to navigate through paged content items using existing gov.uk conventions e.g.
http://govuk-component-guide.herokuapp.com/components/previous_and_next_navigation

This feature adds a Kaminari theme to the views folder to add the gov-uk style hooks to previous and next links. There are new styles ported from govuk_components which should be removed once / if slimmer is added to the application.

# Preview

screen shot 2016-12-22 at 17 30 42


pmanrubia:
@leebrigs (cc/ @selfthinker). 

Thank you for your work on this.

I would moved the feature tests into the view tests.
Integration tests are slow and are not intended for exhaustive testing. I would prefer to test the label of the pagination component within the view.

If you run 
```
$ bundle exec spec --profile
```

You can see how the pagination spec is taking 50% of the testing time.

```
Top 10 slowest examples (1.53 seconds, 66.5% of total time):
  Content Item's Pagination When user navigates to the next page from page 1 the next page number is 3
    0.70708 seconds ./spec/features/pagination_spec.rb:45`
```

View tests are not creating anything in the database, hence, they will perform much faster.

pmanrubia:
@theleebrigs sorry I didn't tell you but we have a [guideline for commits](https://github.com/alphagov/styleguides/blob/master/git.md). 
I'd try to use imperative / present tense for the commit title, instead of past tense (tell what applying the commit will do, rather than what you did)

nickcolley:
What's blocking using static (slimmer) currently?

theleebriggs:
@nickcolley (cc/ @selfthinker) As far as I know nothing is blocking us using static at the moment it was more that we'd started with govuk-elements having enough for us to use and we wanted to test if the pagination included in static worked for users with the content in this app or if we'd need a new pattern. I'll check with @selfthinker on this.

selfthinker:
@nickcolley, when we were talking about using slimmer/static we didn't merge its implementation in the end because we thought it would be wrong to make this app look like it's part of mainstream.
Additionally, this was the first (and likely to be only) instance where we needed something from static.
Otherwise the [implementation of static](https://github.com/alphagov/content-performance-manager/tree/use-slimmer) would be ready and we could potentially use it in case we needed another component. The only thing sort of blocking that is that this app isn't yet set up the way other gov.uk apps are set up (which can be `bowl`ed with dependencies etc).

nickcolley:
Had a chat with @theleebriggs and @selfthinker, who said they're going to stay as is for now since they're still prototyping, when the application is stabler can look at using static.

selfthinker:
@theleebriggs, I have just force pushed some changes. I separated my main commit into 3 smaller commits (which I planned to do before I got ill) and I incorporated your test fixes into those commits accordingly.

I didn't plan to change any of the code but ended up simplifying the selector in the test. I changed e.g. the original `.govuk-previous-and-next-navigation .next-page .pagination-label` to `.govuk-previous-and-next-navigation [rel=next]`. That makes it more flexible and easier to change in case the markup changes.
(I'm not sure how much of that will change when those tests get moved to view tests.)

PR #37, opened 2016-12-29
Refactor Importers::Organisations

pmanrubia:
This is the first iteration into refactoring `Importers::Organisations`.  There is more work to do, it is now in a good position to create a PR.

## Why

This is a big refactor for [Importers::Organisations][1]. This class was doing too many things; it broke the single responsibility principle and it is [making things really hard to test][2]. We decided to postpone refactoring it because we were not sure about the source and format of the information we were fetching from the Content Store and the Search API.

To list a few things this class needed to know about:

1. SearchAPI query format and URLs
2. Content Store query format and URLs
3. Pagination through search API responses
4. Data transformation
5. Persistence of ContentItems and Organisations
 
## Changes

1. Extraction of two services: one to retrieve Content Items from the ContentStore, and another one to fetch all the Organisations.
2. Isolation of the clients to connect to the Content Store and the Search API
3. New rake task to import all organisations.
4. Creation of a new importer, to load all the organisations in one go.
5. Add feature tests to exercise all the stack because we are mocking most of the interactions.

## Next steps

1. Introduce query objects to query Content Store and Search API.
2. Decrease the number of mocks.
3. Return ContentItems in the ContentItemsService and Organisations in the OrganisationsService. 
4. Move to the models the methods `create_or_update!` 
5. Clients for the content store and the search API would need to be replaced by GOVUK adapters. We will do it once we move to the new infrastructure.
6. Use `Webmock` instead of stubbing `HTTParty`

[1]: https://github.com/alphagov/content-performance-manager/blob/cd0add35d9ae70869bd58d1ac9e7bca009c5671d/app/models/importers/organisation.rb
[2]: https://github.com/alphagov/content-performance-manager/blob/064351e6bba267ac5d47da28629419b92120e712/spec/models/importers/organisation_spec.rb


pmanrubia:
Superseeded by #38

PR #38, opened 2016-12-30
Refactor Importers::Organisations

pmanrubia:
This is the first iteration into refactoring `Importers::Organisations`.  There is more work to do, it is now in a good position to create a PR.

## Why

This is a big refactor for [Importers::Organisations][1]. This class was doing too many things; it broke the single responsibility principle and it is [making things really hard to test][2]. We decided to postpone refactoring it because we were not sure about the source and format of the information we were fetching from the Content Store and the Search API.

To list a few things this class needed to know about:

1. SearchAPI query format and URLs
2. Content Store query format and URLs
3. Pagination through search API responses
4. Data transformation
5. Persistence of ContentItems and Organisations
 
## Changes

1. Extraction of two services: one to retrieve Content Items from the ContentStore, and another one to fetch all the Organisations.
2. Isolation of the clients to connect to the Content Store and the Search API
3. New rake task to import all organisations.
4. Creation of a new importer, to load all the organisations in one go.
5. Add feature tests to exercise all the stack because we are mocking most of the interactions.

## Next steps

1. Introduce query objects to query Content Store and Search API.
2. Decrease the number of mocks.
3. Return ContentItems in the ContentItemsService and Organisations in the OrganisationsService. 
4. Move to the models the methods `create_or_update!` 
5. Clients for the content store and the search API would need to be replaced by GOVUK adapters. We will do it once we move to the new infrastructure.
6. Use `Webmock` instead of stubbing `HTTParty`

[1]: https://github.com/alphagov/content-performance-manager/blob/cd0add35d9ae70869bd58d1ac9e7bca009c5671d/app/models/importers/organisation.rb
[2]: https://github.com/alphagov/content-performance-manager/blob/064351e6bba267ac5d47da28629419b92120e712/spec/models/importers/organisation_spec.rb


spec/services/organisations_service_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      subject.find_each {}
+    end
+
+    it 'yields the response' do
+      result = []
+      allow(Clients::SearchAPI).to receive(:find_each).and_yield(:a).and_yield(:b)
+      subject.find_each { |value| result << value }
+
+      expect(result).to eq([:a, :b])
```
ikennaokpala: I suggest `match_array` here and also in [this](https://github.com/alphagov/content-performance-manager/blob/b979b439c41b39e4602409febc7c99fd26cc68df/spec/services/content_items_service_spec.rb#L26) as well.. see [this](https://www.relishapp.com/rspec/rspec-expectations/docs/built-in-matchers)

README.md
```diff
@@ -32,11 +32,17 @@ In the browser navigate to:
...
 ```bash
-$ rake import:organisation[{department-slug}]
+$ rake import:all_organisations
 ```
 
-Every time you add a new field to the database, you need to delete the existing organisations, and then re-import them.
+* To create or update the content items for an organisation:
+
+```bash
+$ rake import:content_items_by_organisation[{department-slug}]
```
ikennaokpala: This is needs to state that ``` rake import:all_organisations ``` needs be run before rake 'import:content_items_by_organisation[public-health-england]'

app/models/importers/content_items_by_organisation.rb
```diff
@@ -0,0 +1,20 @@
+module Importers
+  class ContentItemsByOrganisation
+    def run(slug)
+      organisation = Organisation.find_by(slug: slug)
```
ikennaokpala: Previously, we used ```find_or_create_by``` for similar, if so then no need for this [comment](https://github.com/alphagov/content-performance-manager/pull/38/files#r94231710)

spec/models/importers/all_organisations_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'rails_helper'
+
+RSpec.describe Importers::AllOrganisations do
+  describe '#import_all' do
+    context 'when the organisation does not exist' do
```
ikennaokpala: I think you mean ```run``` here

spec/models/importers/all_organisations_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'rails_helper'
+
+RSpec.describe Importers::AllOrganisations do
+  describe '#import_all' do
```
ikennaokpala: I think you mean ```run``` here

README.md
pmanrubia: Addressed in d01880d: I have added the word `existing` to the description of the rake task, because this is good enough for me.
You said in your first comment, that there is a lot of documentation missing so please, feel free to add it in a separate PR.

app/models/importers/content_items_by_organisation.rb
pmanrubia: I would prefer a hard fail if the organisation is not found because:

1. It is a bit ackward to load the content items of an organisation that does not exist. 
2. The organisation has other attributes than the slug. It doesn't feel right to me to have organisations only with a slug
3. This would allow the creation of an organisation when the user has a typo.

spec/services/organisations_service_spec.rb
pmanrubia: Addressed in a2d559fac05ee4ff327ef4b732080daf4377dd29 and fb0d8ece361b38e52ca2c5414e246da9af8d1206

spec/models/importers/all_organisations_spec.rb
pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/38/commits/f68ab65364b9529a8166dff70abd82f882bc3edc

pmanrubia:
> The naming of classes and methods in importers and the name-spacing in services could mirror the aforementioned theme. Ideally naming of a class follow noun pattern and methods mostly verbs.

Not sure what do you mean by this. I do think I am mirroring it using proper naming for classes and methods. Indeed I find that your suggestion is not following the pattern you advice: `Importers::ContentItem.new.by_organisation` is not using a verb for the method name.

> You may want to consider the renaming Importers::AllOrganisations to Importers::Organisation and renaming (i.e in this context) the somewhat obscure def run..;end method to def all..; end so this would read Importers::Organisation.new.all.

> Same goes for Importers::ContentItemsByOrganisation, here the relationship between ContentItem and Organisation is made too apparent and to a large extent rigid. IMHO it should be ContentItem and then this exposing how (i.e a ContentItem) it establishes its relationship (i.e with an Organisation via it's slug) in its behaviour or method. So run which imports ContentItems by an organisation's slug should be renamed todef by_organisation(slug);...end;, Hence Importers::ContentItem.new.by_organisation('slug')

> The intent here is to foster sensible defaults, avoid plurality, avoid any future mis-representation of identity (i.e easier to visualize / represent in Diagrams and easier to know where future changes should be made) and to an extent make clearer the specs/tests narratives. An example of the latter can be found here.

> Observe that expect { subject.run('the-slug') }.to change { ContentItem.count }.by(0) and expect { subject.by_organisation('the-slug') }.to change { ContentItem.count }.by(0) the latter more explicit and keeps the context.

> Also this will lead to narrative like these:

Sorry @ikennaokpala, I am not very happy with using the pattern `Importers::ContentItem.new.by_organisation`, for the same reason I would prefer not adding all the `find` methods to a model. This is why:

1. It is likely that you end up with a **fat importer** in the same way you would end up with a **fat model**. It might feel right at the beginning, but it doesn't evolve well. 
2. This approach promote long classes doing more than 1 thing at a time. This could cause a number of issues like **merge conflicts** and other clashes when developers modify the same class for two different things.
3. Adding **new importers** for Content Items will also be more tricky, as they **need to be accommodated in the same class**. Say you want to import all content items that have been updated in the last 24 hours. I would prefer not to add a new method to the Importers::ContentItem, and create a new class with its unit test.
4. As a general principle, it feels to me that it **breaks the Single Responsibility Principle**, or at least, we are not promoting it.
5. When it comes to your RSpec example, agree that it does not read very well, but I think it is related to a misused of `subject`. The alternative is not great either. I will prepare a commit to make it more explicit.
6. When it comes to the narrative examples, this narrative is good enough for me, I honestly don't see a big difference between your examples and mine:

```
Importers::ContentItemsByOrganisation
  #run
    when the content item does not exist
      creates a content item per attribute group
      updates the attributes
    when the content item already exists
      updates the attributes
      does not create a new one
```


> The same theme also goes for all under Services, consider name-spacing with Services preceding rather the ...Service suffix. The Clients namespace is an example of this, SearchAPI and ContentStore stand out and their find_each and find methods convey a somewhat multiple and singular message respectively.

I am not sure either about using a namespace instead of a Suffix, although I don't have a strong opinion about it. 

These are the pros for using a Suffix for Services:

1. **It is the way it is usually done in Rails community, or at least, I have always seen it this way**. If you search for Service Objects and Rails in Google, most of the references uses the suffix (in other languages like Java is also a common practice)
2. Using suffix seems to be the **Rails way for: controllers, helpers, mailers and jobs**. 
3. I don't favor using a namespace unless I have a folder for it, although, I know this is just a personal preference.

> This commit e21ef60 appears redundant further down the line of commits. A possible fixup.

Not sure why you think this is redundant. Our test suite had a timezone issue that surfaced once I run them in Spain. 

I placed the fix as the first commit because I wanted to make it explicit that I was fixing an existing issue. If I fix it up I will hide this fact, so I am not very happy doing it.

> DRY-UP and other smaller commits IMHO can also be squashed or fixed up.

Sorry, I disagree here as well. A dry-up commit do not need to be squashed. I didn't dry-up before because I wanted to have readable diffs to facilitate the review, as the whole PR is not easy to follow.

> Enjoyed the Collectors idea but noticed later you move away from it, really why? 🤔. I encourage you to re-visit this area and possibly to squash or fix it up to make this easier to follow. IMHO importers and its related parts are growing, we should consider moving it back to /lib, this is because it plays a complimentary role and to a large extent it is becoming isolated/independent in it's own right.

1. Agree that collectors is a cool name, but I wasn't sure about them because what we have is very simple, and move into collectors feels like over-engineering. Once we start doing more advanced stuff, we can revisit it.
2. Not sure why you want to move stuff to the lib folder. `lib` is not `autoloaded` in Rails 5, and if you autoload it, you need to add it to the `eager_load_paths` folder; this is not recommended, as the lib folder is supposed to be used by non-vendor plugins.
3. But even without the lib issue in Rails 5, I can't see the benefit of having the importers in the lib folder, to be honest. Indeed, I would like to run away from having a lib folder in our app, or use it only for vendor plugins.

> Readme section on the importers rake tasks needs to be a lot more explicit so none Rails devs can follow with little hand holding. Also, some background given of the SearchAPI and ContentStore and also information on errors like Errno::ETIMEDOUT: Failed to open TCP connection to www.gov.uk:443 (Operation timed out - connect(2) for "www.gov.uk" port 443) would be a nice to have.

I don't know how I can be **a lot more explicit** on something so simple, but I would love to learn how to do it. I will be more than happy to accept a PR with improvements to the documentation.

If you want to document TCP errors in the PR, although I am ok with it, but please, add a markdown document into the docs folder in the app.

@ikennaokpala I appreciate all your feedback, and although I think it is legitimate, I just think we have very different ways to develop Rails apps. I guess this is the reason behind not being happy accepting these changes.

ikennaokpala:
@pmanrubia glad to read through your comments.. the context of the renaming part of this review was from a more traditional OOP view. I understand this doesn't alway cut it. I am delighted that your changes here have brought in some progress. I hope this PR comments from both of us will be re-visited at some point.

pmanrubia:
@ikennaokpala thank you again for your review. I am very happy to revisit this approach at any time, but I would like to merge it as this PR is moving forward the codebase into a better position.

I am closing this PR, so the comments and `fixup`s are not removed.

Superseeded by #39

PR #36, opened 2016-12-23
Show the number of views per content item in the table

ikennaokpala:
# This is no longer ready for review due to the recent merged changes from #39 

- Following chats from #38, moving Services::GoogleAnalytics from lib to app
- Also changing metric from ga:pageviews to ga:uniquePageviews, According to Performance analysts.. "*ga:uniquePageviews is a better measure of times the content was viewed in a session"*"

[Trello card](https://trello.com/c/a3I8aKal/89-8-show-the-number-of-views-per-content-item-in-the-table)

## Description 

For every content item, content designers want to see the number of page views for a content item belonging to an organisation.

Here a new column/attribute (i.e `number_of_views`) has been added to the
content_item(s).

The `number_of_views` attribute’s value is obtained from the Google Analytics Reporting API V4 [1].

This response from the API is returned as an instance of [1]. when this is converted to JSON it returns [3].

For more on query, see [2], [4] and [5]

For obtaining the page views a piece of content_item, this endpoint [1] is queried with the base_path supplied.

This PR also updates the content items view table to show the page views.

[1] https://analyticsreporting.googleapis.com/v4/reports:batchGet

[2] https://developers.google.com/analytics/devguides/reporting/core/v4/parameters
[3] https://developers.google.com/analytics/devguides/reporting/core/v4/rest/v4/reports/batchGet#ReportRequest.FIELDS.date_ranges
[4] [API explorer request for just one](https://developers.google.com/apis-explorer/?hl=en_GB#p/analyticsreporting/v4/analyticsreporting.reports.batchGet?_h=1&resource=%257B%250A++%2522reportRequests%2522%253A+%250A++%255B%250A++++%257B%250A++++++%2522viewId%2522%253A+%252212345678%2522%252C%250A++++++%2522filtersExpression%2522%253A+%2522ga%253ApagePath%253D%253D%252Fcheck-uk-visa%2522%252C%250A++++++%2522metrics%2522%253A+%250A++++++%255B%250A++++++++%257B%250A++++++++++%2522expression%2522%253A+%2522ga%253ApageViews%2522%250A++++++++%257D%250A++++++%255D%252C%250A++++++%2522dateRanges%2522%253A+%250A++++++%255B%250A++++++++%257B%250A++++++++++%2522startDate%2522%253A+%25227daysAgo%2522%252C%250A++++++++++%2522endDate%2522%253A+%2522today%2522%250A++++++++%257D%250A++++++%255D%250A++++%257D%250A++%255D%250A%257D&)
[...]

PR #45, opened 2017-01-12
Show the number of views per content item in the table

leenagupte:
Trello card: https://trello.com/c/a3I8aKal
Supersedes PR #36

## Expected changes

* Show the number of views per content item in the table on the organisations page

### Before

screen shot 2017-01-16 at 13 46 48

### After

screen shot 2017-01-16 at 13 45 36

## Deployment notes

* Add three new environment variables to the heroku config:
   * CPM_GOVUK_VIEW_ID
   * GOOGLE_PRIVATE_KEY
   * GOOGLE_CLIENT_EMAIL
* Finally import the number of page views for the organisation.


README.md
```diff
@@ -72,12 +78,43 @@ followed by:
...
+$ heroku rake import:all_organisations
+$ heroku rake import:content_items_by_organisation[{department-slug}]
+$ heroku rake import:number_of_views_by_organisation[{department-slug}]
 ```
 
+### Set up Google Analytics credentials in development
+
+To be able to import data from Google Analytics, a few environment variables need to be set.
+
+1) Create files `config/local_env.yml`
```
pmanrubia: I have created a [Trello card](https://trello.com/c/hdoeQ1zT/111-use-rails-configuration-for-environment-variables) so we explore two things:

1. Provide a `config/local_env.yml.copy` with the template including all environment variables and update the readme to advise the reader to copy to `config/local_env.yml` and fill the environment variables.
2. Use `Rails.configuration` object to access environment variables within the app.
3. Set a default value for CPM_GOVUK_VIEW_ID

README.md
```diff
@@ -72,12 +78,43 @@ followed by:
...
+
+N.B. This file has been added to `.gitignore`, so you don't need to worry about accidentally checking it in.
+
+`local_env.yml` is used to 
+* Create the environment variables required to connect to and query the Google API for Google Analytics 
+* Store your Google API credentials. They can either be downloaded from the [google developers console](https://console.developers.google.com/apis/credentials) or from another team member.
+
+2) Add an entry to `config/local_env.yml` for each of the following:
+
+* `CPM_GOVUK_VIEW_ID`
```
pmanrubia: Could you please add a separate commit where you rename, for consistency:

1.`CPM_GOVUK_VIEW_ID` to `GOOGLE_ANALYTICS_VIEW_ID`. 
2. `GOOGLE_PRIVATE_KEY` to `GOOGLE_ANALYTICS_PRIVATE_KEY`
3. `GOOGLE_PRIVATE_EMAIL` to `GOOGLE_ANALYTICS_CLIENT_EMAIL`

The key and the email are specific to Google Analytics and it would clash if we add other Google credentials for other APIs

app/models/importers/number_of_views_by_organisation.rb
```diff
@@ -0,0 +1,16 @@
+module Importers
+  class NumberOfViewsByOrganisation
+    def run(slug)
+      organisation = ::Organisation.find_by(slug: slug)
+
+      content_items = ::ContentItem.where(organisation_id: organisation.id)
+      base_paths = content_items.pluck(:base_path)
+
+      results = GoogleAnalyticsService.new.page_views(base_paths)
```
pmanrubia: I have created a ticket so we explore pagination here. We could have two type of issues with big organisations:

1. We `pluck` too many items from the database so we could have memory issues. I would advise using [find_each][1]
2. We need to investigate what is the limit in Google Analytics as, it is very likely, that there is an upper limit for the response.

No need to take an action now. Just pointing it out for the incoming ticket.

[1]: http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each

README.md
leenagupte: Sounds good. Though I think we should keep the "govuk" in the name, i.e. `GOOGLE_ANALYTICS_GOVUK_VIEW_ID`.

leenagupte: @pmanrubia I can rename `CPM_GOVUK_VIEW_ID` because this is an environment variable that we have created, but `GOOGLE_PRIVATE_KEY` and `GOOGLE_CLIENT_EMAIL` are used by the Google Client, and we have no control over that code.

leenagupte: Changed the name `CPM_GOVUK_VIEW_ID` in 19aeb12

spec/services/google_analytics/requests/page_views_request_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+                ],
+                date_ranges: [
+                  {
+                    start_date: "7daysAgo",
+                    end_date: "today"
+                  }
+                ]
+              }
+            ]
+          }.with_indifferent_access
```
leenagupte: Done in commit e45b646

PR #48, opened 2017-01-16
Set up Jenkins

leenagupte:
The continuous integration environment is switching to Jenkins 2.
As content-performance-manager is a brand new app we should start by
creating a properly configured `Jenkinsfile`, rather than just one that
points to the old style `jenkins.sh`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,52 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'content-performance-manager'
+
+node {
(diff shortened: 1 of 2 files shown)


pmanrubia:
The existing jenkins file is not being used in Jenkins1, so it is good to remove it.
Wouldn't it be better to merge it once it is tested with the new infrastructure?. Let's create another ticket for it, as we need to be sure that this is working properly. Thoughts?

leenagupte:
@pmanrubia I spoke with Alex in infrastructure yesterday. He asked that I created the PRs first before sitting with them, and then we could check over everything together.

pmanrubia:
That makes sense 👍

PR #50, opened 2017-01-17
Show sort arrow always fix to the right hand side.

theleebriggs:
# Description
Quick fix for sortable header text + arrow.
When text wraps in sortable table headers the up/down arrow positioning was off, this correctly positions them on the righthand side of the header.

# Before
before

# Now
now


pmanrubia:
👍

PR #51, opened 2017-01-17
Set up Errbit error reporting

leenagupte:
Trello card: https://trello.com/c/cXXXWGXG

Using an older version of the airbrake because after version 5, the
view tests fail with the following error:

```
Failure/Error: <%= sort_table_header "Title", "title" %>
     
ActionView::Template::Error:
  undefined method `sort_table_header' for #<#<Class:0x007fe658b743f0>:0x007fe6508420b8>
```

The Airbrake gem was re-written when it moved to version 5 and split
into two gems: `airbrake` and `airbrake-ruby`. Some investigation will
need to be done to determine why version 5 is not compatible with the
`sort_table_header` used to order the columns in the table on the
content items page.

Using config recommended in: https://github.gds/pages/gds/opsmanual/infrastructure/howto/setting-up-new-rails-app.html#configuring-errbit

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,6 +40,7 @@ gem 'httparty'
 gem 'kaminari', '~> 0.17.0'
 gem 'google-api-client', '~> 0.9'
 
+gem 'airbrake', '4.3.1'
 
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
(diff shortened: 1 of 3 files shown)


leenagupte:
We are talking about using [Sentry](https://sentry.io) instead of Errbit for error reporting, so I am closing this PR.

PR #53, opened 2017-01-18
Handle content store 404s

theleebriggs:
# Trello
https://trello.com/c/wyB5V6m0/108-a-content-item-on-gov-uk-is-causing-import-task-import-content-items-by-organisation-to-fail

# Description
Some content items found using the search api may not have a corresponding entry in the content store API - these will return a 404 which in turn will stop the import of an organisations content items at that point.

This PR adds 404 logic to the content store service as well as an overhaul of the http request mocking in the test suite, adding in WebMock rather than hijacking HTTParty. By doing this the tests remain http client agnostic.

### DIFF
--- a/app/services/clients/content_store.rb
+++ b/app/services/clients/content_store.rb
@@ -3,9 +3,13 @@ module Clients
     class << self
       def find(path, attributes)
         response = HTTParty.get(end_point(path))
-        content_item = JSON.parse(response.body).symbolize_keys
-
-        content_item.slice(*attributes)
+        case response.code
+        when 200
+          content_item = JSON.parse(response.body).symbolize_keys
+          content_item.slice(*attributes)
+        when 404
+          nil
+        end
       end
 
     private
--- a/app/services/content_items_service.rb
+++ b/app/services/content_items_service.rb
@@ -7,8 +7,8 @@ class ContentItemsService
 
     Clients::SearchAPI.find_each(query: query, fields: fields) do |response|
       base_path = response.fetch(:link)
-
-      yield Clients::ContentStore.find(base_path, attribute_names)
+      content_item = Clients::ContentStore.find(base_path, attribute_names)
+      yield content_item if content_item
     end
   end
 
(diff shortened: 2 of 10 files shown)


app/services/clients/content_store.rb
```diff
@@ -3,9 +3,13 @@ class ContentStore
...
         response = HTTParty.get(end_point(path))
-        content_item = JSON.parse(response.body).symbolize_keys
-
-        content_item.slice(*attributes)
+        case response.code
+          when 200
+            content_item = JSON.parse(response.body).symbolize_keys
+            content_item.slice(*attributes)
+          when 404
+            false
```
pmanrubia: I would prefer to return `nil` here. It feels a [bit more idiomatic to Ruby](https://ruby-doc.org/core-2.2.3/Enumerable.html#method-i-find).

app/services/content_items_service.rb
```diff
@@ -8,7 +8,10 @@ def find_each(organisation_slug)
     Clients::SearchAPI.find_each(query: query, fields: fields) do |response|
       base_path = response.fetch(:link)
 
-      yield Clients::ContentStore.find(base_path, attribute_names)
+      item = Clients::ContentStore.find(base_path, attribute_names)
+      if item
```
pmanrubia: what about:

```ruby
content_item = Clients::ContentStore.find(base_path, attribute_names)

yield content_item if content_item
```

spec/services/content_items_service_spec.rb
```diff
@@ -26,6 +26,16 @@
       expect(result).to match_array([:a])
     end
 
+    it "doesn't yield empty content items" do
```
pmanrubia: This is a good test 👍  Would you mind updating the copy to something that make it explicit that it is not yielding `nil` responses from the `ContentStore`

app/services/clients/content_store.rb
theleebriggs: Addressed now.

app/services/content_items_service.rb
theleebriggs: Addressed now.

spec/services/content_items_service_spec.rb
theleebriggs: Changed to "does not yield nil responses from the content store".

PR #55, opened 2017-01-19
Move gems to test group

leenagupte:
Jenkins does not install any gems that are in the `development` group, so the
master branch was failing to build because the linter and listen gems were
missing.

Moving these two gems has fixed the problem.


leenagupte:
Thanks @ikennaokpala, I've updated the commit note.

PR #59, opened 2017-01-23
Update instructions

leenagupte:
Add/Update Instructions for:

* Starting the app in the dev VM
* Deploying the application
* Running rake tasks


leenagupte:
@chao-xian Thanks for your comments. I've addressed them in commits 0685e2f and 61e6ef0

PR #60, opened 2017-02-03
Move WebMock into test group

chao-xian:
The WebMock gem was also being used in the development group but should only be available for test.


chao-xian:
@ikennaokpala thanks for your comment I'll make that PR next

PR #63, opened 2017-02-06
ManytoMany: ContentItems<=>Organisations

ikennaokpala:
[Trello card](https://trello.com/c/pLg019CM/109-5-create-a-many-to-many-association-between-contentitem-and-organisations)

## Description 

The import fails when a content item appears to belong to more than one
organisations.

This commit seeks to re-defines the association between ContenItems and the
respective Organisations to a ManytoMany association.

This allows for content items to be associated with the relevant
organisations (i.e more than one organisation).

As a result of this re-definition of association some tests have failed,
requiring the updates to factory girl’s definition of organisation and
also similar updates to some tests and concrete code.

I have updated factory girl organisation_with_content_items. This has been
done to fix failing tests and provide insight on the nature of the relationship
between content item and organisation from a factory girl point.

This references [1].

[1] http://www.rubydoc.info/gems/factory_girl/file/GETTING_STARTED.md#Associations

### DIFF
--- a/app/models/importers/content_items_by_organisation.rb
+++ b/app/models/importers/content_items_by_organisation.rb
@@ -11,9 +11,10 @@ module Importers
 
     def create_or_update!(attributes, organisation)
       content_id = attributes.fetch(:content_id)
-      content_item = organisation.content_items.find_or_create_by(content_id: content_id)
+      content_item = ContentItem.find_or_create_by(content_id: content_id)
 
       content_item.update!(attributes)
+      content_item.organisations << organisation unless content_item.organisations.include?(organisation)
     end
   end
 end
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -1,5 +1,5 @@
 class ContentItem < ApplicationRecord
-  belongs_to :organisation
+  has_and_belongs_to_many :organisations
 
   def url
     "https://gov.uk#{base_path}"
--- a/app/models/importers/number_of_views_by_organisation.rb
+++ b/app/models/importers/number_of_views_by_organisation.rb
@@ -3,8 +3,7 @@ module Importers
     def run(slug)
       organisation = ::Organisation.find_by(slug: slug)
 
-      content_items = ::ContentItem.where(organisation_id: organisation.id)
-      base_paths = content_items.pluck(:base_path)
+      base_paths = organisation.content_items.pluck(:base_path)
 
       results = GoogleAnalyticsService.new.page_views(base_paths)
       results.each do |result|
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -1,3 +1,3 @@
 class Organisation < ApplicationRecord
-  has_many :content_items
+  has_and_belongs_to_many :content_items
 end
--- a/db/migrate/20170203113501_create_content_items_organisations_join_table.rb
+++ b/db/migrate/20170203113501_create_content_items_organisations_join_table.rb
@@ -1,0 +1,5 @@
+class CreateContentItemsOrganisationsJoinTable < ActiveRecord::Migration[5.0]
+  def change
+    create_join_table :content_items, :organisations
+  end
+end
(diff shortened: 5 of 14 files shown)


app/models/importers/content_items_by_organisation.rb
```diff
@@ -11,9 +11,14 @@ def run(slug)
...
-      content_item = organisation.content_items.find_or_create_by(content_id: content_id)
+      content_item = ::ContentItem.find_or_create_by(content_id: content_id)
 
-      content_item.update!(attributes)
+      content_item.update!(attributes.except(:content_id))
+
+      organisations = content_item.organisations
+      unless organisations.include?(organisation)
+        organisations << organisation
+      end
```
pmanrubia: Very minor: This reads to me a bit better. The `unless` clause is not crucial to understand what the method does. Inlining it feels ok.

```ruby
    def create_or_update!(attributes, organisation)
      content_id = attributes.fetch(:content_id)
      content_item = ContentItem.find_or_create_by(content_id: content_id)

      content_item.update!(attributes)
      content_item.organisations << organisation unless content_item.organisations.include?(organisation)
    end
```

db/migrate/20170203113625_remove_organisation_reference_from_content_items.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveOrganisationReferenceFromContentItems < ActiveRecord::Migration[5.0]
+  def change
+    remove_reference :content_items, :organisation, foreign_key: true
```
pmanrubia: I would create a migration to migrate the one-to-many foreign keys to the many-to-many table; if we don't do it, this PR will force all developers to populate all the data again. Same applies to the heroku app and staging.

```ruby
class MigrateManyToManyContentItemsOrganisation < ActiveRecord::Migration[5.0]
  class ContentItem < ActiveRecord::Base
    belongs_to :organisation
  end

  def change
  MigrateManyToManyContentItemsOrganisation::ContentItem.find_each do |content_item|
      content_item.organisations << content_item.organisation
    end
  end
end
```

This migration need to be inserted before deleting the foreign key.

ikennaokpala: We discussed this earlier on Friday and I think Hong didn't add this to the checklist..

Since the data is can be derived at this stage.. it appeared low risk to simple drop and regenerated data.

app/models/importers/content_items_by_organisation.rb
ikennaokpala: I found the former's width easier. Though this option has one less side effect.

app/models/importers/content_items_by_organisation.rb
```diff
@@ -11,9 +11,14 @@ def run(slug)
 
     def create_or_update!(attributes, organisation)
       content_id = attributes.fetch(:content_id)
-      content_item = organisation.content_items.find_or_create_by(content_id: content_id)
+      content_item = ::ContentItem.find_or_create_by(content_id: content_id)
```
ikennaokpala: Done

db/migrate/20170203113625_remove_organisation_reference_from_content_items.rb
pmanrubia: I would prefer not to make developers reload all the data unless it is needed, because:

- It takes a while to load all the data from scratch
- We would also need to do it in Heroku

If there is a way to prevent this from happening (I have checked the code above and it is working), do you have any particular reason for not addressing this comment?

ikennaokpala: I have addressed the comment.. I was only recall our previous conversatioin as the reason why this was omitted.

pmanrubia: Apologies @ikennaokpala. I thought it wasn't addressed.

pmanrubia:
Thank you for addressing all comments. LGTM

PR #67, opened 2017-02-09
Remove nested show for content item

chao-xian:
Since we reworked the model to be many to many between organisations and content items, the view for a single item within an organisation does not need to be at a nested route. As in:

`/organisations/hm-revenue-customs/content_items/30339`

can just be

`/content_items/30339`

as that content item might also belong to another org at:

`/organisations/public-health-england/content_items/30339`

This change also displays the multiple organisations that an item belongs to in the `content_items#show` view.

This change is also in preparation for https://trello.com/c/QWhNMeyr

### DIFF
--- a/app/views/content_items/show.html.erb
+++ b/app/views/content_items/show.html.erb
@@ -30,7 +30,9 @@
     </tr>
     <tr>
       <td>Organisation</td>
-      <td><%= @organisation.title %></td>
+      <td>
+        <%= stringify_organisations @content_item.organisations %>
+      </td>
     </tr>
   </tbody>
 </table>
--- a/spec/factories/content_items_factory.rb
+++ b/spec/factories/content_items_factory.rb
@@ -6,5 +6,14 @@ FactoryGirl.define do
     sequence(:document_type) { |index| "document_type-#{index}" }
     base_path "api/content/item/path"
     public_updated_at { Time.now }
+
+    factory :content_item_with_organisations do
+      transient do
+        organisations_count 1
+      end
+      after(:create) do |content_item, evaluator|
+        create_list(:organisation, evaluator.organisations_count, content_items: [content_item])
+      end
+    end
   end
 end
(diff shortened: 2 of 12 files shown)


chao-xian:
@pmanrubia addressed in https://github.com/alphagov/content-performance-manager/pull/67/commits/83b93becb8f26bd9815ac444b4536d2f30a22e5c

app/views/content_items/show.html.erb
```diff
@@ -30,7 +30,9 @@
     </tr>
     <tr>
       <td>Organisation</td>
-      <td><%= @organisation.title %></td>
+      <td>
+        <%= list_comma_separated_words @content_item.organisations.collect(&:title) %>
```
pmanrubia: What about something like:
```ruby
<%= render stringify_organisations(@content_item.organisations) %>
```

I find this approach more expressive, and the view is less dependent of implementation details.
What are your thoughts?

spec/factories/content_items_factory.rb
```diff
@@ -6,5 +6,14 @@
     sequence(:document_type) { |index| "document_type-#{index}" }
     base_path "api/content/item/path"
     public_updated_at { Time.now }
+
+    factory :content_items_with_org do
```
pmanrubia: I would rename the factory to: `content_item_with_organisations`

spec/factories/content_items_factory.rb
```diff
@@ -6,5 +6,14 @@
...
+      transient do
```
pmanrubia: Why are you choosing two organisations here? I would prefer to have only 1, if no big reason for 2.

spec/views/content_items/show.html.erb_spec.rb
```diff
@@ -1,12 +1,10 @@
 require 'rails_helper'
 
 RSpec.describe 'content_items/show.html.erb', type: :view do
-  let(:content_item) { build(:content_item) }
-  let(:organisation) { build(:organisation) }
+  let(:content_item) { create(:content_items_with_org, organisations_count: 2) }
```
pmanrubia: I would prefer to stick to 1 organisation per content item, and only use 2 when we are checking the way we render organisations. The tests would be more expressive, and we they will also perform better.

chao-xian:
@pmanrubia fixed up and rebased against master.

PR #72, opened 2017-02-16
Render all items

chao-xian:
[Trello card](https://trello.com/c/QWhNMeyr)

## Description

We need to be able to see content items without a filter. This adds a view and controller methods to do that, however we've also refactored to introduce a new page title helper and a ContentItemsQuery.

The unfiltered view is at URL:

```
<app_root>/content_items/
```

while the filtered by organisation view remains at:

```
<app_root>/<organisation-slug>/content_items
```

Both of these routes are mapped to `ContentItemsController#index`

This PR also depended on #67

### DIFF
--- a/app/helpers/content_items_helper.rb
+++ b/app/helpers/content_items_helper.rb
@@ -1,0 +1,9 @@
+module ContentItemsHelper
+  def content_items_header
+    if @organisation.present?
+      @organisation.title
+    else
+      'GOV.UK'
+    end
+  end
+end
--- a/app/helpers/organisations_helper.rb
+++ b/app/helpers/organisations_helper.rb
@@ -1,13 +1,9 @@
 module OrganisationsHelper
-  def stringify_organisations(organisations, options = {})
-    organisation_titles = if options[:linkify]
-                            organisations.collect do |organisation|
-                              link_to(organisation.title, organisation_content_items_path(organisation_slug: organisation.slug))
-                            end
-                          else
-                            organisations.collect(&:title)
-                          end
+  def stringify_organisations(organisations)
+    names = organisations.collect do |organisation|
+      link_to(organisation.title, organisation_content_items_path(organisation_slug: organisation.slug))
+    end
 
-    organisation_titles.join(', ').html_safe
+    names.join(', ').html_safe
   end
 end
--- a/app/helpers/organisations_helper.rb
+++ b/app/helpers/organisations_helper.rb
@@ -1,5 +1,13 @@
 module OrganisationsHelper
-  def stringify_organisations(organisations)
-    organisations.collect(&:title).join(', ').html_safe
+  def stringify_organisations(organisations, options = {})
+    organisation_titles = if options[:linkify]
+                            organisations.collect do |organisation|
+                              link_to(organisation.title, organisation_content_items_path(organisation_slug: organisation.slug))
+                            end
+                          else
+                            organisations.collect(&:title)
+                          end
+
+    organisation_titles.join(', ').html_safe
   end
 end
--- a/app/queries/content_items_query.rb
+++ b/app/queries/content_items_query.rb
@@ -1,0 +1,17 @@
+class ContentItemsQuery
+  def self.build(options)
+    ContentItemsQuery.new.results(options)
+  end
+
+  def results(options = {})
+    relation = if options[:organisation]
+                 options[:organisation].content_items
+               else
+                 ContentItem.all
+               end
+
+    relation
+      .order("#{options[:sort]} #{options[:order]}")
+      .page(options[:page])
+  end
+end
--- a/app/views/content_items/_content_item.html.erb
+++ b/app/views/content_items/_content_item.html.erb
@@ -1,5 +1,12 @@
 <tr>
   <td><%= link_to content_item.title, content_item_path(id: content_item.id) %></td>
+  <% if @organisation.blank? %>
+    <td>
+      <% content_item.organisations.each do |organisation| %>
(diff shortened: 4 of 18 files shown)


app/helpers/page_title_helper.rb
```diff
@@ -0,0 +1,9 @@
+module PageTitleHelper
```
pmanrubia: Could you please move this method to `ContentItemsHelper`?

app/helpers/page_title_helper.rb
```diff
@@ -0,0 +1,9 @@
...
+  def page_title
+    if @organisation.present?
+      @organisation.title
+    else
+      t('.title')
```
pmanrubia: I would not use internationalisation unless this is a requirement because would need to be consistent for every piece of text of our app. The whole process would imply more work with no clear user need.

app/queries/content_items_query.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def results(options = {})
+    relation = if options[:organisation]
+                 options[:organisation].content_items
+               else
+                 ContentItem.all
+               end
+
+    relation
+    .order("#{options[:sort]} #{options[:order]}")
```
pmanrubia: Missing 2 spaces here?

app/views/content_items/_content_item.html.erb
```diff
@@ -1,5 +1,12 @@
 <tr>
   <td><%= link_to content_item.title, content_item_path(id: content_item.id) %></td>
+  <% if @organisation.blank? %>
+    <td>
+      <% content_item.organisations.each do |organisation| %>
```
pmanrubia: would it be a good place to use the helper to stringify the organisations?

spec/controllers/content_items_controller_spec.rb
```diff
@@ -2,53 +2,47 @@
...
     end
 
-    context "sorts with parameters" do
-      let(:content_items) do
-        [
-          build(:content_item, id: 1, public_updated_at: Time.parse("2016-12-09")),
-          build(:content_item, id: 2, public_updated_at: Time.parse("2015-12-09"))
-        ]
+    context "find by organisation" do
+      let(:organisation) { create(:organisation_with_content_items, content_items_count: 2) }
```
pmanrubia: As you have extracted the query object, I would `mock` it.

spec/features/content_items_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+require 'features/pagination_spec_helper'
+
+RSpec.feature "Content Items List", type: :feature do
+  background do
+    @content_items = create_list :content_item_with_organisations, 3
+  end
+
+  scenario "the user sees a list of all unfiltered content items and clicks through to an organisation" do
```
pmanrubia: I would split this scenario into two:

1. to list all the content items
2. to click through an organisation

I will allow it to grow a bit easier in time.

spec/features/content_items_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    visit 'content_items'
+    click_on @content_items[0].organisations[0].title.to_s
```
pmanrubia: No need to use `to_s`

spec/queries/content_items_query_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.describe ContentItemsQuery, type: :query do
```
pmanrubia: Not sure if we are covering here the organisations filter.

app/helpers/organisations_helper.rb
```diff
@@ -1,5 +1,13 @@
 module OrganisationsHelper
-  def stringify_organisations(organisations)
-    organisations.collect(&:title).join(', ').html_safe
+  def stringify_organisations(organisations, options = {})
```
pmanrubia: Have you thought about reusing this function every time we have to render the organisations of a Content Item?
The code would be more readable.

app/helpers/content_items_helper.rb
```diff
@@ -0,0 +1,9 @@
+module ContentItemsHelper
+  def page_title
```
pmanrubia: As this helper is shared across all controllers I would rename it to something more specific, or I would configure Rails to not load all helpers in all views and be explicit at controller level.

spec/queries/content_items_query_spec.rb
chao-xian: Hopefully addressed in https://github.com/alphagov/content-performance-manager/pull/72/commits/c5cb55df5dcd077bcb5370b85f123aa568e28134

pmanrubia:
Good work @chao-xian  👍  

Two things we just need to bear in mind for incoming PRs:

1. Remove the dependency of `@organisations` from the `TableHelper`
2. Start using decorators and remove helpers

As discussed, these two things were out of the scope of this PR.

chao-xian:
@pmanrubia good point made by @theleebriggs that we shouldn't have the logic to add or remove the column showing an organisation for each content item when in filtered by an organisation, because an item may belong to other organisations and it would be useful to see that. He's also spotted that the URLs are inconsistent where we use the slug or an org which will be `spine-cased` but the rest of the url will use Rails convention of `snake_case` so we have URLs that look like `/organisations/slug-of-organisation/content_items/`

pmanrubia:
@chao-xian @theleebriggs thank you for the follow up. 
Not sure what you mean with your first point. 
Regarding the consistency of the URL... I don't see a bit problem with it but I might be missing something. Let's talk about it tomorrow once I am in.

PR #73, opened 2017-02-17
Import all content items

leenagupte:
Trello card: https://trello.com/c/PjAzxNQS

## Motivation

As the application is currently in production, we need to load all content items for all organisations.

This will allow us to:

1. Distribute the application to other users within GDS
2. Discover performance issues in advance

### DIFF
--- a/app/models/importers/all_content_items.rb
+++ b/app/models/importers/all_content_items.rb
@@ -1,0 +1,6 @@
+module Importers
+  class AllContentItems
+    def run
+    end
+  end
+end
--- a/app/models/importers/all_content_items.rb
+++ b/app/models/importers/all_content_items.rb
@@ -1,6 +1,11 @@
 module Importers
   class AllContentItems
     def run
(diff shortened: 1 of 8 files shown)


app/models/importers/all_content_items.rb
```diff
@@ -1,6 +1,12 @@
 module Importers
   class AllContentItems
     def run
+      organisations = ::Organisation.pluck(:slug)
```
pmanrubia: This is loading all organisations into memory.
I would go for something like:

```ruby
importer = ContentItemsByOrganisation.new
Organisation.find_each do |organisation| 
  importer.run(organisation.slug)
end
```

leenagupte: Thanks. I'll give it a go and make the change.

leenagupte: @pmanrubia Updated and fixed up in commit 5a48bb5

pmanrubia:
👍

PR #74, opened 2017-02-20
Use `guard` and `spring` to run RSpec tests

pmanrubia:
Include an improvement to run our tests with less disruptions:

* **Spring**: Spring is a Rails application preloader. It speeds up development by keeping your application running in the background so you don't need to boot it every time you run a test, rake task or migration. It is already installed, but there is no bin/stub for RSpec. 
* **Guard** is a command line tool to easily handle events on file system modifications. In CPM we will use it to run RSpec tests when the source code changes.

## Use

```terminal
$ guard
```


spec/rails_helper.rb
```diff
@@ -7,6 +7,8 @@
 require 'rspec/rails'
 require 'support/factory_girl'
 require 'webmock/rspec'
+require 'shoulda/matchers'
+
```
leenagupte: @pmanrubia Do we need the shoulda-matchers to use spring?

pmanrubia: It is explained in commit 4f249fb. Please, let me know if this is not clear enough.

> shoulda matchers is not required in the Gemfile, and it is required in
the `spec_helper` because the matchers were not identified otherwise.
It is the recommended way to get over this issue.

pmanrubia:
Thank you for your review @leenagupte

I have address your comments regarding the documentation in this commit d1661d9, and I have added a note asking if the existing explanation about `shoulda_matchers` and `spring-rspec` is clear enough.

spec/rails_helper.rb
leenagupte: Sorry completely missed that sentence.

pmanrubia:
Thank you @ikennaokpala, I have amended it.

Rebasing with master and force pushing in preparation to merge.

PR #76, opened 2017-02-22
Import Google Analytics data in batches

leenagupte:
Trello: https://trello.com/c/AuddFX1d

## Motivation

We are not using batches to retrieve ContentItems from the Database, so while we are able to retrieve GA data for small departments, large organisations like HMRC time out waiting for Google Analytics to respond.

## Changes 

Modify the import to request data from Google Analytics in batches of 3000.

### DIFF
--- a/spec/models/importers/number_of_views_by_organisation_spec.rb
+++ b/spec/models/importers/number_of_views_by_organisation_spec.rb
@@ -19,6 +19,9 @@ RSpec.describe Importers::NumberOfViewsByOrganisation do
           },
         ]
       )
+
+      stub_const("Importers::NumberOfViewsByOrganisation::BATCH_SIZE", 1)
+
       subject.run('the-slug')
 
       content_item_one = ContentItem.find_by(base_path: 'the-link/first')
@@ -28,10 +31,22 @@ RSpec.describe Importers::NumberOfViewsByOrganisation do
       expect(content_item_two.unique_page_views).to eq(2)
     end
 
-    it "perform the requests in batches" do
-      expect_any_instance_of(GoogleAnalyticsService).to receive(:page_views).twice.and_return([])
+    context "performs the request to google api in batches" do
+      it "makes two requests when the batch size is one" do
+        expect_any_instance_of(GoogleAnalyticsService).to receive(:page_views).twice.and_return([])
 
-      subject.run('the-slug')
+        stub_const("Importers::NumberOfViewsByOrganisation::BATCH_SIZE", 1)
+
+        subject.run('the-slug')
+      end
+
+      it "makes one request when the batch size is two" do
+        expect_any_instance_of(GoogleAnalyticsService).to receive(:page_views).once.and_return([])
+
+        stub_const("Importers::NumberOfViewsByOrganisation::BATCH_SIZE", 2)
+
+        subject.run('the-slug')
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/lib/tasks/import_spec.rb
```diff
@@ -53,9 +53,13 @@
     end
 
     it 'runs the number_of_views_by_organisation task' do
+      @cached_pageview_batch_size = ENV["GA_PAGEVIEW_BATCH_SIZE"]
```
chao-xian: Does this need to be `@` scoped?

leenagupte: @chao-xian No it doesn't. You caught my copy and paste error. Have fixed and fixed up.

pmanrubia:
@leenagupte the default value does not change per environment, so I am not sure we need to store it in an environment variable. What is the benefit you see?
I would get rid of 114c487

leenagupte:
@pmanrubia I had a similar, thought. Just updating the value of a constant works for me. 

I thought we discussed (and agreed on) putting the value in an environment variable though?

pmanrubia:
Hi @leenagupte, yeah, we had a quick discussion, I thought it was a good idea, but once I saw the PR I just thought that it might not be needed as this variable does not change per environment.
Sorry if I added noise to this PR, and thank you for addressing the changes.

leenagupte:
@pmanrubia That's cool. I'll update this one. I think using a constant will make [story 142](https://trello.com/c/jwrJjjhs/142-8-get-one-month-and-twelve-months-of-unique-page-view-data) easier, perhaps.

I'll make the changes and close the PR in puppet: https://github.com/alphagov/govuk-puppet/pull/5521.

theleebriggs:
Just looking at this in relation to https://trello.com/c/02lWowaO/157-prep-for-cpm-testing-on-1-march. The use of `ONE_MONTH` as the name for the batch size constant a bit confusing? As far as I can tell this constant has nothing to do with the date range we fetch, just the number of items we query for? So with this PR we would still be importing `Last 7 Days` but in batches of 3000? Maybe i have the wrong end of the stick.

theleebriggs:
Just had another look and i see we override the `7DaysAgo`. My bad.

leenagupte:
@theleebriggs I was thinking that later we could add another variable for `TWELVE_MONTHS` that gets the optimal batch size for 12 months worth of data and pass that to the `find_in_batches` function.

leenagupte:
@theleebriggs I've taken the suggestion you made in meat-space and renamed the batch size constant from `ONE_MONTH` to `BATCH_SIZE`.

spec/models/importers/number_of_views_by_organisation_spec.rb
```diff
@@ -5,8 +5,9 @@
     let!(:organisation) { create(:organisation, slug: 'the-slug') }
     let!(:content_item_first) { create(:content_item, base_path: 'the-link/first', organisations: [organisation]) }
     let!(:content_item_second) { create(:content_item, base_path: 'the-link/second', organisations: [organisation]) }
+    let!(:content_item_third) { create(:content_item, base_path: 'the-link/third', organisations: [organisation]) }
```
theleebriggs: Minor thing but not sure you need a third item to test with - I think two items and a batch of `1` followed by a batch of `2` would accomplish the same.

leenagupte: @theleebriggs That's true. I'll update. Thanks!

pmanrubia:
@theleebriggs @chao-xian feel free to merge if you are happy with it.

leenagupte:
I've squashed commits 2, 3 and 4, because whilst they showed how my thought processes for approaching the problem, ultimately they were not part of the solution, so it seems wrong to keep them in our history.

PR #77, opened 2017-02-27
Fix display for no last updated at

chao-xian:
[Trello card](https://trello.com/c/o48vQ12o)

Where we have content items that have never been published, we are
attempting to apply a helper to nil values in the view, which is causing
server errors. To fix this we need to check in the view before attempting
to output anything for that field.

Includes a test that replicates the bug.

### DIFF
--- a/app/views/content_items/_content_item.html.erb
+++ b/app/views/content_items/_content_item.html.erb
@@ -7,5 +7,11 @@
   <% end %>
   <td><%= content_item.document_type %></td>
   <td><%= content_item.unique_page_views %></td>
-  <td><%= time_ago_in_words(content_item.public_updated_at) %> ago</td>
+  <td>
+    <% if content_item.public_updated_at %>
+      <%= time_ago_in_words(content_item.public_updated_at) %> ago
+    <% else %>
+      Never
+    <% end %>
+  </td>
 </tr>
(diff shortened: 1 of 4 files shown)


app/views/content_items/_content_item.html.erb
```diff
@@ -7,5 +7,9 @@
   <% end %>
   <td><%= content_item.document_type %></td>
   <td><%= content_item.unique_page_views %></td>
-  <td><%= time_ago_in_words(content_item.public_updated_at) %> ago</td>
+  <td>
+    <% if content_item.public_updated_at %>
+    <%= time_ago_in_words(content_item.public_updated_at) %> ago
+    <% end %>
```
theleebriggs: Some default text e.g never might be a good idea here, so we don't leave empty table cells.

spec/views/content_items/index.html.erb_spec.rb
```diff
@@ -114,6 +114,15 @@
 
         expect(rendered).to have_link('Last Updated', href: href)
       end
+
+      context 'items that have never been published' do
+        it 'renders for no last updated value' do
+          content_items[0].public_updated_at = nil
+          render
+
+          expect(rendered).to have_selector('table tr th:nth(4)')
```
theleebriggs: See previous comment then this could change to `never` as a selector.

chao-xian:
Thanks @theleebriggs! Good suggestion. Have applied and fixed up.

PR #81, opened 2017-03-01
Count the number of PDFs attached to a content item

theleebriggs:
### Motivation
Going forward it will be useful to not only highlight content items with large numbers of PDFs attached, but also give a sense of the total number of PDFs found across gov.uk

### Description
This PR grabs an additional field (`details`) from the content store - this field contains chunks of html representing attached documents, these chunks can be parsed to count the number of pdf documents found. This figure is then stored in each `ContentItem` under a new field `number_of_pdfs`.

### Before
screen shot 2017-03-01 at 17 44 04

### After
screen shot 2017-03-01 at 17 43 23

### DIFF
--- a/app/models/content_item.rb
+++ b/app/models/content_item.rb
@@ -4,4 +4,14 @@ class ContentItem < ApplicationRecord
   def url
     "https://gov.uk#{base_path}"
   end
+
+  def self.create_or_update!(attributes, organisation)
+    content_id = attributes.fetch(:content_id)
+    content_item = self.find_or_create_by(content_id: content_id)
+
+    attributes = attributes.slice(*content_item.attributes.symbolize_keys.keys)
+
+    content_item.update!(attributes)
+    content_item.organisations << organisation unless content_item.organisations.include?(organisation)
+  end
 end
--- a/app/services/metrics/number_of_pdfs_metric.rb
+++ b/app/services/metrics/number_of_pdfs_metric.rb
@@ -2,11 +2,40 @@ module Metrics
   class NumberOfPdfsMetric
     attr_accessor :content_item
 
+    PDF_XPATH = "//*[contains(@class, 'attachment-details')]//a[contains(@href, '.pdf')]".freeze
+
     def initialize(content_item)
       @content_item = content_item
     end
 
     def run
+      documents_string = extract_documents
+      documents = parse documents_string
+      { number_of_pdfs: number_of_pdfs(documents) }
+    end
+
+  private
+
+    def extract_documents
+      if content_item[:details].is_a?(Hash)
+        details = content_item[:details].symbolize_keys
+        documents = document_keys.map do |k|
+          details.fetch(k, nil)
+        end
+        documents.join('')
+      end
+    end
+
+    def document_keys
+      %i(documents final_outcome_documents)
+    end
+
+    def parse(string)
+      Nokogiri::HTML(string)
+    end
+
+    def number_of_pdfs(html_fragment)
+      html_fragment.xpath(PDF_XPATH).length
     end
   end
 end
--- a/app/services/metrics/number_of_pdfs_metric.rb
+++ b/app/services/metrics/number_of_pdfs_metric.rb
@@ -1,0 +1,12 @@
+module Metrics
+  class NumberOfPdfsMetric
+    attr_accessor :content_item
+
+    def initialize(content_item)
+      @content_item = content_item
+    end
+
+    def run
+    end
+  end
+end
--- a/spec/models/content_item_spec.rb
+++ b/spec/models/content_item_spec.rb
@@ -1,12 +1,45 @@
-require 'rails_helper'
+require "rails_helper"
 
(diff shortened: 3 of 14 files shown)


app/models/content_item.rb
```diff
@@ -4,4 +4,14 @@ class ContentItem < ApplicationRecord
   def url
     "https://gov.uk#{base_path}"
   end
+
+  def self.create_or_update!(attributes, organisation)
+    content_id = attributes.fetch(:content_id)
+    content_item = self.find_or_create_by(content_id: content_id)
+    # byebug
```
chao-xian: Errant byebug

app/services/metrics/number_of_pdfs_metric.rb
```diff
@@ -0,0 +1,38 @@
...
+      details = @content_item[:details].symbolize_keys
+      documents = []
+      document_keys.each do |k|
+        documents.push(details.fetch(k)) if details.key?(k)
+      end
+      documents.join('')
+    end
+
+    def document_keys
+      %i(documents final_outcome_documents)
```
leenagupte: Is it possible for two document keys to have the same attachment? Would you get a more accurate number if the links to the PDFs were extracted into a Set in `number_of_pdfs` and then counted?

spec/services/metric_builder_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+      expect_any_instance_of(Metrics::NumberOfPdfsMetric).to receive(:run).exactly(1).times.and_return(metrics)
+
+      subject.run_all({})
+    end
+
+    it "returns a merged hash of metrics" do
+      allow_any_instance_of(Metrics::NumberOfPdfsMetric).to receive(:run).and_return(metrics)
+      result = subject.run_all({})
+
+      expect(result).to eq(metrics)
```
leenagupte: I'm not sure what this test is doing. The return value from `Metrics::NumberOfPdfsMetric` is being mocked, so it's not necessary to test the value.

spec/models/content_item_spec.rb
```diff
@@ -1,12 +1,48 @@
...
+    before do
+      create(:content_item, content_id: "first_id")
+    end
+
+    it "creates a content item if it does not exist" do
+      expect { ContentItem.create_or_update!(content_item_2, organisation) }.to change { ContentItem.count }.by(1)
+    end
+
+    it "updates a content item if it already exists" do
+      expect { ContentItem.create_or_update!(content_item_1, organisation) }.to change { ContentItem.count }.by(0)
```
leenagupte: I would expect this test to check whether a value had been updated. You could replicate the original test `it 'updates the attributes'` (that is removed in commit 7c7ca5c63e4c4909d0c419b9e55dc66b72fc2f8a) that checks the value of the title.

spec/models/importers/content_items_by_organisation_spec.rb
```diff
@@ -3,45 +3,27 @@
...
 
-        attributes = ContentItem.find_by(base_path: 'the-link-value').attributes.symbolize_keys
-        expect(attributes).to include(title: 'the-title')
-      end
+      expect(content_item.reload.number_of_pdfs).to eq(10)
     end
 
-    context 'when the content item already exists' do
-      let(:content_item) { create(:content_item, base_path: 'the-link', organisations: [organisation]) }
+    it 'creates or update all the content items' do
```
leenagupte: I would change the name of the test to make it clear what it's actually doing. An existing content item is being updated and a new content item is being created, so something like"
`it 'creates a new content item and updates an existing one in the same import`

leenagupte:
I forget to mention, the final result looks really good, it's just the journey that needs a bit of explanation.

app/services/metrics/number_of_pdfs_metric.rb
theleebriggs: You're right, we would get a more accurate representation of unique pdfs attached to a content item, but i don't think that's something we care about; this is about pdf heavy content, so whether unique or not it still should still take them into account.

app/models/content_item.rb
theleebriggs: Removed this now.

spec/services/metric_builder_spec.rb
theleebriggs: Yep - agreed, removed this check now.

spec/models/content_item_spec.rb
theleebriggs: Updated this test to check for a changed attribute.

spec/models/importers/content_items_by_organisation_spec.rb
theleebriggs: Renamed as per suggestion.

theleebriggs:
@leenagupte I've reworked the commits into more of the flow you suggested, hopefully this makes it more readable.

chao-xian:
@ikennaokpala has approved but can't merge whereas I can so I'm clicking the button. Nice one @theleebriggs!

PR #87, opened 2017-03-10
Use decorators with Draper and remove helpers

pmanrubia:
As a way to have more manageable views we are extracting presentation logic into decorators (presenters)

We have decided to use Draper as a way to facilitate the code to support our decorators.

paired: @leenagupte @ikennaokpala @chao-xian @theleebriggs


pmanrubia:
No need to explicitly approve as this PR has been done by more than 2 developers.

PR #90, opened 2017-03-14
Add 6 months unique page views

chao-xian:

### Before
image
image

### After
image
image


chao-xian:
@pmanrubia the order of the commits in this PR is wrong but correct if you look locally. Also note: https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/

pmanrubia:
@chao-xian thank you for the links above.
What about superseding this PR with a new one with the commits in order?

chao-xian:
Superseded by #91

PR #91, opened 2017-03-15
Add 1 month and 6 months UPVs

chao-xian:
Supersedes #90

### Before
image
image

### After
image
image

### DIFF
--- a/app/services/google_analytics/requests/page_views_request.rb
+++ b/app/services/google_analytics/requests/page_views_request.rb
@@ -6,15 +6,16 @@ module GoogleAnalytics
     class PageViewsRequest
       include Google::Apis::AnalyticsreportingV4
 
-      def build(base_paths, start_dates: ["7daysAgo"], end_date: "today")
+      def build(args = {})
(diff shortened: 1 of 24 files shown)


app/services/google_analytics/requests/page_views_request.rb
```diff
@@ -6,27 +6,30 @@ module Requests
     class PageViewsRequest
       include Google::Apis::AnalyticsreportingV4
 
-      def build(base_paths, start_date: "7daysAgo", end_date: "today")
+      def build(base_paths, start_dates: ["7daysAgo"], end_date: "today")
```
pmanrubia: What about removing the defaults for `start_dates`?. It feels a bit misleading when you read the method signature. It seems that we are not using it in our codebase.

chao-xian:
@pmanrubia I've finally got a PR with the correct commit history order: https://github.com/alphagov/content-performance-manager/pull/93

So closing this. Sorry for the faff. Your comments won't be on the new PR though.

PR #92, opened 2017-03-15
Case insensitive search

theleebriggs:
### Motivation

The first iteration of the search by title functionality was case sensitive, ideally users should not have to know the capitalisation of the content item titles to search for them.

### Description
Simple change to downcase the search term and the model field when querying.

### DIFF
--- a/app/queries/content_items_query.rb
+++ b/app/queries/content_items_query.rb
@@ -9,7 +9,7 @@ class ContentItemsQuery
                else
                  ContentItem.all
                end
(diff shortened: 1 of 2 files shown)


theleebriggs:
@ikennaokpala i've implemented your suggestion, best to add less code at the moment as like you say this is certainly not the final search solution 👍 . Happy for me to merge?

ikennaokpala:
@theleebriggs either way it is all fine..

PR #93, opened 2017-03-16
Six months pageviews

chao-xian:
Supersedes #91

### Before
image
image

### After
image
image


chao-xian:
This PR will depend on a PR for govuk-puppet being merged in https://github.com/alphagov/govuk-puppet/pull/5637

chao-xian:
Closing as we should be using https://github.com/alphagov/govuk_sidekiq instead of `delayed_job`. Will re-open once we've switched.

PR #96, opened 2017-03-20
Add Endpoints to show and create Groups

pmanrubia:
[Trello card](https://trello.com/c/OiKWmsiP/194-8-api-endpoint-to-create-a-group-of-content-items)

### Goal:

Add API endpoints to support basic operations on Groups.

### Groups

* A group **is a set of content IDs** with no predefined order with the following properties: `slug`, `name`, `group_type`, `parent_group_id`

* Groups have a **tree structure**, with no limit in the number of descendants.
* Groups **slugs are uniq**
* The set of content IDs is stored as an Array of the group. This approach could have drawbacks in the long run if we need to include complex logic, but the approach is so easy to change, that it seems a better approach to keep things simple, and not make modelling decisions yet. 
* I am not creating a many-to-many association with ContentItem as the ContentItem model is will no be persisted in the long run as we will rely on the [Publishing API](https://github.com/alphagov/publishing-api).
* Groups are hierarchical, so we can assign a parent to any group.
* Authentication is done via the environment variable ‘CONTENT-PERFORMNCE-MANAGER-TOKEN’  and the URL parameter `api-token`. In incoming sprints we will use the GOV.UK SignOn to generate and validate the requests.

### API Samples

* Create a group

```terminal
curl -X "POST" "http://content-performance-manager.dev.gov.uk/groups.json" \
     -H "Content-Type: application/json" \
     -d $'{
  "group": {
    "slug": "the-slug",
    "group_type": "the-group-type",
    "name": "the-name",
    "content_item_ids": [
      "content-id-1",
      "content-id-2"
    ],
    "parent_group_slug": "the-parent-slug"
  },
  "api_token": "1234567"
}'
```

* Show a group

```terminal
curl "http://content-performance-manager.dev.gov.uk/groups/3.json?api_token=1234567" \
     -H "Content-Type: application/json"
```

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -1,0 +1,7 @@
+class Group < ApplicationRecord
+  validates :slug, uniqueness: true
+  validates :slug, :name, :group_type, presence: true
+
+  belongs_to :parent, class_name: 'Group', foreign_key: :parent_group_id, optional: true
+  has_many :children, class_name: 'Group', foreign_key: :parent_group_id
+end
--- a/doc/api.md
+++ b/doc/api.md
@@ -1,0 +1,44 @@
+# Set up API
+
+To be able to use the API a few environment variables need to be set.
+
+1) Create file `config/local_env.yml` (if the file does not exist yet)
+
+2) Add an entry to `config/local_env.yml` for: `CONTENT-PERFORMANCE-MANAGER-TOKEN`
+
+e.g.
+
+```bash
+CONTENT-PERFORMANCE-MANAGER-TOKEN: "1234567"
+```
+
+Use this the token value in any API request
+
+# Samples
+
+* Create a group
+
+```terminal
+curl -X "POST" "http://content-performance-manager.dev.gov.uk/groups.json" \
+     -H "Content-Type: application/json" \
+     -d $'{
+  "group": {
+    "slug": "the-slug",
+    "group_type": "the-group-type",
+    "name": "the-name",
+    "content_item_ids": [
+      "content-id-1",
+      "content-id-2"
+    ],
+    "parent_group_slug": "firstname49"
+  },
+  "api_token": "1234567"
+}'
+```
+
+* Show a group
+
+```terminal
+curl "http://content-performance-manager.dev.gov.uk/groups/3.json?api_token=1234567" \
+     -H "Content-Type: application/json"
+```
--- a/doc/api.md
+++ b/doc/api.md
@@ -39,6 +39,6 @@ curl -X "POST" "http://content-performance-manager.dev.gov.uk/groups.json" \
 * Show a group
 
 ```terminal
-curl "http://content-performance-manager.dev.gov.uk/groups/3.json?api_token=1234567" \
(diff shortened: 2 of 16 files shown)


app/models/group.rb
```diff
@@ -0,0 +1,7 @@
+class Group < ApplicationRecord
+  validates :slug, uniqueness: true
+  validates :slug, :name, :group_type, presence: true
+
+  belongs_to :parent, class_name: 'Group', foreign_key: :parent_group_id, optional: true
+  has_many :children, class_name: 'Group', foreign_key: :parent_group_id
```
pmanrubia: Not sure if I am fully understanding your question here. In the `has_many` association there is no need to set up `optional: true` as the collection could be empty (leaf node)
Optional is only needed in the `belongs_to` as by default is not true?

spec/controllers/groups_controller_spec.rb
```diff
@@ -9,4 +9,37 @@
...
+        }.to change(Group, :count).by(1)
+      end
+
+      it "assigns a newly created group as @group" do
+        post :create, format: :json, params: { group: valid_attributes }
+        expect(assigns(:group)).to be_a(Group)
+        expect(assigns(:group)).to be_persisted
+      end
+
+      it "redirects to the created group" do
```
pmanrubia: Thank you, I have updated it to

```ruby
it "redirects to the created group" do
  post :create, format: :json, params: { group: valid_attributes, api_token: 'a-token' }
  expect(response.status).to eq(201)
end
```

doc/api.md
```diff
@@ -0,0 +1,44 @@
+# Set up API
+
+To be able to the API a few environment variables need to be set.
```
pmanrubia: Updated in efcb1dc

spec/controllers/groups_controller_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe GroupsController, type: :controller do
+  describe "GET #show" do
+    it "assigns the requested user as @user" do
```
pmanrubia: Updated in 880001a

pmanrubia:
@chao-xian thank you for your review.
I have addressed your comments. Please, let me know if you are happy with them.

pmanrubia:
@chao-xian I have added the `slug` attribute to the `show` method in 8efdc8a

spec/controllers/groups_controller_spec.rb
```diff
@@ -21,8 +21,8 @@
 
   describe "GET #show" do
     it "assigns the requested user as @group" do
```
chao-xian: I'm really sorry I didn't spot the word `user` here. I'm not sure if this is fixed in one of the other 2 PRs (which is confusing).

pmanrubia: Sorry Kelvin, I messed up things with my rebase. Things should be fine now

chao-xian:
🎉

pmanrubia:
Thank you for the review @chao-xian.
I have rebased with master and force pushed in preparation for merging.

PR #100, opened 2017-03-20
Add misc folder to CPM

pmanrubia:
[Trello Card](https://trello.com/c/i8iZuvPU/200-add-misc-folder-to-cpm)

We have a few classes that do not fit very well in the folder structure that we currently have. Unfortunately, as have no common pattern for them yet it is a bit difficult to place them
It is a common practice to create a `misc` folder to locate those files and move them to a different place once patterns start to arise.


theleebriggs:
LGTM 👍

PR #101, opened 2017-03-21
Import organisations using the publishing api

theleebriggs:
### Motivation
We already import taxonomies using the publishing api, this is another step toward unifying the services we use when importing data. 
https://trello.com/c/Ds2NC2Yb/187-use-publishing-api-to-retrieve-organisations

### Description
Previously `organisations` were imported using the `search-api`, this has now been switched to use the `publishing-api`.

As part of this work the publishing api client in the app was expanded to be more flexible with the queries it can make.

### DIFF
--- a/app/services/clients/publishing_api.rb
+++ b/app/services/clients/publishing_api.rb
@@ -13,20 +13,16 @@ module Clients
       @per_page = 100
     end
 
-    def find_each(fields)
+    def find_each(fields, options = {})
       current_page = 1
-      loop do
-        response = publishing_api.get_content_items(
-          document_type: 'taxon',
-          order: '-public_updated_at',
-          q: '',
-          page: current_page,
-          per_page: per_page,
-          states: ['published'],
-        )
 
-        taxonomies = map_taxon_results(response['results'], fields)
-        taxonomies.each { |taxon| yield taxon }
+      query = build_base_query(fields, options)
+
+      loop do
+        query = build_current_page_query(query, current_page)
+        response = publishing_api.get_content_items(query)
+        response["results"].each { |result| yield result.symbolize_keys }
+
         break if last_page?(response)
         current_page = response["current_page"] + 1
       end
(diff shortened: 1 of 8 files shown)


app/services/organisations_service.rb
```diff
@@ -1,12 +1,36 @@
...
-      yield attributes
+  def query_options
+    @options ||= { document_type: 'organisation' }
+  end
+
+  def query_fields
+    @fields ||= %i(base_path title)
+  end
+
+  def map_slug(organisation)
```
pmanrubia: What about renaming it to `build_slug`?

app/services/clients/publishing_api.rb
```diff
@@ -13,33 +13,42 @@ def initialize
       @per_page = 100
     end
 
-    def find_each(fields)
+    def find_each(fields, options = {})
       current_page = 1
+
+      query = build_query(fields, options)
```
pmanrubia: What about renaming it to: `build_base_query`?

app/services/clients/publishing_api.rb
```diff
@@ -13,33 +13,42 @@ def initialize
...
-          order: '-public_updated_at',
-          q: '',
-          page: current_page,
-          per_page: per_page,
-          states: ['published'],
-        )
-
-        taxonomies = map_taxon_results(response['results'], fields)
-        taxonomies.each { |taxon| yield taxon }
+        query = page_query(query, current_page)
```
pmanrubia: What about renaming it to: `build_current_page_query`?

app/services/clients/publishing_api.rb
```diff
@@ -13,33 +13,42 @@ def initialize
...
       end
     end
 
   private
 
-    def last_page?(response)
-      response["pages"] == response["current_page"]
+    def build_query(fields, options)
+      {
+        document_type: options[:document_type] || 'taxon',
```
pmanrubia: I would remove the default; it might be better to fail than to return unexpected information. What do you think?

app/services/clients/publishing_api.rb
```diff
@@ -13,33 +13,42 @@ def initialize
...
+        states: ['published'],
+        per_page: per_page,
+        fields: fields || []
+      }
     end
 
-    def map_taxon_results(taxon_results, fields)
-      taxon_results.map { |taxon_hash| taxon_hash.symbolize_keys.slice(*fields) }
+    def page_query(query, page)
+      query.tap do |q|
```
pmanrubia: Why not doing this?

```ruby
query[:page]=page
```

app/services/organisations_service.rb
theleebriggs: Addressed now.

app/services/clients/publishing_api.rb
theleebriggs: Addressed now.

app/services/clients/publishing_api.rb
theleebriggs: Addressed now.

app/services/clients/publishing_api.rb
theleebriggs: Agreed - removed now.

app/services/clients/publishing_api.rb
theleebriggs: Done.

theleebriggs:
@pmanrubia address you comments now. :)

pmanrubia:
Great work @theleebriggs 👍 👍

PR #102, opened 2017-03-22
Use publishing api to retrieve content items

theleebriggs:
### Motivation
The `content_items` import part of the application is now the only part still using a combination of `search-api` and `content-store`. 

It is desirable for our import services to be as unified in their approach as possible, in addition to this the publishing api contains a richer set of information e.g. revision history.
https://trello.com/c/VBqWEPj8/188-8-use-publishing-api-to-retrieve-the-content-items

### Description
Replaces calls to `content_store` and `search_api`. Instead content is retrieved by `document_type` from the `publishing-api`.

The document types are configured in a .yml file for now, this will change once there is an endpoint on the `publishing-api` to retrieve a live list.

When importing `content-items` the associated organisations and taxonomies are retrieved using the publishing api's `expanded_links` endpoint.

Calls to get/import content by organisation have been removed along with their respective code.


app/models/importers/all_content_items.rb
```diff
@@ -1,10 +1,20 @@
...
+    attr_accessor :metric_builder, :content_items_service
+
+    def initialize
+      @metric_builder = MetricBuilder.new
+      @content_items_service = ContentItemsService.new
+    end
+
     def run
-      content_item_importer = ContentItemsByOrganisation.new
+      document_types = Rails.configuration.document_types
```
pmanrubia: I would create `attr_writer` with the document_types so it can be updated in the test, and I would create a reader this way:

```ruby
def document_types
  @document_types ||= Rails.configuration.document_types
```

app/services/content_items_service.rb
```diff
@@ -1,23 +1,27 @@
...
+    raise 'missing block!' unless block_given?
 
-    Clients::SearchAPI.find_each(query: query, fields: fields) do |response|
-      base_path = response.fetch(:link)
-      content_item = Clients::ContentStore.find(base_path, attribute_names)
-      if content_item
-        content_item[:taxons] = TaxonomyParser.parse(content_item)
-        yield content_item
-      end
+    publishing_api.find_each(query_fields, build_query_options(document_type)) do |content_item|
```
pmanrubia: what about extracting a variable:

```ruby
query_options = { document_type: document_type, expand_links: true }
```

app/services/clients/publishing_api.rb
```diff
@@ -15,19 +15,28 @@ def initialize
...
 
       loop do
         query = build_current_page_query(query, current_page)
         response = publishing_api.get_content_items(query)
-        response["results"].each { |result| yield result.symbolize_keys }
+        response["results"].each do |result|
+          if options[:expand_links]
+            result[:expanded_links] = expand_links(result["content_id"])
+          end
+          yield result.symbolize_keys
```
theleebriggs: Updated now.

spec/services/clients/publishing_api_spec.rb
```diff
@@ -3,39 +3,62 @@
 RSpec.describe Clients::PublishingAPI do
   subject { Clients::PublishingAPI.new }
 
-  it "loops through paged results from the gds publishing api" do
-    subject.publishing_api = double
-    page_1 = { "results" => [{}], "pages" => 2, "current_page" => 1 }
-    page_2 = { "results" => [{}], "pages" => 2, "current_page" => 2 }
+  describe "#find_each" do
```
theleebriggs: Agreed, will keep this in mind for next PR.

spec/services/clients/publishing_api_spec.rb
```diff
@@ -3,39 +3,62 @@
...
-    expect(subject.publishing_api).to receive(:get_content_items).exactly(2).times.and_return(page_1, page_2)
+      expect(subject.publishing_api).to receive(:get_content_items).exactly(2).times.and_return(page_1, page_2)
 
-    subject.find_each([]) {}
-  end
+      subject.find_each([]) {}
+    end
+
+    it "yields the results" do
+      yielded = []
```
theleebriggs: Updated now.

spec/services/clients/publishing_api_spec.rb
```diff
@@ -3,39 +3,62 @@
...
-  it "queries with the passed in options" do
-    subject.publishing_api = double
-    expected_query = {
-      document_type: 'a_document_type',
-      order: 'a_order',
-      q: 'a_search_term',
-      fields: [:field_1, :field_2]
-    }
-    result = { "results" => [{}], "pages" => 1, "current_page" => 1 }
+  describe "#expand_links" do
```
theleebriggs: Is now a private method.

app/services/content_items_service.rb
theleebriggs: As discussed, keeping this as a function for now.

theleebriggs:
Thanks for the review @pmanrubia - I've addressed your comments now - both the code changes and adding more details to the commit notes on getting `links` from the publishing api.

I've also switched from using `expanded_links` to just `links` to speed up the import. This means there are 2 additional commits to add `content_id` to the org table (1608429551a51bbccf1a1188b81a4c59c040806b and 6cf5a09b541f11a5d2d65128a9b854b2d91e8b2d) - this enables us to associate orgs with content items by content_id - the only field returned by the `/links/CONTENT_ID` endpoint.

I have some speed test figures to run past you as well - maybe before this is merged?

Cheers.

app/models/importers/all_content_items.rb
theleebriggs: Implemented as you suggested now.

pmanrubia:
LGTM 👍

PR #103, opened 2017-03-22
Update url parameter to use organisation

pmanrubia:
[Trello card](https://trello.com/c/kIgXEGsM)

### Goal

Change the Query parameter from `organisation_slug` to `organisation_id`

### Why

Content Performance Manager is moving towards searching Content 
Items using the Search API. As part of this process, it seems sensible 
to mimic the Search API query parameters so we can bypass them 
and get the most of the existing work.

This is a PR to prepare our codebase to the transition from using our
local database and migrate to Search API. We are first adding the 
`content_id` attribute to each organisation locally, as this is going to
be the attribute name that is going to be reused in the Search API.

### DIFF
--- a/app/views/organisations/_organisation.html.erb
+++ b/app/views/organisations/_organisation.html.erb
@@ -1,4 +1,4 @@
 <tr>
-  <td><%= link_to organisation.title, content_items_path(organisation_slug: organisation.slug) %></td>
+  <td><%= link_to organisation.title, content_items_path(organisation_id: organisation.content_id) %></td>
   <td><%= organisation.content_items.count %></td>
 </tr>
--- a/spec/controllers/content_items_controller_spec.rb
+++ b/spec/controllers/content_items_controller_spec.rb
@@ -22,11 +22,11 @@ RSpec.describe ContentItemsController, type: :controller do
       get :index, params: expected_params
     end
 
-    it "assigns the organisation provided the slug" do
-      create(:organisation, slug: 'the-slug')
+    it "assigns the organisation provided the content_id" do
+      create(:organisation, content_id: 'the-organisation-id')
 
-      get :index, params: { organisation_slug: 'the-slug' }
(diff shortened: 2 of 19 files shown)


pmanrubia:
@theleebriggs I have rebased on top of #102 as suggested.

app/views/organisations/_organisation.html.erb
```diff
@@ -1,4 +1,4 @@
 <tr>
-  <td><%= link_to organisation.title, content_items_path(organisation_slug: organisation.slug) %></td>
+  <td><%= link_to organisation.title, content_items_path(organisation_id: organisation.content_id) %></td>
   <td><%= organisation.content_items.count %></td>
-</tr>
+</tr>
```
theleebriggs: TIny one - missing newline.

spec/controllers/content_items_controller_spec.rb
```diff
@@ -48,7 +48,7 @@
       let(:organisation) { create(:organisation_with_content_items, content_items_count: 1) }
 
       before do
-        get :show, params: { organisation_slug: organisation.slug, id: organisation.content_items.first.id }
+        get :show, params: { organisation_id: organisation.content_id, id: organisation.content_items.first.id }
```
theleebriggs: Not really to do with your work but this test for #show doesn't need organisation at all - just needs a content item.

pmanrubia: Addressed in 9fe8a68a03c40057f2cceb9128c76a335cb6796a

app/views/organisations/_organisation.html.erb
pmanrubia: Addressed in 9fe8a68a03c40057f2cceb9128c76a335cb6796a

pmanrubia:
Thank you for your review. I have addressed your comments.

theleebriggs:
Merge away 👍

PR #106, opened 2017-03-24
Fix broken heading for filtering by taxonomy

chao-xian:
A recent switch to using content ids for taxonomies affected the
decorator that determined which taxonomy was being used to filter by.

This commit fixes that bug.

### DIFF
--- a/spec/decorators/content_items_decorator_spec.rb
+++ b/spec/decorators/content_items_decorator_spec.rb
@@ -4,7 +4,7 @@ RSpec.describe ContentItemsDecorator, type: :decorator do
   include Draper::ViewHelpers
 
   let(:organisation) { create(:organisation, title: "Organisation title", slug: "the-slug") }
-  let(:taxonomy) { create(:taxonomy, title: "taxonomy a") }
+  let(:taxonomy) { create(:taxonomy, title: "taxonomy a", content_id: "content-id-for-taxon123") }
 
   context "without filter params" do
     it "renders the default page title" do
@@ -28,7 +28,7 @@ RSpec.describe ContentItemsDecorator, type: :decorator do
 
   context "with taxonomy filter" do
     it "renders the taxonomy title as page title" do
-      allow(helpers).to receive(:params).and_return(taxonomy: "taxonomy a")
+      allow(helpers).to receive(:params).and_return(taxonomy_content_id: "content-id-for-taxon123")
       content_items = [build(:content_item, taxonomies: [taxonomy])]
       subject = ContentItemsDecorator.new(content_items)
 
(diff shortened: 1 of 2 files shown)


pmanrubia:
@chao-xian thank you for your work on this
Can you think of a test that would have prevented us from this bug?

chao-xian:
@pmanrubia the problem was that we were mocking the helpers in the tests before. This change https://github.com/alphagov/content-performance-manager/pull/106/commits/2976ed031c5f4e9e73edb3e2beac83cf479ce51c should address it now.

spec/decorators/content_items_decorator_spec.rb
```diff
@@ -18,31 +18,28 @@
 
   context "with organisation filter" do
     it "renders the organisation title as page title" do
-      allow(helpers).to receive(:params).and_return(organisation_slug: "the-slug")
       content_items = [build(:content_item, organisations: [organisation])]
```
theleebriggs: None of the tests need `content_items` anymore, the decorator can be instantiated with an empty collection and still pass the tests e.g. `subject = ContentItemsDecorator.new([])`

spec/decorators/content_items_decorator_spec.rb
```diff
@@ -4,7 +4,7 @@
   include Draper::ViewHelpers
 
   let(:organisation) { create(:organisation, title: "Organisation title", slug: "the-slug") }
-  let(:taxonomy) { create(:taxonomy, title: "taxonomy a") }
+  let(:taxonomy) { create(:taxonomy, title: "taxonomy a", content_id: "content-id-for-taxon123") }
 
   context "without filter params" do
     it "renders the default page title" do
```
theleebriggs: Can remove `allow(helpers)..` here and the import of `include Draper::ViewHelpers` as they're no longer used.

spec/decorators/content_items_decorator_spec.rb
```diff
@@ -4,7 +4,7 @@
   include Draper::ViewHelpers
 
   let(:organisation) { create(:organisation, title: "Organisation title", slug: "the-slug") }
-  let(:taxonomy) { create(:taxonomy, title: "taxonomy a") }
+  let(:taxonomy) { create(:taxonomy, title: "taxonomy a", content_id: "content-id-for-taxon123") }
```
theleebriggs: These can be changed to `build()` now as we don't exercise the DB anymore - and we only need the `title` populated.

theleebriggs:
Cheers for this @chao-xian - all looks great, my only comments were about removing some redundancy in the test now that the decorator is simplified.

PR #109, opened 2017-03-30
Total live pages for view set of content items

theleebriggs:
### Motivation
As part of ongoing design iteration we are breaking up the apps main view into a few areas - one of these is a `summary` area containing aggregated metrics for the currently viewed set of content items. This PR adds a new metric "total pages" which is rendered in a new summary area above the table of content items seen at `/content_items`.

### Description
Extended the `ContentItemsQuery` class to provide both `paged` and `unpaged` results - the views use paged, the aggregated metrics use unpaged.

Added a partial for the `summary` area, this is where the aggregated collection metrics are rendered.

### Before
before

### After
after

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -8,6 +8,7 @@ class ContentItemsController < ApplicationController
   def index
     query = ContentItemsQuery.new(@query_options)
     @content_items = query.paginated_results.decorate
+    @metrics = MetricBuilder.new.run_collection(query.results)
   end
 
   def show
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -1,18 +1,13 @@
 class ContentItemsController < ApplicationController
   before_action :set_organisation, only: :index
   before_action :set_taxonomy, only: :index
+  before_action :set_query_options, only: :index
   before_action :set_all_organisations, only: :index
   before_action :set_all_taxonomies, only: :index
 
   def index
-    @content_items = ContentItemsQuery.build(
-      sort: params[:sort],
-      order: params[:order],
-      query: params[:query],
-      page: params[:page],
-      taxonomy: @taxonomy,
-      organisation: @organisation
-    ).decorate
+    query = ContentItemsQuery.new(@query_options)
+    @content_items = query.paginated_results.decorate
   end
 
   def show
@@ -20,6 +15,17 @@ class ContentItemsController < ApplicationController
   end
 
 private
+
+  def set_query_options
+    @query_options = {
+      sort: params[:sort],
+      order: params[:order],
+      query: params[:query],
+      page: params[:page],
+      taxonomy: @taxonomy,
+      organisation: @organisation
+    }
+  end
 
   def set_all_organisations
     @organisations = Organisation.order(:title)
--- a/app/misc/metric_builder.rb
+++ b/app/misc/metric_builder.rb
@@ -1,6 +1,6 @@
 class MetricBuilder
   def run_all(content_item_attributes)
-    run(all_metrics, content_item_attributes)
+    run(content_item_metrics, content_item_attributes)
   end
 
   def run_collection(content_items)
@@ -16,7 +16,7 @@ private
     metrics.reduce(:merge)
   end
 
-  def all_metrics
+  def content_item_metrics
     [
       Metrics::NumberOfPdfs
     ]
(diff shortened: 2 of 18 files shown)


spec/queries/content_items_query_spec.rb
```diff
@@ -2,92 +2,78 @@
...
-
-  context "search title" do
-    let!(:content_items) do
-      [
-        create(:content_item, title: 'title 1'),
-        create(:content_item, title: 'title 2')
-      ]
-    end
 
+  context "searching by title" do
```
pmanrubia: 👍 Nice one!

app/queries/content_items_query.rb
```diff
@@ -3,14 +3,21 @@ def self.build(options)
     ContentItemsQuery.new.results(options)
   end
 
-  def results(options = {})
+  def self.build_paged(options)
+    ContentItemsQuery.new.results(options, true)
+  end
+
+  def results(options = {}, should_page = false)
```
pmanrubia: I would rename this method to `build` just to be consistent with `build_paginated`.
On the other side, what about using named parameters this way?:

```ruby
def results(options = {}, should_page: false)
```

It will be benefitical as calling the method will read a bit better.

pmanrubia: English, like the force, is not very strong in me... but `should_page` does not sound familiar to me. Does it read well in English? To me it sounds better `should_paginate`, but this is basically your call.

spec/queries/content_items_query_spec.rb
```diff
@@ -76,4 +76,22 @@
       expect(results).to match_array(content_items)
     end
   end
+
+  describe "paging results" do
+    it "limits the number of items returned to the maximum per page" do
+      create_list(:content_item, 26)
```
pmanrubia: Creating a list of 26 items is a bit too much in terms of performance. I would prefer to modify the `Kaminary` page side in a `before` block, and then set the default in an `after` block.
There should some tests with a similar approach in our codebase.

app/misc/metric_builder.rb
```diff
@@ -1,16 +1,30 @@
...
 private
 
+  def run(metrics_to_run, data)
+    metrics = metrics_to_run.map do |metric_klass|
+      metric_klass.new(data).run
+    end
+    metrics.reduce(:merge)
+  end
+
   def all_metrics
```
pmanrubia: If we are adding the method `collection_metrics`, I would rename `all_metrics` to `content_item_metrics`? The only difference seems to be what is the target of the calculation.

As a side note, it would be good to remove the suffix in both metrics, as we have the class scoped. I think this is my bad, as I started using the suffix myself.

Feel free to add separate commits for these two things, if you are happy with my feedback.

app/misc/metrics/total_pages_metric.rb
```diff
@@ -6,7 +6,7 @@ def initialize(content_items)
       @content_items = content_items
     end
 
-    def build
+    def run
```
pmanrubia: 👍

app/controllers/content_items_controller.rb
```diff
@@ -7,6 +7,9 @@ class ContentItemsController < ApplicationController
 
   def index
     @content_items = ContentItemsQuery.build_paged(@query).decorate
+
+    unpaged_content_items = ContentItemsQuery.build(@query)
+    @metrics = MetricBuilder.new.run_collection(unpaged_content_items)
```
pmanrubia: I would love to be able to express this method this way:

```ruby
def index
  @content_items = ContentItemsQuery.build_paged(@query).decorate
  @collection_metrics = MetricBuilder.new.run_collection(@query)
end
```

Ideally the controller should not know what the `MetricBuilder` object needs. If we pass the query, when running all collection metrics, then at that point, we now that we should not paginate the results.

It might be that we need to move into something like:
```ruby
def index
  query = ContentItemsQuery.new(options)
  @content_items = query.build_paginated
  @metrics = MetricBuilder.new.run_collection(query)
end
```

Let's have a quick chat about it!

spec/queries/content_items_query_spec.rb
theleebriggs: Address now.

app/queries/content_items_query.rb
theleebriggs: Addressed now.

app/misc/metric_builder.rb
theleebriggs: Addressed now.

app/controllers/content_items_controller.rb
theleebriggs: Addressed now.

theleebriggs:
@pmanrubia I've addressed all your feedback now. Let me know if that's all ok :)

pmanrubia:
👍 👍

PR #116, opened 2017-05-08
Move view tests to feature tests

pmanrubia:
Some view tests were repeated in the feature tests, which we have now 
removed; this also indicated that the view tests should be removed 
altogether as the feature tests give us enough confidence in our test 
suite. 

Organisations views are not migrated as the whole feature will go away
shortly.

As a side effect, the feature tests are also more concise.

No changes affecting the user.

paired with @chao-xian


chao-xian:
Awesome!

PR #120, opened 2017-05-12
Add a default Rake task that runs rspec, then lints

tuzz:


chao-xian:
Nice on @tuzz! Could we update the `Guardfile` as well?

tuzz:
@chao-xian: added to Guardfile: https://github.com/alphagov/content-performance-manager/pull/120/commits/0c6db90d77a33b5b0fda498735acde6b95599169

I've checked that individual changes work as well as Guard's 'all' runner.

chao-xian:
@tuzz wow that's awesome! Thanks!

PR #121, opened 2017-05-12
Add support for scheduled jobs

cbaines:
Trello: https://trello.com/c/ZaOJUZZB/233-5-automate-data-collection-on-a-daily-basis


cbaines:
I'm going to go via a Jenkins approach, rather than using cron/whenever as it seems that its difficult to get the jobs to run once, and the deployment details leak in to the app (e.g. govuk_setenv in config/schedule.rb).

PR #128, opened 2017-05-16
Show/Hide filters

theleebriggs:
# Motivation

Currently the "useful" filters are `search term` and `organisation` as the `taxonomy` filter currently only applies to education content. To deal with showing additional filters while not overwhelming the user it would be useful to be able to by default hide filters perceived as "less useful".

Trello: https://trello.com/c/5rIJ0cg3/241-5-collapse-taxonomy-filter

# Description

This PR wraps the `taxonomy` filter in a collapsible `div` which can be toggled by clicking a link in the sidebar.

It also adds logic to the `ContentItemsController` to render the additional filters in an open state if the user is currently filtering by one of them.

New `gems` added to allow feature testing with `javascript`, namely `selenium` and `poltergeist` for headless browser support.

# Before
before

# After
after_1
after_2

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -18,3 +18,23 @@
 @import 'pagination';
 @import 'phase_banner';
 @import 'summary';
+
+// TODO: move any styles below to modules once more established
+
+[data-toggle] {
+
+  &[aria-expanded="true"] {
+
+    [data-close-label] {
+      display: inline;
+    }
+
+    [data-open-label] {
+      display: none;
+    }
+  }
+
+  [data-close-label] {
+    display: none;
+  }
+}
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -18,3 +18,23 @@
 @import 'pagination';
 @import 'phase_banner';
 @import 'summary';
+
+// TODO: move any styles below to modules once more established
```
pmanrubia: There is a pending TODO here

theleebriggs:
@pmanrubia I've rebased - due to the addition of SSO I've had to add the `DatabaseCleaner` gem to allow javascript enabled tests to have access to the `user` for the suite. Feel free to advise on a better way to configure it in the `rails_helper.rb` as you'd know best!

PR #129, opened 2017-05-16
Use signon for user authentication

cbaines:
Trello: https://trello.com/c/SEYqW8Cg/232-8-move-app-behind-signon

There are a couple of a dependencies, as there are a couple of new secrets that need to be specified as environment variables [1].

1: https://github.com/alphagov/content-performance-manager/blob/use-signon-for-user-authentication/config/initializers/gds-sso.rb#L3-L4

Dependencies:
 - [x] https://github.digital.cabinet-office.gov.uk/gds/deployment/pull/1326
 - [x] https://github.com/alphagov/govuk-puppet/pull/5917


cbaines:
I've updated the last commit to use factories for the users in the feature tests @pmanrubia .

PR #140, opened 2017-05-18
Add link model

cbaines:
Trello: https://trello.com/c/5l45X1eC/262-add-link-model


app/jobs/import_content_item_job.rb
```diff
@@ -12,9 +12,32 @@ def perform(*args)
     content_id = args[0]
 
     attributes = content_items_service.fetch(content_id)
-    attributes = attributes.merge(content_items_service.links(content_id))
+    links = content_items_service.links(content_id)
```
pmanrubia: What about if our Jobs have a similar structure as our `Rake` tasks. Just one line of code calling a method within an importer.

```ruby
class ImportContentItemJob < ApplicationJob
  def perform(*args)
    content_id = args[0]
    Importers::SingleContentItem.import(content_id)
  end
``` 

This way we will be able to call this method within the console, for example. I know existing code is not designed this way, but feels to me like a good way to go.

app/jobs/import_content_item_job.rb
```diff
@@ -12,9 +12,32 @@ def perform(*args)
...
+    attributes = attributes.merge(links)
     attributes = attributes.merge(metric_builder.run_all(attributes))
 
-    ContentItem.create_or_update!(attributes)
+    # Use a single transaction, so that the content change and links
+    # change is atomic with respect to other transactions.
+    ActiveRecord::Base.transaction do
+      ContentItem.create_or_update!(attributes)
+      create_links(content_id, links)
```
pmanrubia: I might consider going towards something along the lines of ContentItem

```ruby
    ActiveRecord::Base.transaction do
      ContentItem.create_or_update!(attributes)
      Link.create_or_update!(content_id, links)
    end
```

app/jobs/import_content_item_job.rb
cbaines: For testing this, I've been calling `ImportContentItemJob.perform` from the console, and I think that has been working fine.

app/jobs/import_content_item_job.rb
cbaines: I would normally consider create_or_update acting on an individual item in a table, where it has some identifying information to find the item to update with, and some attributes that need setting or updating.

Links don't really fit with this, as all the information is identifying, without any attributes. Another way of putting this, if you change any field, its a different link.

That's not to say that the code here wouldn't be better somewhere else, just that my preconceptions on what create_or_update don't match with what the create_links method currently does.

PR #141, opened 2017-05-18
Minor UI changes

pmanrubia:
As per discussion with Natalie Baron, a few updates that improve
how content designers perceive the CPM


pmanrubia:
Thank you @chao-xian, I have removed the header. 
@cbaines also mentioned that the details page will look better without the header
Thanks both.

PR #142, opened 2017-05-19
Audit questionnaire page

tuzz:


config/application.rb
```diff
@@ -22,5 +22,8 @@ class Application < Rails::Application
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
     config.active_job.queue_adapter = :sidekiq
+
+    config.autoload_paths += %W(#{config.root}/lib)
+    config.autoload_paths += %W(#{config.root}/app/models/question)
```
pmanrubia: It seems that this is needed because of `FactoryGirl`. What do you think about being explicit in the Factory with:

```ruby
require_relative '../../app/models/question/boolean_question'
require_relative '../../app/models/question/free_text_question'
```
And we don't autoload at the application level?

config/application.rb
```diff
@@ -22,5 +22,8 @@ class Application < Rails::Application
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
     config.active_job.queue_adapter = :sidekiq
+
+    config.autoload_paths += %W(#{config.root}/lib)
```
pmanrubia: If we add `lib` to the `autoload_path`, it might be needed to append it as well to `eager_load_path` (Rails 5 thing)

app/controllers/audits_controller.rb
```diff
@@ -0,0 +1,37 @@
+class AuditsController < ApplicationController
+  def show
+    audit.questions = audit.template.questions if audit.new_record?
+  end
+
+  def save
+    audit.user = current_user
+
+    if audit.update(audit_params)
+      flash.now.notice = "Audit saved successfully."
```
tuzz: I tried this initially, but we lose the user's pre-filled form data if we redirect. I don't think there's a way to preserve that when redirecting?

PR #154, opened 2017-05-30
Upgrade to rails 5.1 && Bundle update

pmanrubia:
The main reason behind this update is to fix a Vagrant reload issue
when working on the hosting machine. There were some timestamps
that were not properly calculated that prevented reloading from
working properly.

I am also removing all the version constraints from Gemfile. This seems
to me a preferred way to work with versioning: rely on application tests, 
Gemfile.lock and developer diligence.

I have also updated all gems to their latest version. All tests are passing
and the application is working properly on the VM and in Integration.


greysteil:
@pmanrubia - would you be up for running [Dependabot](https://dependabot.com) on this repo to automate dependency updates? It's already running on a bunch of [MoJ](https://github.com/ministryofjustice/prison-visits-2/pulls?q=is%3Apr+is%3Aclosed) repos. (Obviously totally free.)

PR #155, opened 2017-05-31
Add sidebar to filter by Audit Status

pmanrubia:
[Trello card](https://trello.com/c/iAv4iHSW/269-3-filter-content-items-by-audit-status)

>As a **Content Designer**
>I want to filter by Audit status
>So I can see what Content Items have been audited and what is left to audit

image

### Developer Note

Please, delete commit c0d0f45 as this is just a fixup of PR #154


app/domain/search.rb
```diff
@@ -38,13 +44,28 @@ def page=(value)
 
   def sort=(identifier)
     query.sort = identifier
+    raise_if_unrecognised_sort
```
tuzz: It might be more consistent to move this to `Query` with the other raises. What do you think?

app/domain/search/audit_filter.rb
```diff
@@ -0,0 +1,42 @@
+class Search
+  class AuditFilter
+    AUDIT_STATUSES_IDENTIFIERS = [
+      :audited,
+      :non_audited,
+    ].freeze
```
tuzz: I was thinking this could be a binary thing, e.g. `audited=true` / `audited=false` but thinking about it, this is probably better if we have more states like 'in progress'. 👍

app/domain/search/audit_filter.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    attr_accessor :status
+
+
+    def initialize(status)
+      self.status = status ? status.to_sym : nil
```
tuzz: Why do we need to handle `nil`? I was thinking the `nil` case would be handled by not calling the `audit_status=` method so that `filters` doesn't include any instance of `AuditFilter`. What do you think?

app/domain/search/audit_filter.rb
```diff
@@ -0,0 +1,42 @@
...
+    def raise_if_unrecognised_audit_status
+      if status && !AUDIT_STATUSES_IDENTIFIERS.include?(status)
+        raise AuditStatusError, "unrecognised audit status: #{status}"
+      end
+    end
+
+    def self.all_status
+      [
+        OpenStruct.new(id: 'audited', name: "Audited"),
+        OpenStruct.new(id: 'non_audited', name: "Non Audited"),
```
tuzz: Please can we avoid `OpenStruct`? I don't like that a simple typo can make its way all the way into production before someone realises there's a problem, e.g. `OpenStruct.new(foo: "foo").fooo => nil`. I think it's better if a `NoMethodError` is raise much earlier during development.

Perhaps we could pair on this?

app/domain/search/executor.rb
```diff
@@ -23,17 +23,7 @@ def execute
...
-        if filter.by_source?
-          nested = nested.where(source_content_id: filter.source_ids)
-          nested = nested.select("target_content_id as content_id")
-        else
-          nested = nested.where(target_content_id: filter.target_ids)
-          nested = nested.select("source_content_id as content_id")
-        end
-
-        self.scope = scope.where(content_id: nested)
+        self.scope = filter.apply(scope)
```
tuzz: 👍

app/domain/search/link_filter.rb
```diff
@@ -0,0 +1,36 @@
...
+    def apply(scope)
+      nested = Link.where(link_type: link_type)
+
+      if by_source?
+        nested = nested.where(source_content_id: source_ids)
+        nested = nested.select("target_content_id as content_id")
+      else
+        nested = nested.where(target_content_id: target_ids)
+        nested = nested.select("source_content_id as content_id")
+      end
```
tuzz: I think we could use polymorphism here and have two types of `LinkFilter` ? We could pair?

app/models/content_item.rb
```diff
@@ -43,9 +45,8 @@ def add_taxonomies_by_id(taxons)
 private
 
   def linked_content(link_type)
-    search = Search.new
-    search.filter_by(link_type: link_type, source_ids: content_id)
-    search.execute
-    search.content_items
+    links = Link.where(link_type: link_type, source_content_id: content_id)
+
+    ContentItem.where(content_id: links.select(:target_content_id))
```
tuzz: What's the reason for this change? I think it's nice that `Search` encapsulates the knowledge of how `links` relate to `content_items`. If we do this, then that knowledge is in two places. Maybe that's not a problem? What do you think?

If we do keep this, `#pluck` might be more efficient so we skip constructing AR objects for each link.

app/views/audits/_sidebar.html.erb
```diff
@@ -0,0 +1,20 @@
+<div class="col-sm-3 sidebar">
+  <h1>Filter</h1>
+  <%= form_tag audits_path, method: :get do %>
```
tuzz: 👍

spec/features/audit/filter_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    visit audits_path
+    select "Non Audited", from: "audit_status"
+
+    click_on "Filter"
+
+    expect(page).to_not have_content("Audited content item")
+    expect(page).to have_content("Non-audited content item")
+    expect(page).to have_select("audit_status", selected: "Non Audited")
+  end
+end
```
tuzz: Should we also check that the page number resets to 1?

Also, you've added `hidden_fields` for `sort` and `order`. I suppose we should check that the sort selection is preserved when filtering. Maybe we could add a pending test as a reminder to do this once we have other sorts?

app/domain/search.rb
pmanrubia: As discussed this was done in preparation to extract the `Sort`ing responsabilities into objects in the same way I have done for Filtering. The validation will be placed within the `Sort` object (see [Filter](https://github.com/alphagov/content-performance-manager/blob/d089ef2701d154735d4f4396f8c816fdcefce817/app/domain/search/audit_filter.rb#L14-L14))

app/domain/search/audit_filter.rb
pmanrubia: Good point!, let's pair so we decide where to put the validation: in the controller or in Search.

app/domain/search/audit_filter.rb
pmanrubia: Not a fan of `OpenStruct` either. In this case the test coverage is very good, so it is not very likely to have something failing in production. Let's pair to find a better abstraction.

app/domain/search/link_filter.rb
pmanrubia: The code is moved from our last pairing session, and I do agree that need some refactoring. Polymorphism sounds like a good plan to tackle it.

app/models/content_item.rb
pmanrubia: Happy to talk about it. But it feels a bit better to me if the Domain objects use the ActiveRecord queries.

spec/features/audit/filter_spec.rb
pmanrubia: Good spot. I will remove the `hidden_fields` (I copied them over from the CPM)

app/domain/search/link_filter.rb
tuzz: ok

spec/features/audit/filter_spec.rb
tuzz: I was thinking a test like:

The user clicks on page 2,
The user changes their filters,
The user sees the first page of results.

app/models/content_item.rb
```diff
@@ -1,6 +1,8 @@
 class ContentItem < ApplicationRecord
   has_and_belongs_to_many :organisations
   has_and_belongs_to_many :taxonomies
+  has_one :audit, primary_key: :content_id, foreign_key: :content_id
```
tuzz: Do we need this association? Should content have visibility of whether it's been audited?

My mental model is that auditing is an entirely separate, standalone concept. By adding this association, it feels like we're coupling content with auditing. Happy to discuss.

tuzz:
As discussed, I think we should merge this in its current state pending further discussions regarding the app's domain model and a process for resolving minor refactors.

pmanrubia:
@tuzz thank you for your review and feedback

I have rebased with `master` in preparation for merging

PR #159, opened 2017-06-01
Revert environment setup via local_env.yml

pmanrubia:
This was accidentally deleted in https://github.com/alphagov/content-performance-manager/pull/154


pmanrubia:
Merging as this is a revert of some code that should've not been deleted.
cc/ @tuzz @cbaines @chao-xian 
Let me know if you need to clarify anything, please

PR #161, opened 2017-06-02
Standardise naming of databases across dev, test and prod

whoojemaflip:
The data replication script was failing to update Rails for me. Because of a naming mismatch
between dev, test, and production, my locally checked out out application wasn't picking up on the replicated data.

screen shot 2017-06-02 at 11 20 59

This change will break your local dev and test environments.

Keep 'em ticking over by running the following SQL in a postgres console:
```
$ psql
ALTER DATABASE content-performance-manager_development RENAME TO content_performance_manager_development;
```

Your test database doesn't exist anymore either, so recreate it using
the Rails commands:
`$ RAILS_ENV=test bundle exec rake db:drop db:test:prepare`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,52 +1,6 @@
 #!/usr/bin/env groovy
 
-REPOSITORY = 'content-performance-manager'
-
 node {
   def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
-
-  try {
-    stage('Checkout') {
-      checkout scm
-    }
-
-    stage('Clean') {
-      govuk.cleanupGit()
-      govuk.mergeMasterBranch()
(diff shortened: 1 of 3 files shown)


tuzz:
The database doesn't exist on Jenkins, hence the branch build is failing. I'm not sure what's responsible for creating this. It looks like it's [this](https://github.com/alphagov/content-performance-manager/blob/master/Jenkinsfile#L28) but I don't understand why it's not basing that on the name in database.yml.

tijmenb:
I've made this use the default build script, which handles the database thing better (I don't remember how, but it works like the other apps).

whoojemaflip:
@tijmenb I think the previous build script was failing to set `RAILS_ENV=test` before `rake db:environment:set`

PR #164, opened 2017-06-06
Create a true many-to-many relationship for Groups and ContentItems

whoojemaflip:
By creating a join table we make it faster to query larger groups. 

This migration assumes there are no Groups as there are none in my recently replicated data. 
*It will totally squash any data that's in the Groups->content_item_ids field*

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -4,4 +4,6 @@ class Group < ApplicationRecord
 
   belongs_to :parent, class_name: 'Group', foreign_key: :parent_group_id, optional: true
   has_many :children, class_name: 'Group', foreign_key: :parent_group_id
+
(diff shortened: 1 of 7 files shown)


app/models/importers/pagelist_for_concept_mining.rb
```diff
@@ -0,0 +1 @@
+pagelist_for_concept_mining.rb
```
pmanrubia: Do we need this importer? It seems unrelated to this PR

whoojemaflip: ah good catch. thought I had stashed that.

whoojemaflip:
@pmanrubia done

PR #165, opened 2017-06-06
Add Logstasher to app config

whoojemaflip:
This logging configuration will allow our ELK logging stack to interpret the production logs from the application. It means better searching and filtering of logs in Kibana.

Tidies up the Gemfile along the way.


cbaines:
Lovely cleanup of the Gemfile Tom :+1:

I've spun this up locally to test. The log output in the non json log looks a bit less useful, I'm seeing:
`[5d791472-76d2-405c-9788-c8715684789f] Authenticating with gds_sso strategy`
rather than:
`I, [2017-06-06T06:51:48.547868 #4901]  INFO -- : [7dff8500-fda1-47cf-8ad3-954d5137db06] Processing by ContentItemsController#index as */*`

Do you know why this might be?

As for the json log, that looks like a good addition, and it looks like JSON, which is good enough for me :)

whoojemaflip:
@cbaines thanks for taking a look at the output. 

I was under the impression that setting `config.logstasher.suppress_app_log = true` in `config/environments/production.rb` would prevent Rails from writing anything to `logs/production.log`. 

It would be nice if this app only logged ELK compatible JSON in production from now on. 

I've reverted the log formatter change, since it makes no real difference either way but might be clouding this issue.

cbaines:
> I've reverted the log formatter change, since it makes no real difference either way but might be clouding this issue.

Ok, could you fixup the previous commit with this change, just for neatness? Apart from that, this looks good to me :+1:

PR #171, opened 2017-06-07
Add feature to generate terms

tijmenb:


app/controllers/taxonomy_todos_controller.rb
```diff
@@ -0,0 +1,27 @@
+class TaxonomyTodosController < ApplicationController
+  def redirect_to_first
+    taxonomy_project = TaxonomyProject.find(params[:taxonomy_project_id])
```
whoojemaflip: This feels like it should be a method on the TaxonomyProject controller

app/controllers/taxonomy_todos_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def show
+    taxonomy_todo = TaxonomyTodo.find(params[:id])
+    @todo_form = TaxonomyTodoForm.new(taxonomy_todo: taxonomy_todo)
+  end
+
+  def update
+    taxonomy_todo = TaxonomyTodo.find(params[:id])
+    todo_form = TaxonomyTodoForm.new(params.require(:taxonomy_todo_form).permit(:new_terms))
```
whoojemaflip: could this be wrapped in a `todo_form` method?

app/controllers/taxonomy_todos_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    todo_form.taxonomy_todo = taxonomy_todo
+    todo_form.user = current_user
+    todo_form.save
+
+    next_item = taxonomy_todo.taxonomy_project.taxonomy_todos.still_todo.first
```
whoojemaflip: this feels like it should be a method on the todo model. `TaxonomyTodo#next` or somesuch.

tijmenb: I've added it to the TaxonomyProject.

app/controllers/taxonomy_todos_controller.rb
tijmenb: Extracted the parameters, is that ok too?

app/controllers/taxonomy_todos_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    redirect_to taxonomy_project.taxonomy_todos.still_todo.first
+  end
+
+  def show
+    taxonomy_todo = TaxonomyTodo.find(params[:id])
+    @todo_form = TaxonomyTodoForm.new(taxonomy_todo: taxonomy_todo)
+  end
+
+  def update
+    taxonomy_todo = TaxonomyTodo.find(params[:id])
```
whoojemaflip: this could be a memoised method to save a few lines in this controller

app/controllers/taxonomy_todos_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    todo_form = TaxonomyTodoForm.new(todo_params)
+    todo_form.taxonomy_todo = taxonomy_todo
+    todo_form.user = current_user
+
+    # since there are no validations yet, we're fine with crashing here. If we
+    # add validations the errors should be shown on the `show` page.
+    todo_form.save!
```
whoojemaflip: How could this crash?

app/controllers/taxonomy_todos_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def show
+    taxonomy_todo = TaxonomyTodo.find(params[:id])
+    @todo_form = TaxonomyTodoForm.new(taxonomy_todo: taxonomy_todo)
+  end
+
+  def update
+    taxonomy_todo = TaxonomyTodo.find(params[:id])
+    todo_form = TaxonomyTodoForm.new(todo_params)
```
whoojemaflip: could this class use a todo form builder method?

app/models/content_item.rb
```diff
@@ -1,6 +1,7 @@
 class ContentItem < ApplicationRecord
   has_and_belongs_to_many :organisations
   has_and_belongs_to_many :taxons
+  has_and_belongs_to_many :terms
```
whoojemaflip: Does the ContentItem have terms, or is it the TaxonomyTodo? Feels like the latter to me, and we could avoid change to core classes.

app/controllers/taxonomy_todos_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    todo_form.taxonomy_todo = taxonomy_todo
+    todo_form.user = current_user
+
+    # since there are no validations yet, we're fine with crashing here. If we
+    # add validations the errors should be shown on the `show` page.
+    todo_form.save!
+
+    next_item = taxonomy_todo.taxonomy_project.next_todo
```
whoojemaflip: this is still too long. `taxonomy_todo.next` could delegate to it's project's implementation?

app/controllers/taxonomy_todos_controller.rb
whoojemaflip: Also the method's not implemented

app/models/taxonomy_todo.rb
```diff
@@ -1,4 +1,11 @@
 class TaxonomyTodo < ApplicationRecord
   belongs_to :content_item
   belongs_to :taxonomy_project
+  has_many :terms, through: :content_item
+
+  scope :still_todo, -> { where(completed_at: nil) }
```
whoojemaflip: `incomplete`?

app/views/taxonomy_projects/show.html.erb
```diff
@@ -1,15 +1,38 @@
...
     <td>Title</td>
   </tr>
   </thead>
   <tbody>
-    <% @project.content_items.each do |content_item| %>
+    <% @project.taxonomy_todos.each do |taxonomy_todo| %>
       <tr>
-        <td><%= link_to content_item.title, content_item %></td>
+        <td><%= link_to taxonomy_todo.content_item.title, taxonomy_todo %></td>
+        <td><%= taxonomy_todo.terms.map(&:name).join(', ') %></td>
```
whoojemaflip: Should the `taxonomy_todo` here have a method to return the list of term names? Something like `taxonomy_todo.term_names`?

app/views/taxonomy_todos/show.html.erb
```diff
@@ -0,0 +1,77 @@
...
+    </div>
+
+    <div id="guidance">
+      <h4>Guidance</h4>
+      <p><%= @todo_form.content_item.guidance %></p>
+    </div>
+
+    <div id="topics">
+      <h4>Topics</h4>
+      <p><%= @todo_form.content_item.topics %></p>
```
whoojemaflip: Maybe the `todo_form` should delegate all these methods

app/controllers/taxonomy_todos_controller.rb
whoojemaflip: it definitely helps

app/models/taxonomy_todo.rb
```diff
@@ -1,4 +1,11 @@
 class TaxonomyTodo < ApplicationRecord
   belongs_to :content_item
```
whoojemaflip: Should a Todo behave like a ContentItem with Terms that belongs to a Project?

Maybe it should transparently delegate to all of ContentItem's property getters?

PR #177, opened 2017-06-08
Tune the `targets_of` query

tuzz:
The previous query was slow because it needed to
sort the results before it could apply the group
by operation. This approach is about 4x faster.

The audit listing page now loads in ~5 seconds, which is still slow. Perhaps this can be improved further.


tuzz:
Going to merge this because it's faster than it was. We can do more work later to speed it up if we need to.

PR #178, opened 2017-06-08
Inventory Manager (MVP)

tuzz:
https://trello.com/c/SL3csCd0/288-5-filter-by-sub-theme-first
https://trello.com/c/iXQWh5Ap/255-spike-can-we-import-filter-queries

This PR contains a work-in-progress inventory manager that we can use to set up the themes and subthemes based on the spreadsheet in the audit tool. Validation errors aren't handled currently. There's a separate ticket to address this:

https://trello.com/c/vfuNmaCD/160-handle-validation-errors-in-the-inventory-manager

After code review, I think it would be good to merge this so Ben A, Hollie, Chris, etc can play with it.


db/migrate/20170607101716_create_themes.rb
```diff
@@ -0,0 +1,10 @@
+class CreateThemes < ActiveRecord::Migration[5.1]
+  def change
```
pmanrubia: Have you considered using Groups/Subgroups for this?

tuzz: I hadn't. I forgot those existed.

Do we have a use-case other than Themes and Subthemes for grouping content items?

pmanrubia: Hi @tuzz, sorry I didn't talk to you about `Groups` before.

In the Audit Tool, we started thinking about Groups as a way to build a static list (Themes/Subthemes) of Content Items to be Audited by Content Designers. After a few conversations with Natalie, it felt that the CPM would also benefit from having groups of Content Items: Content Designers could create them to measure their performance over time.

Groups exist in the CPM as an API endpoint because the first idea we explored was to expose an API for other teams so they can build the themes via an external tool, but this plan was discarded after a few weeks.

In the Audit Tool, we are now adding filtering rules to Themes and Subthemes, and it feels to me that Themes and Subthemes are quite related to Groups/Subgroups, as I can't see any specific behaviour that justifies having separate Tables in the database. We could model them as domain objects, but it could be worth exploring reusing the Group/Subgroups infrastructure that is already in place.

My understanding is that Theme/Subthemes are a subtype of a Group. In the same way that the Term Generation team has spiked modelling with a Group Sample of Content Items to apply term generation. 

The current implementation of Groups will allow us to have a hierarchical structure (of many levels if needed) in case we want to split Sub-themes in the future, and also assigning individual Content Items to a Group which is also aligned with some user research from Natalie/Joe

Filtering would need to be done once (in the Query), although the `Search` API could perfectly speak of Themes and Subthemes.

There is a lot of work already done, so we could address this refactor in a future PR.

tuzz: Sounds good. I really like the idea of using `Group` as it would mean things could be more flexible in the future without us having to make structural changes to the database.

PR #186, opened 2017-06-15
Setup transport

tuzz:
Depends on #185


tuzz:
Good idea @chao-xian. Added.

PR #195, opened 2017-06-20
Enable RSpec feature to run only failing tests

pmanrubia:
I find useful being able to run only failing tests:

```bash
  rspec --only-failures
```

see: https://relishapp.com/rspec/rspec-core/docs/command-line/only-failures#running-`rspec---only-failures`-loads-only-spec-files-with-failures-and-runs-only-the-failures


tuzz:
Moved into `tmp/`

PR #196, opened 2017-06-21
Report page

tuzz:
- https://trello.com/c/KP0BnFzF/303-2-report-page-page-3-display-the-number-of-content-items-on-the-reports-page
- https://trello.com/c/c93bmgUk/304-3-report-page-page-3-display-the-audit-progress-on-the-reports-page
- https://trello.com/c/ZsET1yUA/305-report-page-page-3-display-the-number-of-items-passing-and-failing-on-the-reports-page

The 'export' link works but only exports a minimal number of fields. We'll address this in:
https://trello.com/c/SOUSMhFj/290-3-ability-to-export-csv-file-from-the-report-page

We also need to decide what to do with the the 'Audit status' filter, which doesn't make sense for this page. We'll address this separately, too.

screen shot 2017-06-21 at 14 53 28

### DIFF
--- a/app/domain/search.rb
+++ b/app/domain/search.rb
@@ -16,6 +16,8 @@ class Search
   DIMENSIONS = {
     audited: -> (s) { s.audit_status = :audited },
     not_audited: -> (s) { s.audit_status = :non_audited },
+    passing: -> (s) { s.passing = true },
+    not_passing: -> (s) { s.passing = false },
   }.freeze
 
   def initialize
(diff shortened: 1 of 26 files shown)


app/domain/search.rb
```diff
@@ -13,6 +13,11 @@ class Search
     Link::TOPICS,
   ].freeze
 
+  DIMENSIONS = {
```
pmanrubia: Love the dimensions 👍

tuzz: 😃

PR #197, opened 2017-06-22
Report CSV

tuzz:
As discussed with @pmanrubia, I added some code to precompute rows for the report so we can generate it much more quickly. This adds some complexity.

After merging we need to re-import all content items into the system so the `publishing_app` field is set. This should automatically populate the `report_rows` table.


tuzz:
Thanks, I've added more explanation to my final commit and amended it.

PR #198, opened 2017-06-22
Fix next link for audit show page

cbaines:
Trello: https://trello.com/c/B8YQWsfv/308-possible-bug-if-you-click-the-last-item-on-a-list-page-you-dont-get-a-next-link

### DIFF
--- a/app/domain/search/result.rb
+++ b/app/domain/search/result.rb
@@ -35,6 +35,11 @@ class Search
 
         if current_item_index.nil?
           limit *= 2
+          # Give up if the limit is greater than 500, as this process
+          # is probably taking too long, and might not even finish if
+          # somehow the content item provided has fallen out of the
+          # filter criteria
+          return false if limit > 500
           next
         elsif limit == (current_item_index + 1)
           limit += 1
--- a/app/domain/search/result.rb
+++ b/app/domain/search/result.rb
@@ -19,5 +19,30 @@ class Search
     def options_for(identifier)
       filter_options.fetch(identifier)
     end
+
+    def next_content_item(content_item)
+      next_scope =
+        sort.where_only_after(
+          content_items.unscope(:limit, :offset),
+          content_item
+        )
+
+      limit = 10
+
+      loop do
+        ids = next_scope.limit(limit).pluck(:id)
+        current_item_index = ids.index(content_item.id)
+
+        if current_item_index.nil?
+          limit *= 2
+          next
+        elsif limit == (current_item_index + 1)
+          limit += 1
+          next
+        else
+          return next_scope.find_by(id: ids[current_item_index + 1])
+        end
+      end
+    end
   end
 end
--- a/app/domain/search/result.rb
+++ b/app/domain/search/result.rb
@@ -1,10 +1,11 @@
 class Search
   class Result
-    attr_accessor :scope, :filter_options
+    attr_accessor :scope, :filter_options, :sort
 
-    def initialize(scope, filter_options)
+    def initialize(scope, filter_options, sort)
       self.scope = scope
       self.filter_options = filter_options
+      self.sort = sort
     end
 
     def content_items
--- a/app/controllers/audits_controller.rb
+++ b/app/controllers/audits_controller.rb
@@ -60,7 +60,7 @@ private
   end
 
   def next_item
-    @next_item ||= content_items.next_item(content_item)
+    @next_item ||= search.next_content_item(content_item)
   end
 
   def search
--- a/app/domain/search.rb
+++ b/app/domain/search.rb
@@ -85,6 +85,10 @@ class Search
     result.options_for(identifier)
   end
 
+  def next_content_item(content_item)
+    result.next_content_item(content_item)
+  end
+
   def dimension(identifier)
     @dimension ||= {}
 
--- a/app/domain/search/executor.rb
+++ b/app/domain/search/executor.rb
@@ -20,7 +20,7 @@ class Search
       apply_sort!
       paginate!
 
-      Result.new(scope, filter_options)
+      Result.new(scope, filter_options, query.sort)
     end
 
   private
--- a/app/domain/search/sort.rb
+++ b/app/domain/search/sort.rb
@@ -20,6 +20,13 @@ class Search
       scope.order("#{field} #{order}")
     end
 
+    def where_only_after(scope, content_item)
+      scope.where(
+        "#{field} #{{ desc: '<=', asc: '>=' }[order]} ?",
+        content_item.object.send(field)
+      )
+    end
+
   private
 
     attr_accessor :field, :order
--- a/app/domain/search/sort.rb
+++ b/app/domain/search/sort.rb
@@ -4,23 +4,24 @@ class Search
 
     def self.all
       [
-        Sort.new(:page_views_desc, "content_items.six_months_page_views desc"),
+        Sort.new(:page_views_desc, "content_items.six_months_page_views", :desc),
       ]
     end
 
     attr_reader :identifier
 
-    def initialize(identifier, sort_query)
+    def initialize(identifier, field, order)
       @identifier = identifier
-      @sort_query = sort_query
+      @field = field
+      @order = order
     end
 
     def apply(scope)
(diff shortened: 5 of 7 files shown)


app/domain/search/result.rb
```diff
@@ -14,5 +15,35 @@ def content_items
...
+          return false if limit > 500
+          next
+        elsif limit == (current_item_index + 1)
+          limit += 1
+          next
+        else
+          return next_scope.find_by(id: ids[current_item_index + 1])
+        end
+      end
+    end
```
tuzz: I'm a little concerned this approach is quite complicated. It might be difficult for someone new to the code to understand how this works. I wonder if we could extract this behaviour into something that's unit tested? Or maybe there's a simpler way to do this?

tuzz: I wonder if this code should live somewhere else, anyway. The `Result` object is really just supposed to be a value object that holds the results of executing a search. I'm not sure we should be running more queries here?

cbaines:
> I'm concerned we're introducing a lot of complexity to fix a small-ish bug. I'm not really sure why we can't use the existing next_item approach and just request the next page of results if the item can't be found?

Ok, I've tried thinking about this holistically today, as when I wrote this implementation back on Thursday, it was written as a general solution to the issue of the next link not appearing when the last audit on a page was viewed.

Thinking about this functionality, I believe it would be really valuable to defer the decision about what is next, so I've written up that as a separate change https://github.com/alphagov/content-performance-manager/pull/209 .

I've also extracted one commit from this branch to a separate commit, as I think its important and separate https://github.com/alphagov/content-performance-manager/pull/210 .

Back to the question of what to do about finding the next item, given the id of the current item, and the page it was on. Changing the `self.next_item` function to use a window one longer than the page size, and then finding the ContentItem directly is something I wrote initially, it looks something like: 

```
  def self.next_item(current_item)
    ids = all.limit(all.limit_value + 1).pluck(:id)
    index = ids.index(current_item.id)
    ContentItem.find_by(id: ids[index + 1]) if index
  end
```

With this patch, the problem moves to the behaviour on the first item of the 2nd page, or in the general case, the first item of any page, where it was reached from the previous page. This is because the page parameter will still be set to the previous page, and thus the offset and limit governing the ids selected in `self.next_item` won't contain the current content item.

I think it would be possible to solve this if this logic was moved to the controller. Then the controller could alter the page parameter if the next item is on the next page. I haven't tried implementing this, but I don't have a reason to believe it wouldn't work.

Thinking further ahead with this implementation, I'm unsure how well it would hold up around the audits at the start of a page when multiple users are using the system.

At the start of the page, if the currently audited item drops of the start of the page, then the Next item wouldn't be found. The most likely scenario I can think of where this would happen is where two users are using the system, one user starts at the start of the 1st page, and the second starts on the 2nd page. If both users filter by "Non audited", and the first user works faster than the second, the current item for the second user could fall in to the first page. Note that this involves users using the `next_item` functionality through the Save action, and not the Next action.

There is also a fringe case in the back of my mind where users return to the audit page for an item long after originally loading it, in which case the page parameter might not be correct.

I decided to select ids based on the value of the field that the results were sorted by, with the theory that this will be more resilient, as the next item if it exists is guaranteed to be in that part of the results, whereas with pages, it could now be on the page before (given the scenario I gave above), or possibly the page after (I'm having trouble coming up with a scenario for this).

One other final thought, is that we might be able to just increase the page size. There will literally and figuratively be less edge cases with larger pages, and this approach is quite simple to execute.

app/domain/search/result.rb
cbaines: > I wonder if this code should live somewhere else, anyway. The Result object is really just supposed to be a value object that holds the results of executing a search. I'm not sure we should be running more queries here?

Yeah, I agree, I put it here as it had access (once the sort was passed in) to all the necessary data, but it would probably be better in the Search, or in something that the Search class can delegate to.

cbaines:
> At the start of the page, if the currently audited item drops of the start of the page, then the Next item wouldn't be found. The most likely scenario I can think of where this would happen is where two users are using the system, one user starts at the start of the 1st page, and the second starts on the 2nd page. If both users filter by "Non audited", and the first user works faster than the second, the current item for the second user could fall in to the first page. Note that this involves users using the next_item functionality through the Save action, and not the Next action.

For clarity, this is the kind of scenario that I think its important to try and optimise for here, and not actually the "clicking the Next link" scenario, as hopefully people will be using the tool via saving audits, and being taken to the next item to audit, more than clicking Next over and over again.

tuzz:
You've clearly put a lot of thought into this 👍 

I had initially thought handling this in the controller and setting the `page` parameter to the next page might be the way to go, but I think you've highlighted other problems.

When we started the MVP, we tried to reduce scope by deferring the handling of concurrent users until later. We thought about ideas such as:

- Pre-allocating content items to specific users
- Showing a notification to users if they're on the same audit page (websockets?)
- Jumping to random items in the list instead of the next item

We decided this needed more thought and was too much to tackle this quarter.

I think this bug fix is in danger of snowballing into a potential fix for this concurrency problem. It's good to put thought into this, but I'm worried the scope will grow substantially as we discover more scenarios and edge-cases that aren't handled.

I also think there are trade-offs to consider. We need to optimise for readability so other developers can understand and maintain the system. You're a very smart developer and I'm worried that someone new to the application might struggle to understand how this works.

I'd almost argue it might be worth living with this bug in the short-term and consider alternatives that help mitigate this problem such as increasing the page size or showing a link to the 'Next page' of results instead of the 'Next item', instead to keep complexity down so we can iterate faster.

I'm really not sure. Maybe @pmanrubia can take a look.

(As an aside, I think deferring the 'next item' decision to the next request is definitely a good idea and we should do that regardless of the mechanism for determining the next item)

pmanrubia:
Thank you both for your work on this 👍 

I would close this PR as @cbaines has already extracted #209 and #210; 
I would attach all these findings int he Trello card so we can have as much context as possible to provide a robust solution for user navigation.

Not sure if we have captured this text from @cbaines as a potential bug 
> There is also a fringe case in the back of my mind where users return to the audit page for an item long after originally loading it, in which case the page parameter might not be correct.

cbaines:
> You've clearly put a lot of thought into this :+1:
> 
> I had initially thought handling this in the controller and setting the page parameter to the next page might be the way to go, but I think you've highlighted other problems.
> 
> When we started the MVP, we tried to reduce scope by deferring the handling of concurrent users until later. We thought about ideas such as:
> 
>     Pre-allocating content items to specific users
>     Showing a notification to users if they're on the same audit page (websockets?)
>     Jumping to random items in the list instead of the next item
> 
> We decided this needed more thought and was too much to tackle this quarter.
> 
> I think this bug fix is in danger of snowballing into a potential fix for this concurrency problem. It's good to put thought into this, but I'm worried the scope will grow substantially as we discover more scenarios and edge-cases that aren't handled.
> 
> I also think there are trade-offs to consider. We need to optimise for readability so other developers can understand and maintain the system. You're a very smart developer and I'm worried that someone new to the application might struggle to understand how this works.
> 
> I'd almost argue it might be worth living with this bug in the short-term and consider alternatives that help mitigate this problem such as increasing the page size or showing a link to the 'Next page' of results instead of the 'Next item', instead, to keep complexity down so we can iterate faster.

I think I understand where you are coming from about the complexity issue. I'd still quite like to at least fix the issue of the Next (and go to next on save) function not working at the end of pages. I've written up the other most clear option on the table, which I hope is similar to what you alluded to above, where the scope which is searched is slightly expanded, and the page parameter value incremented when crossing the page boundary https://github.com/alphagov/content-performance-manager/pull/223

> (As an aside, I think deferring the 'next item' decision to the next request is definitely a good idea and we should do that regardless of the mechanism for determining the next item)

At least for today, I've stopped working on this, as while I think it is an improvement, it does clearly stray in to the area of thinking about how multiple users would concurrently use the system.

> (Also, I'm not on this team next week so do whatever you want. :yum: )

Well, I'm in the same situation, and at least for me, even if I wanted to continue working on this, I would guess that my priorities are going to shift with a change in teams.

pmanrubia:
See #223

PR #200, opened 2017-06-23
WIP: Add audit sorting

cbaines:
Trello: https://trello.com/c/gs4d7GDP/267-3-add-additional-sort-criteria-on-content-list-page

### DIFF
--- a/app/domain/search/sort.rb
+++ b/app/domain/search/sort.rb
@@ -4,18 +4,18 @@ class Search
 
     def self.all
       [
-        Sort.new(:page_views_desc, "Six months page views (Decending)",
+        Sort.new(:page_views_desc, "Most views",
                  "six_months_page_views", :desc),
-        Sort.new(:page_views_asc, "Six months page views (Ascending)",
+        Sort.new(:page_views_asc, "Least views",
                  "six_months_page_views", :asc),
-        Sort.new(:title_desc, "Title (Decending)",
-                 "title", :desc),
-        Sort.new(:title_asc, "Title (Ascending)",
+        Sort.new(:title_desc, "A - Z",
                  "title", :asc),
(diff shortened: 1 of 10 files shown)


tuzz:
This is looking good so far!

app/domain/search/sort.rb
```diff
@@ -4,31 +4,56 @@ class Sort
...
+          # value for the next item could be null, or it might not be.
+          scope
+        when :last
+          scope.where("#{field} IS NULL")
+        end
+      else
+        scope.where(
+          "#{field} #{{ desc: '<=', asc: '>=' }[order]} ?", field_value
+        )
+      end
```
tuzz: What does this code do? What's the issue with NULL values?

cbaines:
I'll leave this as WIP until the direction is clearer on what happens with the "next" functionality. This branch currently includes my first implementation of the improvements to that.

app/domain/search/sort.rb
cbaines: The `where_only_after` function takes a scope and content_item, and returns a new scope based on the one provided, but where the scope has been narrowed as much as possible to only include items after the provided item.

From the Sort class, you know the field that is being used for the sort, and the order. From the provided content item, you know the value of the field which is being used in the sort. From these three bits of information, you can generate a new condition to apply to the scope to further limit it to only items which have a sort value equal to or after (this doesn't quite match the naming, it needs weakening).

This approach won't work where the value of the field that the sort is using is nil, as that will be `NULL` in the query. However, the sort includes the information on where the null values are placed. If they are first, then I don't think there is a way of narrowing down the possibilities, so just return the scope. If the null values are last, you can at least eliminate anything that is not null, so do that.

Does that make some amount of sense?

pmanrubia:
I am closing this PR. We will continue with this work once this feature has been prioritised again.

PR #203, opened 2017-06-26
Update audit criteria and add links

chao-xian:
[Trello card](https://trello.com/c/fwzqa4Fj)

Re-works the criteria questions: passing an audit is inverted from all being “Yes” answers to all being “No” answers - as in from being “Does the content have these valid attributes” to “Does the content have to change for each of these criterion”.

For content items that are editable by Whitehall publisher, include a link to edit that item. Ideally we would do this for all content items, whatever their publishing application. This is just a first iteration for a short deadline.

Add hardcoded auditing guidance content.

### DIFF
--- a/app/views/audits/show.html.erb
+++ b/app/views/audits/show.html.erb
@@ -28,6 +28,14 @@
     <%= @audit.content_item.description %>
   </p>
 
+  <%if @audit.content_item.whitehall_url %>
+    <p>
+      <%= link_to "Open in Whitehall Publisher", @audit.content_item.whitehall_url %>
+    </p>
+  <% end %>
+
+  <hr/>
+
   <%= form_for [@audit.content_item, @audit], url: content_item_audit_path(filter_params) do |form| %>
     <%= form.fields_for :responses do |builder| %>
       <% response = builder.object %>
--- a/app/views/audits/show.html.erb
+++ b/app/views/audits/show.html.erb
@@ -36,6 +36,8 @@
 
   <hr/>
 
+  <h4>Do these things need to change?</h4>
+
   <%= form_for [@audit.content_item, @audit], url: content_item_audit_path(filter_params) do |form| %>
     <%= form.fields_for :responses do |builder| %>
       <% response = builder.object %>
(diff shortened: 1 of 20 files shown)


db/migrate/20170626153519_re_seed_questions.rb
```diff
@@ -0,0 +1,20 @@
+class ReSeedQuestions < ActiveRecord::Migration[5.1]
+  class Audit < ActiveRecord::Base
+  end
+
+  class Question < ActiveRecord::Base
+  end
+
+  class Response < ActiveRecord::Base
+  end
```
tuzz: 👍

db/schema.rb
```diff
@@ -154,4 +187,6 @@
     t.datetime "updated_at", null: false
   end
 
+  add_foreign_key "taxonomy_todos", "content_items"
+  add_foreign_key "taxonomy_todos", "taxonomy_projects"
```
tuzz: I think you have some extra tables in your database from other branches. You might need to remove these changes from the branch and amend the commit.

app/views/audits/show.html.erb
```diff
@@ -28,6 +28,16 @@
     <%= @audit.content_item.description %>
   </p>
 
+  <%if @audit.content_item.whitehall_url %>
+    <p>
+      <%= link_to "Open in Whitehall Publisher", @audit.content_item.whitehall_url %>
+    </p>
+  <% end %>
```
tuzz: 👍

config/routes.rb
```diff
@@ -9,13 +9,16 @@
...
+  resources :audits, only: %w(index guidance)
 
   namespace :audits do
     get :report
     get :export
+    get :guidance
   end
 
+  get "audit-guidance", to: "audits#guidance"
+
```
tuzz: Do we need both routes?

spec/features/audit/guidance_spec.rb
```diff
@@ -0,0 +1,8 @@
+RSpec.feature "Auditing a content item", type: :feature do
+  scenario "visiting the guidance page" do
+    visit "/audit-guidance"
+
+    expect(page).to have_css("h1")
+    expect(page).to have_content("Audit GOV.UK content")
+  end
+end
```
tuzz: 👍

PR #206, opened 2017-06-27
Content preview modal

koetsier:
Trello: https://trello.com/c/4IT79oPH/157-concept-generator-being-able-to-see-side-by-side

When tagging content, users have reported it takes time to switch between the content tagger and the actual content. Users open the content in a new tab and switch between tabs to read content and do the tagging.

This pull request changes the link into a modal popup window so that reviewing content can be done on the same page. It is still possible to open up a window in a new tab by right clicking the link.

There is also a static Iframe at the bottom of the page. We need more user research to determine if the modal or the iframe at the bottom of the page is preferred

screen shot 2017-06-28 at 11 15 57
screen shot 2017-06-27 at 15 34 29

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,7 @@ end
 group :development, :test do
   gem 'capybara'
   gem 'database_cleaner'
+  gem 'phantomjs'
   gem 'factory_girl_rails'
   gem 'govuk-lint'
   gem 'guard-rspec', require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -251,6 +251,7 @@ GEM
     parser (2.4.0.0)
       ast (~> 2.2)
     pg (0.20.0)
+    phantomjs (2.1.1.0)
     plek (2.0.0)
     poltergeist (1.15.0)
       capybara (~> 2.1)
@@ -454,6 +455,7 @@ DEPENDENCIES
   listen
   logstasher
   pg
+  phantomjs
   plek
   poltergeist
   pry-rails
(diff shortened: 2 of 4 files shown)


nickcolley:
When you tested this with users did you find this is simpler than a new tab? It seems as though the journey would be the same with a model, as it's still not side by side.

koetsier:
> is with users did you find this is simpler than a new tab? It seems as though the journey would be the same with a modal, as it's still not side by side.
> Merge state

This has not yet been tested with users - I still need to get some information from real users to see if this is considered an improvement.

The reason I chose a modal over side-by-side is to maximise the use of screen space. The content tagging interface itself already takes up quite a bit of space and it may get more features later, such as completion statistics or a list of used tags. I thought adding a new frame on the page with an entire web page in it may make the interface look a bit too busy. The modal makes switching between content and the tagger quite quick.

tijmenb:
Would something like this work?

- Make the main content area 9 or 10 columns
- Put the iframe under the form

desing

(Artists impression)

tijmenb:
From IRL conversation with @koetsier: we're going to try both options at the same time, and see which ones the content designers prefer in day-to-day use.

whoojemaflip:
Could the iFramed content load in the background, to be shown when the user clicks the link (or taps a hotkey). this way it would load *immediately*, saving seconds.

koetsier:
The initial screen now looks like:

screen shot 2017-06-28 at 11 15 57

Both options are now available: modal and the page at the bottom. Need to hear from users which one is preferred.

tijmenb:
👍 Let's try this!

PR #209, opened 2017-06-27
Find the next audit upon the next action

cbaines:


app/controllers/audits_controller.rb
```diff
@@ -10,6 +10,15 @@ def show
     next_item
   end
 
+  def next
+    if next_item
+      redirect_to content_item_audit_path(next_item, filter_params)
+    else
+      flash.alert = "Unable to find the next item to audit."
```
pmanrubia: I would be more positive about this message. `There is not more items to audit`? Can you please check with Joe?

spec/features/audit/navigation_spec.rb
```diff
@@ -71,23 +69,4 @@
     expected = content_item_audit_path(second, some_filter: "value")
     expect(current_url).to end_with(expected)
   end
-
```
pmanrubia: Could you please add a test for this feature? As we are not able to precalculate the link, shouldn't we test that we see the expected message once we try to navigate to the latest item?

cbaines:
Thanks for your review Pablo.

I'm no longer certain that this is worth perusing as there is general uncertainty about how to handle potential improvements related to this "out of scope" issue of multiple concurrent users.

Joe has raised the issue that this provides an action that might not work, which is bad for the user experience. So on the whole, I think it might be better to consolidate my time into other tickets that are currently not finished off.

PR #210, opened 2017-06-27
Get the next_item before updating the audit [SEAL IGNORE]

cbaines:
This fixes an issue where finding the next item doesn't work when
current audit is saved, and the current filtering conditions are
looking for non-audited items.

### DIFF
--- a/app/controllers/audits_controller.rb
+++ b/app/controllers/audits_controller.rb
@@ -15,6 +15,13 @@ class AuditsController < ApplicationController
 
   def save
     audit.user = current_user
+    # Get the next_item before updating the audit. As the
+    # next_content_item functionality depends on finding the current
+    # item in the search, if updating the audit removes the current
+    # item from the search, this can mean that the next item can't be
+    # found.
+    next_item
+
     updated = audit.update(audit_params)
 
     if next_item && updated
--- a/spec/features/audit/audit_spec.rb
+++ b/spec/features/audit/audit_spec.rb
@@ -6,6 +6,7 @@ RSpec.feature "Auditing a content item", type: :feature do
       description: "All about flooding.",
       base_path: "/flooding",
       publishing_app: "whitehall",
+      six_months_page_views: 10,
     )
   end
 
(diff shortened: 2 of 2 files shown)


app/controllers/audits_controller.rb
```diff
@@ -12,6 +12,13 @@ def show
 
   def save
     audit.user = current_user
+    # Get the next_item before updating the audit. As the
```
pmanrubia: Can you please add a test to drive this fix?

Can you think of a way to get rid of the comment via:
1. Writing more expressive code
2. Describing what you try to fix in the test?

Regardless the fix, it feels to me that it assumes that we rely on having the next item calculation precalculated, but as you described in https://github.com/alphagov/content-performance-manager/pull/198#issuecomment-311403441, it has many drawbacks.

So, it might be better to find a good solution for navigation rather than adding quick fixes?

cbaines:
I've now included this fix in https://github.com/alphagov/content-performance-manager/pull/223 as it didn't seem sensible to keep the issue present there when reworking the related code.

app/controllers/audits_controller.rb
cbaines: > Can you please add a test to drive this fix?

Yep, will see if I can do that today.

> Can you think of a way to get rid of the comment via:
> 
>     Writing more expressive code

I had a think about this yesterday, mostly while writing https://github.com/alphagov/content-performance-manager/pull/223 , and I haven't come up with anything yet.

>     Describing what you try to fix in the test?
> 
> Regardless the fix, it feels to me that it assumes that we rely on having the next item calculation precalculated, but as you described in #198 (comment), it has many drawbacks.
> 
> So, it might be better to find a good solution for navigation rather than adding quick fixes?

In this case, which is the case where the Save button is pressed, the behaviour is simpler than the Next link, as the decision about the next item is deferred until the point at which the user clicks save (rather than the Next link, where its currently decided when the page is rendered). My current thinking on this is that its independent of the navigation, as its just about making sure that the next item can be found when saving an audit. Does this make sense?

cbaines: > Can you please add a test to drive this fix?

I've now added a test that covers moving to the next item to audit when save is clicked.

chao-xian:
@cbaines does this PR need to be closed?

cbaines:
> @cbaines does this PR need to be closed?

Not as far as I'm aware? But I haven't looked at it since moving teams.

chao-xian:
After discussion with @pmanrubia I'm going to close this PR as the fix is in #223, but altered which makes merging this PR harder to accommodate if we decide to merge in #223.

PR #213, opened 2017-06-28
Add a script to compare theme base_paths to a CSV

tuzz:


lib/comparison.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def base_paths_in_csv
+    paths = csv.map(&:first)[1..-1]
+    normalise(paths)
+  end
+
+  def normalise(base_paths)
+    base_paths
+      .reject(&:nil?)
```
cbaines: You could say compact here?

lib/comparison.rb
```diff
@@ -0,0 +1,60 @@
...
+    additions
+  end
+
+private
+
+  attr_accessor :additions, :deletions
+
+  def compute_additions_and_deletions
+    self.additions = base_paths_in_theme - base_paths_in_csv
+    self.deletions = base_paths_in_csv - base_paths_in_theme
```
cbaines: Storing the values here, rather than invoking the methods twice would be neater and more efficient.

lib/comparison.rb
```diff
@@ -0,0 +1,60 @@
...
+  def base_paths_in_csv
+    paths = csv.map(&:first)[1..-1]
+    normalise(paths)
+  end
+
+  def normalise(base_paths)
+    base_paths
+      .reject(&:nil?)
+      .map { |p| p.sub(/^.*gov\.uk/, "") }
+      .map { |p| p.sub(/^\/government/, "") }
```
cbaines: I get the previous normalisation, but I'm unsure why removing the prefix `/government` is something that wants doing?

tuzz:
@cbaines: Thanks for the suggestions. I have addressed those issues.

Going to merge now.

PR #214, opened 2017-06-28
Allow users to create taxonomy projects

tijmenb:
This adds interface for users to create new projects. This should make it easier to test by content designers, because there's no developer involvement.

Because the todos are created in the request, this might not work for large spreadsheets (the page might time out). In that case we can still use the rake task.

https://trello.com/c/Ng5Ggpsw

### DIFF
--- a/app/views/taxonomy_projects/new.html.erb
+++ b/app/views/taxonomy_projects/new.html.erb
@@ -1,0 +1,53 @@
+<h1>Create new project</h1>
+
+<div class='row'>
+  <div class='col-md-8'>
+    <%= form_for @form, url: taxonomy_projects_path do |f| %>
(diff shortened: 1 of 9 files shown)


app/views/taxonomy_projects/new.html.erb
```diff
@@ -0,0 +1,53 @@
...
+
+        <p>The spreadsheet should have a <code>content_id</code> column with the
+          content IDs of the pages. </p>
+      </div>
+
+      <%= f.submit "Save", class: "btn btn-success" %>
+    <% end %>
+  </div>
+
+  <div class='col-md-4'>
```
tijmenb: These instructions come from content-tagger and have been tested with a content designer.

PR #216, opened 2017-06-28
Hide the audit status filter on the report page

cbaines:
Also, don't filter by audit status at all on the report page.

### DIFF
--- a/app/controllers/audits_controller.rb
+++ b/app/controllers/audits_controller.rb
@@ -1,5 +1,6 @@
 class AuditsController < ApplicationController
-  helper_method :filter_params, :primary_org_only?, :org_link_type
+  helper_method :filter_params, :primary_org_only?, :org_link_type,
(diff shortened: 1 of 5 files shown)


tuzz:
We could share the conditional between the controller and the view with:

```
# audits_controller.rb
helper_method :audit_status_filter_enabled?

def audit_status_filter_enabled?
  action_name != "report"
end
```

Then use that in both places.

tuzz:
Tests are failing.

cbaines:
Good idea about the controller method, I'll update.

cbaines:
> Tests are failing and we really ought to test the audit status filter doesn't appear and isn't in effect on the report page (given this constituted a P1 – i.e. it's crucial this works as expected).

Finally I have written some passing tests. I got a bit muddled up with the test data, as I couldn't seem to get the organisation filter to contain anything, turns out I had the link type wrong.

PR #217, opened 2017-06-28
Provide a checkbox to toggle primary/all orgs. filter

tuzz:
https://trello.com/c/MzAVD2dG/320-2-update-org-filter-to-toggle-primary-filter-on-off

The hidden field means the URL param is set to
false when not explicitly overriden by the
checkbox. We need to do this because we want to
checkbox to default to checked.

The count of content for each organisation will
depend on whether the checkbox is selected during
page load. If the user changes the checkbox, those
numbers won’t be updated immediately.

We’d need to progressively enhance the form with
JS to solve this problem.

### DIFF
--- a/app/controllers/audits_controller.rb
+++ b/app/controllers/audits_controller.rb
@@ -1,5 +1,5 @@
 class AuditsController < ApplicationController
-  helper_method :filter_params
+  helper_method :filter_params, :primary_org_only?, :org_link_type
   before_action :content_items, only: %i(index report export)
 
   def index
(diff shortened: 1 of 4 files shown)


cbaines:
Joe and I noticed that clicking on the label didn't toggle the state on the checkbox. This looks to be because the hidden field has the same id as the non-hidden input. We couldn't work out a way of not specifying the id through the helper, but swapping it out for the following fixes clicking on the label, while keeping everything working.

```
<input type="hidden" name="primary" value="false" />
```

tuzz:
Good catch. I'll swap the helper out for that snippet. Thanks.

PR #223, opened 2017-06-29
Fix next link for audit show page (take 2) [SEAL IGNORE]

cbaines:
See comment in #198 

Trello: https://trello.com/c/B8YQWsfv/308-possible-bug-if-you-click-the-last-item-on-a-list-page-you-dont-get-a-next-link

### DIFF
--- a/app/controllers/audits_controller.rb
+++ b/app/controllers/audits_controller.rb
@@ -10,16 +10,22 @@ class AuditsController < ApplicationController
 
   def show
     audit.questions = audit.template.questions if audit.new_record?
-    next_item
+    compute_next_item_and_page
   end
 
   def save
     audit.user = current_user
+    # Compute the next_item before updating the audit. As the
+    # next_content_item functionality depends on finding the current
+    # item in the search, if updating the audit removes the current
+    # item from the search, this can mean that the next item can't be
+    # found.
+    compute_next_item_and_page
     updated = audit.update(audit_params)
 
-    if next_item && updated
+    if @next_item && updated
       flash.notice = "Saved successfully and continued to next item."
-      redirect_to content_item_audit_path(next_item, filter_params)
+      redirect_to content_item_audit_path(@next_item, @next_link_filter_params)
     elsif updated
       flash.now.notice = "Saved successfully."
       render :show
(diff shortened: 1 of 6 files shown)


tuzz:
This is all good work but it really needs tests. This is an important feature / bug fix and without tests we increase our risk this will break again in the future.

cbaines:
> This is all good work but it really needs tests. This is an important feature / bug fix and without tests we increase our risk this will break again in the future.

Ok, I'll take a look at improving the testsuite.

app/controllers/audits_controller.rb
```diff
@@ -59,8 +65,34 @@ def content_items
     @content_items ||= search.content_items.decorate
   end
 
-  def next_item
-    @next_item ||= content_items.next_item(content_item)
+  def compute_next_item_and_page
+    # Get a list of ContentItem ids for everything on the current
```
pmanrubia: Would you say this belongs in the controller?

pmanrubia: What are your thoughts about extracting this calculation into a class? It seems a bit complex to live in the controller.

Something along the lines of:

```ruby
iterator = @search.iterator()
content_item = iterator.next(starting: index)
```

or

```ruby
iterator = Audit::Iterator.new(@search)
content_item = iterator.next()
```

And include all the logic + tests in the `Audit::Iterator` class as a first step to isolate this behaviour?

cbaines: This method has two side effects, the `@next_item`, but also the `@next_link_filter_params`, which is like the `filter_params`, but can include a different page.

I did think of having some function that would return the two bits of information, e.g. `next_item, next_page = next_item_and_page(scope, current_content_item)`, but decided against in as just having a method that set instance variables seemed similar to the style of the AuditsController.

I'm not against having it in a class, I put it in the controller for a start to keep the complexity low.

pmanrubia:
I am closing this PR. 
All the valuable findings have been recorded and tracked via this [Trello card](https://trello.com/c/B8YQWsfv/308-if-you-click-the-last-item-on-a-list-page-you-dont-get-a-next-link). We plan to re-implement navigation from scratch to address the issues that we are currently having.

PR #230, opened 2017-07-11
Proxy iframes

koetsier:
Adds a proxy for all www.gov.uk content, so that these pages can be displayed in an iFrame in the concept generator.

Links in the contents of the iFrame are rewritten so that they point to the proxy. 

See also: https://github.com/alphagov/content-performance-manager/pull/206

Trello: https://trello.com/c/4IT79oPH

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -38,4 +38,12 @@ Rails.application.routes.draw do
   if Rails.env.development?
     mount GovukAdminTemplate::Engine, at: "/style-guide"
   end
+
+  class ProxyAccessContraint
+    def matches?(request)
+      !request.env['warden'].try(:user).nil?
+    end
+  end
+
+  mount Proxies::IframeAllowingProxy.new => Proxies::IframeAllowingProxy::PROXY_BASE_PATH, constraints: ProxyAccessContraint.new
 end
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,7 @@ gem 'selectize-rails'
 gem 'turbolinks'
 gem 'uglifier'
 gem 'unicorn'
+gem 'rack-proxy'
 
 group :development do
   gem 'web-console'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -289,6 +289,8 @@ GEM
       rack (>= 0.4)
     rack-protection (2.0.0)
       rack
+    rack-proxy (0.6.1)
+      rack
     rack-test (0.6.3)
       rack (>= 1.0)
     rails (5.1.2)
@@ -483,6 +485,7 @@ DEPENDENCIES
   poltergeist
   pry-rails
   puma
+  rack-proxy
   rails (~> 5.1)
   rails-controller-testing
   rspec-rails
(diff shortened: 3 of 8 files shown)


config/routes.rb
```diff
@@ -35,4 +35,6 @@
   if Rails.env.development?
     mount GovukAdminTemplate::Engine, at: "/style-guide"
   end
+
+  mount Proxies::GovernmentProxy.new => Proxies::GovernmentProxy::PROXY_BASE_PATH
```
tijmenb: Can we lock this down to authenticated users? You can create a constraint that verifies `request.env['warden'].user` isn't nil. See collections-publisher for an example:

https://github.com/alphagov/collections-publisher/blob/master/config/routes.rb#L39-L49

app/models/proxies/government_proxy.rb
```diff
@@ -0,0 +1,34 @@
+require 'json'
+
+module Proxies
+  class GovernmentProxy < Rack::Proxy
```
tijmenb: Maybe something like `IframeAllowingProxy` would reveal the intent? We could also add a few comments on why we're doing this.

app/models/proxies/government_proxy.rb
```diff
@@ -0,0 +1,34 @@
...
+    PROXY_BASE_PATH = '/proxy/govuk/'.freeze
```
tijmenb: Same thing, maybe `/iframe-proxy/`? (the `/govuk` may be redundant)

app/models/proxies/government_proxy.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    def rewrite_env(env)
+      env['HTTP_HOST'] = 'www.gov.uk:443'
+      env['REQUEST_URI'] = env['HTTP_HOST']
+      env['SERVER_NAME'] = 'www.gov.uk'
+      env['SERVER_PORT'] = 443
+      env['SCRIPT_NAME'] = nil
+      env['HTTP_X_FORWARDED_PORT'] = nil
+      env['HTTP_X_FORWARDED_PROTO'] = nil
+      env['rack.ssl_verify_none'] = true
```
tijmenb: Doesn't SSL work?

spec/models/proxies/government_proxy_spec.rb
```diff
@@ -0,0 +1,41 @@
+RSpec.describe Proxies::GovernmentProxy do
+  before :each do
+    @proxy = Proxies::GovernmentProxy.new
+    @status = {}
+    @headers = {}
+    Proxies::GovernmentProxy::PROXY_BASE_PATH = '/proxy/gov/'.freeze
+  end
+
+  describe '#rewrite_response' do
```
tijmenb: Can you add some line breaks between the blocks to make it consistent with other specs? (Sorry about the style comment, this should be picked up by Rubocop).

app/models/proxies/government_proxy.rb
```diff
@@ -0,0 +1,34 @@
+require 'json'
+
+module Proxies
+  class GovernmentProxy < Rack::Proxy
+    PROXY_BASE_PATH = '/proxy/govuk/'.freeze
+
+    def rewrite_env(env)
+      env['HTTP_HOST'] = 'www.gov.uk:443'
+      env['REQUEST_URI'] = env['HTTP_HOST']
```
tijmenb: Is this correct? Feels like it should be the full URL?

PR #231, opened 2017-07-18
Use FileUpdateChecker as file watcher

pmanrubia:
This proves a bit more reliable when updating the codebase outside of
the VM.


config/environments/development.rb
```diff
@@ -50,7 +50,7 @@
 
   # Use an evented file watcher to asynchronously detect changes in source code,
```
alecgibson: This comment is out-of-date.

pmanrubia:
Thank you @alecgibson I have removed the comment.

PR #232, opened 2017-07-21
Switch from id to content-id in URLs

chao-xian:
We refer to content items by their content_item_id in most GOV.UK apps so we should be consistent. Here we’ve switched the param in the URL path to be for the content-id instead of id when referring to content items.

### DIFF
--- a/app/controllers/audits_controller.rb
+++ b/app/controllers/audits_controller.rb
@@ -51,7 +51,7 @@ private
   end
 
   def content_item
-    @content_item ||= ContentItem.find(params.fetch(:content_item_id)).decorate
+    @content_item ||= ContentItem.find_by(content_id: params.fetch(:content_item_content_id)).decorate
   end
 
   def content_items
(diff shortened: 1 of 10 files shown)


pmanrubia:
Did you consider using [to_param](https://apidock.com/rails/ActiveRecord/Base/to_param) in `ContentItem`?. Would it simplify the PR?

app/controllers/audits_controller.rb
```diff
@@ -51,7 +51,7 @@ def audits
   end
 
   def content_item
-    @content_item ||= ContentItem.find(params.fetch(:content_item_id)).decorate
+    @content_item ||= ContentItem.find_by_param(params.fetch(:content_item_content_id)).decorate
```
alecgibson: Aren't these dynamic methods [deprecated](http://guides.rubyonrails.org/4_0_release_notes.html#active-record-deprecations)?

PR #233, opened 2017-07-21
Use Links/Search to query content in CPM

pmanrubia:
[Trello card](https://trello.com/c/Xo1fWIKb/355-refactor-search-metrics-content-items-in-cpm)

Also fixes [a duplication bug](https://trello.com/c/uZ4UBiCQ/245-meningococcal-the-green-book-chapter-22-appears-3-times)

### Background

The Audit Tool lives in the CPM. When the Audit tool was developed, we made the decision to mimic how Publishing API stores ContentItems and Links. The reasons behind these decisions were:

1. It simplifies the way we query for different entities (Organisations, Policies...) 
2. It simplifies the way we synchronize data with Publishing API

We are tackling this work now because when we developed the Audit Tool we didn't refactor how the CPM search for content.

### About this PR

1. Move the codebase to a better position so CPM and Audit Tool share a common approach to query content.
2. There are some areas that are duplicated between CPM and Audit tool, but I am not addressing those issues here to keep the scope of the PR as small as possible.
3. For the same reason, I am not doing [other refactorings mentioned in this Trello board](https://trello.com/b/laICtTQ8/content-tools-development).
4. I am not deleting the database tables in case we need to rollback this version.


app/models/content_item.rb
```diff
@@ -72,12 +70,4 @@ def whitehall_url
     return unless publishing_app == "whitehall"
     "#{WHITEHALL}/government/admin/by-content-id/#{content_id}"
   end
-
-  def add_organisations_by_id(orgs)
```
chao-xian: This change doesn't seem to fit with this commit 354fad88691ba1a6abb48df4c9150fe33cd3ee96. I think the the deletion in the last commit of adding taxons needs to be swapped with the one here for adding organisations.

app/views/content_items/_header.html.erb
```diff
@@ -0,0 +1 @@
+<h1> <%= stringify_filter %> </h1>
```
chao-xian: Have added https://trello.com/c/t8c9GdS8

pmanrubia:
Rebased and force pushed in preparation for merging.

PR #236, opened 2017-07-21
Fetch next item through SQL query

alecgibson:
This Pull Request makes two main changes (separated into their own commits):

0. Separate the concerns of a `Query` out from a `Search`
0. Change the mechanism behind fetching the "next" content item to use a SQL query

## Separate `Search` and `Query`

Previously, the concepts of a "search" and a "query" were quite confused. You would instantiate a `Search` object, then mutate the query indirectly through mutating the search. You would then manually execute the search, and then read the results. It was also possible to perform similar searches by having your search object clone its own query and mutate it, and produce other results, but in a way that was unclear from the outside that you were performing a second search.

This change separates the query into its own object, which must be configured before being passed into a search. A search now represents a query and its executed results, neither of which
can be changed after running the search.

Importantly, by separating out queries, we also make it much clearer when we are cloning them and creating subqueries (instead of doing it indirectly via the `Search` class) - which is used by the second part of this pull request.

## Fetch next content items through SQL

Previously our "next" content item was fetched by looking up the index of our current content item in a list of results, and then asking for the next index. This is problematic, because if the list of results is paged, then we cannot fetch the next item across the page break. Also, the results can grow stale as other users work on the audit.

In this new approach, we instead rely on the fact that our items are ordered. Relying on this means that we can perform a greater-than query to fetch the next item, according to the ordering. This query will always return the correct result (with the exception of race conditions with multiple users); and is completely independent of paging.

## Trello

https://trello.com/c/B8YQWsfv/308-if-you-click-the-last-item-on-a-list-page-you-dont-get-a-next-link


alecgibson:
[If you click the last item on a list page, you don't get a 'next' link](https://trello.com/c/B8YQWsfv/308-if-you-click-the-last-item-on-a-list-page-you-dont-get-a-next-link)

alecgibson:
This is now replaced by: https://github.com/alphagov/content-performance-manager/pull/252

PR #240, opened 2017-07-25
Add a LinkFactory for tests

alecgibson:
When creating links between Content Items, we previously had to make
the two content items and then also create a Link object between them.

This change allows for links to be implicitly created by passing
the linked content item(s) into the factory method of the parent
content item.


alecgibson:
[Add Factories for organisations, taxons...](https://trello.com/c/QRA7ntOb/369-add-factories-for-organisations-taxons)

PR #241, opened 2017-07-25
Add tag in concept generator when losing focus

koetsier:
Trello: https://trello.com/c/1aXV1MWc/33-concept-gen-too-easy-to-think-youve-added-a-term-when-you-havent

This option adds a tag when shifting focus from the input field. When entering a tag and saving without explicitly adding, will now include the tag anyway.

The terms field can no longer be empty - it will raise an error (screenshot). The buttons 'I don't know', and 'The page is not relevant...' will still allow a user to skip a page.

screen shot 2017-07-25 at 16 42 27

### DIFF
--- a/app/controllers/taxonomy_todos_controller.rb
+++ b/app/controllers/taxonomy_todos_controller.rb
@@ -1,18 +1,26 @@
 class TaxonomyTodosController < ApplicationController
   def show
-    @todo_form = TaxonomyTodoForm.new(
-      taxonomy_todo: taxonomy_todo,
-      terms: taxonomy_todo.terms.order(:name).pluck(:name).join(','),
-    )
+    @todo_form = TaxonomyTodoForm.new(taxonomy_todo: taxonomy_todo)
+    if flash.key?(:taxonomy_todos_params)
+      @todo_form.terms = flash[:taxonomy_todos_params][:terms]
+      @todo_form.valid?
+    else
+      @todo_form.terms = taxonomy_todo.terms.order(:name).pluck(:name).join(',')
+    end
   end
 
   def update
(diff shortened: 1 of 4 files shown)


app/controllers/taxonomy_todos_controller.rb
```diff
@@ -7,12 +7,16 @@ def show
   end
 
   def update
-    todo_form = TaxonomyTodoForm.new(todo_params)
-    todo_form.taxonomy_todo = taxonomy_todo
-    todo_form.user = current_user
-    todo_form.save
-
-    redirect_to_next_item
+    @todo_form = TaxonomyTodoForm.new(todo_params)
```
pmanrubia: I would expect to see some tests in a PR so we can guarantee that we are not breaking anything.

koetsier: 👍

app/controllers/taxonomy_todos_controller.rb
```diff
@@ -7,12 +7,16 @@ def show
...
+    @todo_form.taxonomy_todo = taxonomy_todo
+    @todo_form.user = current_user
+    if @todo_form.valid?
+      @todo_form.save
+      redirect_to_next_item
+    else
+      flash.now.alert = @todo_form.errors.full_messages.join(', ')
+      render :show
```
whoojemaflip: Have you considered a redirect-on-post approach here?

koetsier: 👍

whoojemaflip:
needs rebasing

pmanrubia:
Do you need to have 4 commits for this work? The last 2 commits are addressing comments. Would it make sense to fix them up?

PR #243, opened 2017-07-26
Stabilise sort order

alecgibson:
Our sort was previously unstable. That is, if we sorted on - for
example - Page Views, if two results had equal page views, they may
appear in any random order.

This change breaks all ties on the content items’ base paths, which are
unique, and hence guarantees that a given sort order will always return
items in the same order.


alecgibson:
[`Meningococcal: the green book, chapter 22` appears 3 times](https://trello.com/c/uZ4UBiCQ/245-meningococcal-the-green-book-chapter-22-appears-3-times)

app/domain/search/executor.rb
```diff
@@ -58,7 +58,12 @@ def apply_filters(filters, scope)
     end
 
     def apply_sort!
-      self.scope = query.sort.apply(scope)
+      sort_query = {}
+      sort_query[query.sort] = query.sort_direction
+      # Finally sort by base path (which is unique) to stabilise sort order
+      sort_query[:base_path] = :asc
+
+      self.scope = scope.order(sort_query)
```
pmanrubia: what about:

```ruby
self.scope = scope.order (
  base_path => :asc, # stabilises sort order
  query.sort => query.sort_direction,
)

PR #244, opened 2017-07-26
Add performance module

pmanrubia:
[Trello card](https://trello.com/c/yvOQatkB/383-add-performance-module)

Add a `performance` module to wrap all classes related to performance.
We are trying to make the code more cohesive in each module/folder so
it better reveals what it does.


pmanrubia:
Merged as it was codereviewed in #245

spec/models/importers/single_content_item_spec.rb
```diff
@@ -1,105 +1,105 @@
...
-    end
+      it "updates the content item" do
+        content_item.save!
 
-    it "doesn't create any additional links" do
-      expect { subject.run(content_id) }
-        .not_to change(Link, :count)
+        expect { subject.run(content_id) }
+          .to change { content_item.reload.title }.from("title").to("new title")
+                .and change { content_item.reload.linked_taxons.to_a }.from([]).to([taxon1, taxon2])
```
chao-xian: I think I find this slightly harder to read. Is it a convention to 
```
.chain
  .it
    .like
      .this
```
as opposed to

```
.chain
.it
.like
.this
```
?

It's already clear to me that as each line starts with a `.` that it's a continuation of method chaining without the additional indentation.

pmanrubia:
@chao-xian I have fixed the indentation issue. Thank you for raising this issue.

PR #246, opened 2017-07-27
Add Google Tag Manager script

alecgibson:
This change adds a script to enable tracking with Google Tag Manager.

Notably, we have not included the `noscript` `iframe` part of the
setup, as this is for tracking users with JavaScript disabled, and
requires further configuration. We don’t need this functionality.


alecgibson:
[Set up GTM for the Content Performance Manager](https://trello.com/c/q6vQfoAp/376-set-up-gtm-for-the-content-performance-manager)

PR #249, opened 2017-07-31
Remove filter counts

alecgibson:
The drop-down filters and the inventory manager had counts of content
next to them. For example, in the “Organisations” filter, you might
see “Department for Education (12)”, which would mean that DfE has
12 content items that match your other (non-organisational) filters.

While this is a nice-to-have, there is no confirmed user need for it,
and it complicates the code, because this behaviour is tied heavily
to the Search code (which we hope to refactor).


alecgibson:
[Refactor Search](https://trello.com/c/eQ4lKUE3/387-refactor-search)

alecgibson:
[Remove Filter Counts](https://trello.com/c/3Zg5Uwqs/388-remove-filter-counts)

app/helpers/dropdown_helper.rb
```diff
@@ -20,30 +20,38 @@ def audit_status_options
   end
 
   def taxons_options
-    options = @search.options_for("taxons").order(:title)
+    options = ContentItem
```
pmanrubia: What about inlining this code to:

```ruby
class ContentItem
  def all_taxons
    where(document_type: "taxon").order(:title)
  end
end

taxons = ContentItem.all_taxons
...
```

We could think of a better way to retrieve all taxons in the future, but so far we have that behaviour isolated.

app/helpers/dropdown_helper.rb
```diff
@@ -20,30 +20,38 @@ def audit_status_options
...
       :content_id,
-      :title_with_count,
+      :title,
       selected: params["taxons"],
     )
   end
 
   def organisation_options
-    options = @search.options_for(org_link_type).order(:title)
+    options = ContentItem
```
pmanrubia: Same as comment above for organisations

app/helpers/dropdown_helper.rb
```diff
@@ -20,30 +20,38 @@ def audit_status_options
...
 
   def document_type_options
-    options = @search.options_for(:document_type)
-    options = options.map { |o| DocumentTypeOption.new(o) }
+    options = ContentItem
+      .all
+      .select(:document_type)
+      .group(:document_type)
+      .sort_by(&:document_type)
+      .map { |option| DocumentTypeOption.new(option) }
```
pmanrubia: I might be missing something, but you are returning here a non-unique list of document types. It would be quicker to do something like:

```ruby
ContentItem.pluck(:document_type)
```

And, I would say that you need here the unique list, so what about doing this:

```ruby
ContentItem.pluck(:document_type).uniq
```

spec/features/audit/filter_spec.rb
```diff
@@ -1,7 +1,7 @@
 RSpec.feature "Filter Content Items to Audit", type: :feature do
   # Organisations:
-  let!(:hmrc) { create(:content_item, title: "HMRC") }
-  let!(:dfe) { create(:content_item, title: "DFE") }
+  let!(:hmrc) { create(:content_item, title: "HMRC", document_type: "organisation") }
```
pmanrubia: What about creating a factory here?

```
  let!(:hmr) { create :organisation, title: "HMRC" }
```

app/helpers/dropdown_helper.rb
alecgibson: Doesn't `group` mean that it's unique?

alecgibson:
@pmanrubia I've addressed your comments.

app/helpers/dropdown_helper.rb
pmanrubia: Yes, sorry. You are right.

pmanrubia:
👍

PR #250, opened 2017-07-31
Move audit to module

chao-xian:
This PR moves classes relating to auditing into an Audits module. This is a first pass as the naming of the `Audits::Audit` class is not ideal, but we are treating this as an initial step to help facilitate further refactoring.


app/controllers/audits/audits_controller.rb
```diff
@@ -0,0 +1,124 @@
...
+    end
+
+    def guidance
+      @body = Govspeak::Document.new(File.read("doc/guidance.md")).to_html.html_safe
+    end
+
+  private
+
+    def audit
+      @audit ||= Audits::Audit.find_or_initialize_by(content_item: content_item).decorate
```
pmanrubia: No need of prefix here, as you are in the `Audits` module

app/domain/report.rb
```diff
@@ -10,7 +10,7 @@ def self.generate(*args)
   def initialize(content_items, request)
     self.content_items = content_items
     self.request = request
-    self.questions = Question.order(:id).to_a
+    self.questions = Audits::Question.order(:id).to_a
```
pmanrubia: Could you please move the `Report` to the `audits` module within `domain`?. After doing it, you won't need this prefix.

app/models/audits/audit.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    validates :content_item, presence: true
+    validates :user, presence: true
+
+    scope :passing, -> { where.not(id: failing) }
+    scope :failing, -> { where(id: Response.failing.select(:audit_id)) }
+
+    accepts_nested_attributes_for :responses
+
+    after_save { Audits::ReportRow.precompute(content_item) }
```
pmanrubia: No need of prefix here, as you are in the `Audits` module

app/models/audits/inventory_rule.rb
```diff
@@ -0,0 +1,10 @@
+module Audits
+  class InventoryRule < ApplicationRecord
```
pmanrubia: I wouldn't say that the inventory rule belongs to the Audits module.

app/models/audits/report_row.rb
```diff
@@ -0,0 +1,99 @@
...
+    def responses
+      @responses ||= audit ? audit.responses : []
+    end
+
+    def audit
+      @audit ||= content_item.audit
+    end
+
+    def questions
+      @questions ||= Audits::Question.order(:id)
```
pmanrubia: No need of prefix here, as you are in the `Audits` module

app/models/audits/response.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    belongs_to :audit
+    belongs_to :question
+
+    validates :audit, presence: true
+    validates :question, presence: true
+
+    scope :boolean, -> { joins(:question).where(questions: { type: "Audits::BooleanQuestion" }) }
+    scope :passing, -> { boolean.where(value: Audits::BooleanQuestion::PASS) }
+    scope :failing, -> { boolean.where.not(id: passing) }
```
pmanrubia: No need of prefix here, as you are in the Audits module

app/models/audits/template.rb
```diff
@@ -0,0 +1,12 @@
+module Audits
+  class Template
+    def questions
+      fail "You need to run rake db:seed to create the questions" if Audits::Question.none?
+      Audits::Question.all.order(:id).to_a
```
pmanrubia: No need of prefix here, as you are in the `Audits` module

spec/models/audits/audit_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.describe Audit do
+RSpec.describe Audits::Audit do
```
pmanrubia: If you enclose the test within `module Audits` there is no need to scope every class inside

spec/models/audits/question_spec.rb
```diff
@@ -1,6 +1,6 @@
-RSpec.describe Question do
+RSpec.describe Audits::Question do
```
pmanrubia: If you enclose the test within `module Audits` there is no need to scope every class inside

spec/models/audits/response_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.describe Response do
+RSpec.describe Audits::Response do
```
pmanrubia: If you enclose the test within `module Audits` there is no need to scope every class inside. Same comment applies to remaining RSpec tests.

PR #251, opened 2017-08-02
Refactor search

alecgibson:
This change completely removes the `Search` class. This class perviously had knowledge of several areas of the domain (content, audits, themes). The main aim of the change is to replace this with smaller classes whose knowledge is more isolated.

The change is made in a number of atomic commits, which:

* Introduce the replacements for `Search`: `Content::Query` and `Audits::ContentQuery`
* Replace all instances of `Search` with the new classes
* Remove the `Search` class entirely, and move the unit tests to cover the new classes
* Remove the `Results` class, whose usefulness has all but disappeared with the removal of `Search`

PR #253, opened 2017-08-02
Fetch content after another content item

alecgibson:
This change adds the ability to fetch the content item after a given
content item, according to the sort criteria provided in a query.

This is particularly useful when trying to pick the “next” item when
viewing a single content item.

The previous approach to this used the list of content items on a page,
but that approach breaks when viewing the last item on a page.

## Dependencies

- [x] https://github.com/alphagov/content-performance-manager/pull/251

### DIFF
--- a/spec/domain/content/query_spec.rb
+++ b/spec/domain/content/query_spec.rb
@@ -90,6 +90,7 @@ RSpec.describe Content::Query do
     let!(:content_items) { create_list(:content_item, 26) }
 
     it "defaults to the page 1 with 25 per page" do
+      subject.sort_direction(:asc)
       expect(subject.content_items.count).to eq 25
       expect(subject.content_items).to match_array(content_items[0..24])
     end
--- a/spec/domain/content/query_spec.rb
+++ b/spec/domain/content/query_spec.rb
@@ -144,8 +144,51 @@ RSpec.describe Content::Query do
     let!(:content_item_3) { create(:content_item, six_months_page_views: 1) }
     let!(:content_item_4) { create(:content_item, six_months_page_views: 9) }
 
-    it "defaults to six month page views descending" do
-      expect(subject.content_items.pluck(:six_months_page_views)).to match_array [9, 3, 3, 1]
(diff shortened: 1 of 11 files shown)


alecgibson:
[If you click the last item on a list page, you don't get a 'next' link](https://trello.com/c/B8YQWsfv/308-if-you-click-the-last-item-on-a-list-page-you-dont-get-a-next-link)

spec/domain/content/query_spec.rb
```diff
@@ -143,8 +144,31 @@
...
+        expect(subject.content_items).to contain_exactly(content_item_3)
+      end
+    end
+
+    describe "sorting by page views ascending" do
+      before do
+        subject.sort_direction(:asc)
+      end
+
+      it "can pick items after another item" do
```
pmanrubia: Could you please isolate in a `describe` the tests for `after`.
It would be good to test what happens when the element is the last position, last position of a page, first position, ..., in the middle. We have had so many issues with this feature, that it would be good to describe all the edge cases.

pmanrubia: Could you please add an integration test where you replicate the pagination issue?

alecgibson: > It would be good to test what happens when the element is [in the] ... last position of a page

What do you mean? This `after` function has literally zero idea about paging.

pmanrubia: Usually, you look for the next item within a list. And it is in this context where we have had all the issues. For example, would it return nil if there is no `content_item` after the current one?

You can have integration tests that exercise this query in different contexts, and simplify unit tests?

alecgibson:
@pmanrubia I've added some more tests.

alecgibson:
[When filtering by non-audited items, saving an audit won't take you to the next item to audit.](https://trello.com/c/9gtTPirg/326-when-filtering-by-non-audited-items-saving-an-audit-wont-take-you-to-the-next-item-to-audit)

PR #254, opened 2017-08-03
Remove audit status filter from csv

chao-xian:
Depends on #251

When viewing an audit report, the audit status is removed. From there when we export that audit to CSV from the report screen, the audit status filter was being re-applied. Filtering by audit status is not required in the CSV.

This PR takes advantage of the refactor in #251 where we are able to use a clone of the ContentQuery, ensuring that we don't continue to have the audit status filter.


chao-xian:
[Fix: Report is filtering by audit status](https://trello.com/c/RIporqPl/391-fix-report-is-filtering-by-audit-status)

PR #255, opened 2017-08-03
Extract a `guidance#show` action from AuditsController 

pmanrubia:
It simplifies `AuditsController` by extracting out into a single 
resource action that does not belong to the controller.  

We should be rendering the Govspeak in the view with something 
like `show.govspeak.erb` but this is something to be addressed in the
 future. [I have tracked this work on this Trello card][1]  

[1]: https://trello.com/c/g5hdNO1t/25-render-guidance-content-in-the-view


chao-xian:
[Extract guidance action into its own controller](https://trello.com/c/0MDzbbz9/396-extract-guidance-action-into-its-own-controller)

PR #256, opened 2017-08-04
Fetch Content Items with Locale

alecgibson:
Some content items (approx. 1,600) do not have an “en” locale, which is
the assumed default when querying the Publishing API.

Asking for one of these content items by content ID, but without a valid
locale will result (correctly) in a 404 error from the Publishing API.

In order to address this, when fetching all content IDs from the
`/v2/editions` endpoint, we will also fetch the corresponding locales.
If the “en” locale is present, we use that by default. If it is not
present, we take the first arbitrary locale, so that the content item
is definitely included in the database.

We currently only accept one locale to avoid the race condition of
updating two content items with the same content ID at the same time.

### DIFF
--- a/app/services/clients/publishing_api.rb
+++ b/app/services/clients/publishing_api.rb
@@ -15,20 +15,19 @@ module Clients
       @per_page = 700
     end
 
-    def content_ids
+    def fetch_all(fields)
       publishing_api
         .get_paged_editions(
-          fields: %w[content_id],
+          fields: fields,
           states: %w[published],
           per_page: @per_page,
         )
         .map { |response| normalise(response).fetch(:results) }
         .flatten
-        .map { |result| result.fetch(:content_id) }
     end
 
-    def fetch(content_id)
-      normalise(publishing_api.get_content(content_id))
+    def fetch(content_id, locale)
+      normalise(publishing_api.get_content(content_id, locale: locale))
     end
 
     def links(content_id)
--- a/app/services/content_items_service.rb
+++ b/app/services/content_items_service.rb
@@ -5,13 +5,29 @@ class ContentItemsService
     self.client = Clients::PublishingAPI.new
   end
 
-  def content_ids
-    client.content_ids
+  def fetch_all_with_default_locale_only
+    client
+      .fetch_all(%w[content_id locale])
+      .group_by { |content_item| content_item[:content_id] }
+      .values
+      .map do |content_items_with_the_same_id|
+        content_item_with_en_locale_or_first_other(content_items_with_the_same_id)
+      end
   end
 
-  def fetch(content_id)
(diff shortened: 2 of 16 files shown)


alecgibson:
[Prevent content_id constraint from raising errors](https://trello.com/c/2Ugjs5dh/394-prevent-contentid-constraint-from-raising-errors)

alecgibson:
[Add locale to the request to retrieve content items from Publishing API](https://trello.com/c/fSsnJVwn/393-add-locale-to-the-request-to-retrieve-content-items-from-publishing-api)

app/services/clients/publishing_api.rb
```diff
@@ -15,20 +15,24 @@ def initialize
       @per_page = 700
     end
 
-    def content_ids
+    def all_content_items
```
pmanrubia: The name of this method does not represent very well what it does. I would prefer the client no to group things either. What about something like:

```ruby
def fetch_all_attributes(*attributes)
# returns array [{content_id: "id", locale: "en"},...]
end
```

app/services/clients/publishing_api.rb
```diff
@@ -92,5 +96,13 @@ def publishing_api
     def last_page?(response)
       response["current_page"] == response["pages"]
     end
+
+    def content_item_with_en_locale_or_first_other(content_items)
```
pmanrubia: I would put this logic within the service.

alecgibson:
[Store locale on Content Items](https://trello.com/c/9fTe2hSe/397-store-locale-on-content-items)

app/services/content_items_service.rb
```diff
@@ -5,13 +5,13 @@ def initialize
     self.client = Clients::PublishingAPI.new
   end
 
-  def content_ids
-    client.content_ids
+  def all_content_items
```
pmanrubia: Aligned with my comment above, I would go for something like:

```ruby
def content_ids_with_locale
  all_attributes = client.fetch_all_attributes(:content_id, :locale)
  all_attributes.group_by { |attributes| attributes[:content_id] }
   ... 
  # the decission to pick up an specific locale is implemented at this level
end

File: spec/domain/content/importers/single_content_item_spec.rb [hunk 4]
```diff
@@ -41,7 +43,7 @@ def links
       end
 
       it "imports links" do
-        expect { subject.run(content_id) }
+        expect { subject.run(partial_content_item) }
```
pmanrubia: I would prefer not having something called `partial_content_item`. If we are passing `content_id` and `locale`, let's be explicit in the parameters.

alecgibson:
@pmanrubia I've addressed your comments and re-tested on Integration.

PR #257, opened 2017-08-04
Consolidate publishing api clients

chao-xian:
The client wrapper currently holds 2 instances of the Publishing API client, one instantiated within the wrapper and another initialised on app startup in the initializers folder. We are removing the latter.

Tested import of all content items on local VM and integration as well to be sure that this change doesn't affect the import.


chao-xian:
[The publishing_api adapter is global (lives in an initializer). Move it inside the service.](https://trello.com/c/ZbIKEMRv/400-the-publishingapi-adapter-is-global-lives-in-an-initializer-move-it-inside-the-service)

PR #258, opened 2017-08-04
Add override to running code coverage

chao-xian:
We don't always want to generate the coverage report when we are
running the tests locally during development, as it slows down the tests.

The coverage report will not be generated when running guard. Running
default rake or just rspec will still generate the report, but these can
be overridden with an additional env var.

```
NO_COVERAGE=true bundle exec rspec
```

We can even alias this.

### DIFF
--- a/Guardfile
+++ b/Guardfile
@@ -1,6 +1,6 @@
 require_relative "./lib/guard/lint"
 
-guard :rspec, cmd: 'bundle exec spring rspec' do
+guard :rspec, cmd: 'NO_COVERAGE=true bundle exec spring rspec' do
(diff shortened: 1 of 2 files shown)


chao-xian:
[Rspec tests are taking longer because of test coverage. Run only in integration](https://trello.com/c/P4rs4z6K/401-rspec-tests-are-taking-longer-because-of-test-coverage-run-only-in-integration)

alecgibson:
This change will only affect people running Guard, which isn't necessary for running tests (which can also be run with `rake` or simply `rspec`).

Closing this pull request in favour of: https://github.com/alphagov/content-performance-manager/pull/261

PR #259, opened 2017-08-04
Remove AllInventory importer

chao-xian:
This importer is unnecessary as it only calls Importers::AllContentItems.

The way we generate an inventory has yet to be decided so we're removing
this code now.


chao-xian:
[Remove unused load-all content items code](https://trello.com/c/aop4pVxD/402-remove-unused-load-all-content-items-code)

PR #260, opened 2017-08-07
Simplify audit controller by extracting out domain objects

pmanrubia:
Trello cards:

* https://trello.com/c/mYlFZlO4/403-extract-report-controller-from-audit-controller
* https://trello.com/c/dtLgLHJr/404-simplify-download-action-from-audit-controller
* https://trello.com/c/RiijjTbM/405-extract-logic-to-find-the-next-item

This is the first iteration towards simplifying the `AuditController`, and having a better domain model.

Previously the AuditController was doing too much. We have already [extracted the guidance/support into its own controller][1] but still, there is a long way to go.

I am addressing the following changes:

1. Extract a single resource report controller (`Audits::ReportController`) to show the progress of an Audit
2. Use `respond_to` instead of creating a new action to download a `csv`. This removes the action `export` from the `AuditsController`
3. Encapsulate use case to find the next item when navigating through a collection (`Audits::FindNextItem`)
4. Encapsulate code into a domain object (`Audits::Progress`) to retrieve information about an audit progress
5. Extract domain objects to model the state of an audit (audited/non-audited/passing/failing). `Policies::Audited`, `Policies::NonAudited`, `Policies::Failing`, `Policies::Passing`
6. Encapsulate into a domain object (`Audits::FindContent`) the use case to find content items to audit
7. Unify into (`Audits::Filter`) the user criteria to define a subset of content items to audit. 

Regarding our domain objects... the dev team still have a long journey to better define class names and responsibilities, but this PR move us into a better position as it extracts out logic that does not belong to a controller and places it the domain folder. So this is not a final definition of our domain model, just a step forward

Some considerations:

1. I have defined a domain object per `use case`. So far, `FindNextItem` and `FindContent`. 
2. I have defined `policy` objects so our domain objects speak the same vocabulary as our users. We should be able to use our application without a controller.
3. We still need to think carefully how we query for data in the domain layer, and how we express those queries. We should be able to chain them on an expressive way. I have not addressed this issue in this PR
4. Previously code was based on feature test covering. We will have a conversation shortly about this approach, but I would like to have unit tests for our domain objects, regardless of being already covered by integration tests.

[1]: https://github.com/alphagov/content-performance-manager/pull/255


app/domain/audits/find_content.rb
```diff
@@ -0,0 +1,23 @@
...
+                .page(filter.page)
+                .organisations(filter.organisations, filter.primary_org_only)
+                .document_type(filter.document_type)
+                .after(filter.after)
+                .theme(filter.theme_id)
+                .scope
+
+      return scope if filter.audit_status.blank?
+
+      if filter.audit_status.to_sym == :audited
```
andrewgarner: We're evaluating `filter.audit_status` twice here so we could use a case statement:
 
```rb
case filter.audit_status.to_sym
when :audited
  Policies::Audited.call(scope)
when :non_audited
  Policies::NonAudited.call(scope)
else
  scope
end
```

app/views/audits/reports/show.html.erb
```diff
@@ -0,0 +1,61 @@
+<% content_for :sidebar do %>
+    <%= render 'sidebar', submit_path: audits_report_path %>
```
andrewgarner: Extra indentation.

app/views/audits/reports/show.html.erb
```diff
@@ -0,0 +1,61 @@
...
+<% end %>
+
+<% content_for :header do %>
+    <ul class="nav nav-tabs">
```
andrewgarner: Extra indentation.

app/domain/audits/find_content.rb
pmanrubia: I have addressed this in 9d245286b7cf774a723f0b13aeb43b326749ff70
I have extract a nil object to deal with the `nil` | `blank` situation. It also make `FindContent` agnostic of the list of values for audit_status

app/views/audits/reports/show.html.erb
pmanrubia: Addressed in 17f1dbdcbca9f49c308738712a140991a41a9cf2

app/views/audits/reports/show.html.erb
pmanrubia: Addressed in 17f1dbdcbca9f49c308738712a140991a41a9cf2

pmanrubia:
@andrewgarner thank you for your review, I have addressed your comments.

PR #261, opened 2017-08-07
Only run code coverage reporting on Jenkins

alecgibson:
Running code coverage reporting in development can slow down running
tests, especially as the reporting runs on even single tests (as well
as the whole suite).

This change only runs the test coverage reporting if the environment
variable `TEST_COVERAGE` is `”true”`, and sets this appropriately in
the `Jenkinsfile`, so that test coverage reporting is run on Jenkins.

Test coverage can be run locally by running, for example:
`TEST_COVERAGE=true bundle exec rake`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,5 +2,9 @@
 
 node {
   def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
-  govuk.buildProject()
+  govuk.buildProject(
+    beforeTest: {
+      govuk.setEnvar("TEST_COVERAGE", "true")
+    }
+  )
 }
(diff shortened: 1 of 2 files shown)


alecgibson:
[Rspec tests are taking longer because of test coverage. Run only in integration](https://trello.com/c/P4rs4z6K/401-rspec-tests-are-taking-longer-because-of-test-coverage-run-only-in-integration)

andrewgarner:
I think it would be better to explicitly set an environmental variable that can also be optionally set locally if you so wish.

alecgibson:
@andrewgarner yeah I was thinking about doing that; I can add one if we think it'll actually get used?

andrewgarner:
Possibly/possibly not. Something like `ENV['COVERAGE']` is environment agnostic though which I think is a good thing.

PR #263, opened 2017-08-08
Fix Organisation Links on Content Items

alecgibson:
Since we changed paths to use Content IDs, the organisations presented
on a Content Item have had broken links, because they still use the
ID.

This change fixes them to use the Content IDs, and adds a feature test.


alecgibson:
[BUG: show organisation in CPM after using ID](https://trello.com/c/OHXzSe2b/410-bug-show-organisation-in-cpm-after-using-id)

PR #264, opened 2017-08-08
Extract Pageview Importer into its own class

alecgibson:
This change extracts the Pageview Importer into its own class, similar
to the ContentItem importer.


alecgibson:
[Extract code from Sidekiq job to import pageviews](https://trello.com/c/OdxXOjez/411-extract-code-from-sidekiq-job-to-import-pageviews)

PR #265, opened 2017-08-08
Return 404 for non-existent content items

alecgibson:
The `find_by` method returns `nil` if a model cannot be found. We then
try to run `decorate` on `nil`, which causes a `500` error. This is
incorrect, and we should return a `404`.

Using `find_by!` will instead raise an `ActiveRecord::RecordNotFound`
error, which returns a `404` status, and a default “Not Found” page.

### DIFF
--- a/app/controllers/audits/audits_controller.rb
+++ b/app/controllers/audits/audits_controller.rb
@@ -40,7 +40,7 @@ module Audits
     end
 
     def content_item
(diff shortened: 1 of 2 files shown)


alecgibson:
[Return 404 for non-existent content items](https://trello.com/c/J5Z2oy1h/415-return-404-for-non-existent-content-items)

PR #266, opened 2017-08-08
Specify paging on `FindContent`

alecgibson:
Different consumers of `FindContent` need paged, or unpaged content,
depending on what the task is. For example, showing a list of content
items should be paged, but exporting a CSV should show all content
items.

This change replaces the single `FindContent#call` method with two
other methods for specifying whether the consumer wants paged content,
or all content items.

This fixes a bug with the CSV exporter, where it was only exporting
a single page of content items.

### DIFF
--- a/app/domain/audits/find_content.rb
+++ b/app/domain/audits/find_content.rb
@@ -1,15 +1,23 @@
 module Audits
   class FindContent
-    def self.call(filter = Filter.new)
-      scope = Content::Query.new
-                .page(filter.page)
-                .organisations(filter.organisations, filter.primary_org_only)
-                .document_type(filter.document_type)
-                .after(filter.after)
-                .theme(filter.theme_id)
-                .scope
-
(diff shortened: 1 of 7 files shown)


alecgibson:
[Fix CSV Export](https://trello.com/c/W9jJErRx/416-fix-csv-export)

app/domain/audits/find_content.rb
```diff
@@ -1,15 +1,23 @@
...
+      scope = query_from_filter(filter).content_items
+      filter.audited_policy.call(scope)
+    end
 
+    def self.all(filter)
+      scope = query_from_filter(filter).all_content_items
       filter.audited_policy.call(scope)
     end
+
+    def self.query_from_filter(filter)
```
pmanrubia: What about renaming it to: 
```ruby
def self.query(filter)
```

`from_filter` is implicit in the parameter name.

PR #267, opened 2017-08-08
Move Content-related models into a subdirectory

alecgibson:
This change moves the following classes from `models` into a
`models/content` subdirectory, with a corresponding `Content` module:

* `ContentItem`
* `Group`
* `Link`


alecgibson:
[Move Content Models into a subdirectory](https://trello.com/c/tjGBH538/417-move-content-models-into-a-subdirectory)

app/models/content/content_item.rb
```diff
@@ -0,0 +1,56 @@
+module Content
+  class ContentItem < ApplicationRecord
```
andrewgarner: @pmanrubia - what are your thoughts on `Content::Item`?

alecgibson: I'd considered doing that. I don't mind either way. I guess it just gets a bit weird if you refer within the module to just `Item`s

andrewgarner: I think that is okay. There is always the option of fully qualifying it with `::Content::Item` where just `Item` might be confusing.

alecgibson:
@andrewgarner they're showing as additions/deletions because the whole file has been nested inside a `module` - I don't know if there's any way to get around that without doing it in two (unatomic) commits?

app/models/content/content_item.rb
alecgibson: @andrewgarner done.

andrewgarner:
@alecgibson yep. I'm just reading through https://cirw.in/blog/constant-lookup.html for ideas. I wonder if maintaining git history is better that an atomic commit in this instance?

app/models/content/content_item.rb
pmanrubia: 👍 to Content::Item

PR #271, opened 2017-08-10
Filter allocated content to audit

pmanrubia:
[Trello card](https://trello.com/c/dOMctS8m/426-3-show-content-allocated-to-myself)

Supports content filtering by allocation. We will be able to list all 
content items that are assigned to myself, that has not been
assigned, or complete ignore the filter. 

I have split the PR in two commits because although the 
domain layer should remain the same, the UI is under user 
research.
 
```ruby
  filter = Filter.new(allocated_to: params[:allocated_to])
  result = Audits::FindContent.all(filter)
```
### Before

image

### After

image

### DIFF
--- a/app/domain/audits/filter.rb
+++ b/app/domain/audits/filter.rb
@@ -1,7 +1,7 @@
 module Audits
   class Filter
     include ActiveModel::Model
-    attr_accessor :theme_id, :page, :per_page, :organisations, :document_type, :audit_status, :primary_org_only, :after
+    attr_accessor :theme_id, :page, :per_page, :organisations, :document_type, :audit_status, :primary_org_only, :after, :allocated_to
 
     def audit_status=(value)
       @audit_status = if value.blank?
@@ -9,6 +9,16 @@ module Audits
                       else
                         value.to_sym
                       end
+    end
+
+    def allocated_policy
+      if allocated_to == 'no_one'
+        Policies::Unallocated
+      elsif allocated_to.blank?
+        Policies::AllocatedAndUnallocated
+      else
+        Policies::Allocated
+      end
     end
 
     def audited_policy
--- a/app/domain/audits/filter.rb
+++ b/app/domain/audits/filter.rb
@@ -1,7 +1,15 @@
 module Audits
   class Filter
     include ActiveModel::Model
-    attr_accessor :theme_id, :page, :per_page, :organisations, :document_type, :audit_status, :primary_org_only, :after, :allocated_to
+    attr_accessor :after,
+      :allocated_to,
+      :audit_status,
+      :document_type,
+      :organisations,
+      :page,
+      :per_page,
+      :primary_org_only,
+      :theme_id
 
     def audit_status=(value)
       @audit_status = if value.blank?
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,10 +10,19 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170804092444) do
+ActiveRecord::Schema.define(version: 20170810074052) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
+
+  create_table "allocations", force: :cascade do |t|
+    t.string "content_id", null: false
+    t.string "uid", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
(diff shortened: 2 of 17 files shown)


db/schema.rb
```diff
@@ -10,11 +10,20 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170804092444) do
+ActiveRecord::Schema.define(version: 20170810074052) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "allocations", force: :cascade do |t|
```
andrewgarner: This is different to the schema for Audits. I think we should be consistent in how we model User relationships.

```rb
create_table "audits", id: :serial, force: :cascade do |t|
  t.string "content_id", null: false
  t.string "uid", null: false
  t.datetime "created_at", null: false
  t.index ["content_id"], name: "index_audits_on_content_id", unique: true
  t.index ["uid"], name: "index_audits_on_uid"
end
```

app/domain/audits/policies/non_allocated.rb
```diff
@@ -0,0 +1,9 @@
+module Audits
+  module Policies
+    class NonAllocated
+      def self.call(scope, allocated_to: nil) # rubocop:disable Lint/UnusedMethodArgument
+        scope.where.not(id: Allocation.select(:content_item_id))
```
andrewgarner: I think you can do the following which avoids the sub query:

```rb
scope.left_joins(:allocations).where(allocations: { id: nil })
```

app/domain/audits/policies/non_allocated.rb
```diff
@@ -0,0 +1,9 @@
+module Audits
+  module Policies
+    class NonAllocated
```
andrewgarner: I think `Unallocated` would be the correct name to use here based on [the dictionary](http://www.dictionary.com/browse/unallocated)!

db/schema.rb
pmanrubia: Addressed in 4a77e05d5afce90c9fcfe25d47e17c6e5ee05780

```ruby
  create_table "allocations", force: :cascade do |t|
    t.string "content_id", null: false
    t.string "uid", null: false
    t.datetime "created_at", null: false
    t.index ["content_id"], name: "index_allocations_on_content_id", unique: true
    t.index ["uid"], name: "index_allocations_on_uid"
  end

  create_table "audits", id: :serial, force: :cascade do |t|
    t.string "content_id", null: false
    t.string "uid", null: false
    t.datetime "created_at", null: false
    t.index ["content_id"], name: "index_audits_on_content_id", unique: true
    t.index ["uid"], name: "index_audits_on_uid"
  end
```

app/domain/audits/policies/non_allocated.rb
pmanrubia: Addressed in 4a77e05d5afce90c9fcfe25d47e17c6e5ee05780

app/domain/audits/policies/non_allocated.rb
pmanrubia: 😊  Addressed in 4a77e05d5afce90c9fcfe25d47e17c6e5ee05780

pmanrubia:
@andrewgarner thank you for your review. 
I have addressed your comments. PR looks better now 👍

app/domain/audits/filter.rb
```diff
@@ -11,6 +19,16 @@ def audit_status=(value)
                       end
     end
 
+    def allocated_policy
+      if allocated_to == 'no_one'
+        Policies::Unallocated
+      elsif allocated_to.blank?
+        Policies::AllocatedAndUnallocated
```
andrewgarner: One further thing that has just occurred to me ...

`Policies::Unallocated` is supposed to be the source of truth as to what an unallocated content item is. Likewise `Policies::Allocated` for allocated. So when we want both allocated and unallocated we should apply both policies.

pmanrubia: Hi Andrew, not sure about this one (I chose`Policies::AllocatedAndUnallocated` as a way to name a `NilObject`)
But this might be because I am not able to picture your solution in code. What about tracking this idea in its own separate card/PR? If we go for this way of doing things we would also need to update the `AuditedAndNonAudited` policy.

andrewgarner: I'll branch of this and see what I can come up with in code.

PR #274, opened 2017-08-11
Do not order Content Items in CSV reports

alecgibson:
Ordering Content Items approximately doubles the length of time required
to execute the query.

When selecting a large number of columns, as we do for generating a
report, this can slow down the query enough to cause `504` Gateway
Timeout errors.

This change removes the ordering from CSV report queries in order to
speed up its performance and prevent `504` errors.

### DIFF
--- a/app/domain/audits/report.rb
+++ b/app/domain/audits/report.rb
@@ -44,7 +44,10 @@ module Audits
     end
 
     def rows
-      content_items.joins(:report_row).pluck(:data)
+      content_items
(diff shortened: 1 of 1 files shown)


alecgibson:
[Export CSV 504s sometimes](https://trello.com/c/iU9wwc1Q/443-export-csv-504s-sometimes)

pmanrubia:
👍 Nice one @alecgibson

PR #279, opened 2017-08-15
Add support for feature toggles

andrewgarner:
Adds support for feature toggles using the https://github.com/mgsnova/feature library.


alecgibson:
What benefits does this give us over just using environment variables? Especially given that you're already using one here?

pmanrubia:
What are your thoughts about using the `RedisRepository`?

```ruby
repo = Feature::Repository::RedisRepository.new("feature_toggles")

# add/toggle features in Redis
Redis.current.hset("feature_toggles", "ActiveFeature", true)
Redis.current.hset("feature_toggles", "InActiveFeature", false)
```

This would allow us to enable/disable features just opening a Rails 
console in each environment if we wish to.

alecgibson:
>  just opening a Rails console

Possibly a bit less secure / auditable? Depends on how much you care about that, though.

andrewgarner:
@alecgibson it gives an API for working with features:

```rb
Feature.active?(:feature_name) # => true/false

Feature.inactive?(:feature_name) # => true/false

Feature.active_features # => [:list, :of, :features]

Feature.with(:feature_name) do
  # code
end

Feature.without(:feature_name) do
  # code
end

# this returns value_true if :feature_name is active, otherwise value_false
Feature.switch(:feature_name, value_true, value_false) 

# switch may also take Procs that will be evaluated and it's result returned.
Feature.switch(:feature_name, -> { code... }, -> { code... })
```

And an API for testing behaviour with and without a feature:

```rb
Feature.run_with_activated(:feature) do
  # your test code
end

# you also can give a list of features
Feature.run_with_deactivated(:feature, :another_feature) do
  # your test code
end
```

alecgibson:
I just wonder if the all this extra config boiler plate is over-complicating things for a syntax that's only _slightly_ nicer than `if ENV['MY_FEATURE_ENABLED'].present?`.

andrewgarner:
@pmanrubia I went with a simple backend to start with knowing that we can easily swap it out for something else should we need to in the future.

The YAML backend gives us a bit more visibility and an audit trail for which features we have and whether they are enabled or disabled across environments which I think is a good thing to start with.

andrewgarner:
@alecgibson `ENV` is a primitive. I think it's a good thing to use a domain object such as `Feature`. We could write our own but there isn't a lot of code in the gem so I see no reason to. If we were to adopt a pure `ENV` approach then we have to think about things like how we handle defaults, how we limit surprises by ensuring everyone runs the application from a consistent environment.

alecgibson:
> a consistent environment

Isn't this exactly what the VM should be for? Although I think basically no one on the team other than me uses it...?

But yes, given that a bunch of us don't use the VM, maybe this is better. Alternatively could use something like [dotenv](https://github.com/bkeepers/dotenv), but by the time we're using a gem, we may as well use a gem that actually does what we want.

PR #282, opened 2017-08-17
Fix no filters on csv

chao-xian:
Query parameters were not set on the link for exporting a list of content items to CSV. Hence users were always exporting the entire 300k of content items to CSV.


chao-xian:
[Export CSV doesn't seem to use filters](https://trello.com/c/TlDvElgB/446-export-csv-doesnt-seem-to-use-filters)

PR #283, opened 2017-08-17
Allocation feature

andrewgarner:
This adds the ability to assign a content item audit. I've wrapped the feature in a toggle so that we can get this feature branch merged into master.

### DIFF
--- a/app/controllers/audits/allocations_controller.rb
+++ b/app/controllers/audits/allocations_controller.rb
@@ -5,23 +5,29 @@ module Audits
     end
 
     def create
-      content_ids = params.fetch(:content_ids)
-      user_uid = params.fetch(:allocate_to)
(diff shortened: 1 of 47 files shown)


andrewgarner:
[(3) Show content allocated to myself](https://trello.com/c/dOMctS8m/426-3-show-content-allocated-to-myself)

andrewgarner:
[(3) Allocate content items within current page](https://trello.com/c/Rcyk9UU9/429-3-allocate-content-items-within-current-page)

alecgibson:
Do you think it's worth adding some test cases to check that certain elements don't turn up if the feature is inactive? It would basically be testing the gem, so it depends on how critical we think it is that these things aren't shown to end-users.

andrewgarner:
I did think that. I think we're okay here. We need to be prudent in the way we use the toggles or we'll find ourselves having to run the test suite for every combination of toggle.

PR #286, opened 2017-08-18
Use feature environment config directly

andrewgarner:
This prevents a scenario where the value of the variable could be false yet we'd still enable the toggle.

### DIFF
--- a/config/features.yml
+++ b/config/features.yml
@@ -13,4 +13,4 @@ test:
 production:
   <<: *default
   features:
-    auditing_allocation: <%= ENV['FEATURE_AUDITING_ALLOCATION'].present? %>
+    auditing_allocation: <%= ENV['FEATURE_AUDITING_ALLOCATION'] %>


andrewgarner:
[Add feature toggle to disable allocation in production, and enable in staging and integration](https://trello.com/c/r3iKNHcw/449-add-feature-toggle-to-disable-allocation-in-production-and-enable-in-staging-and-integration)

alecgibson:
@andrewgarner have you tested all these changes together with the puppet changes to make sure it doesn't do anything weird like assume that "true" is a string, rather than a boolean?

andrewgarner:
@alecgibson I've tested the Puppet variable:

```command-line
vagrant@development:/var/govuk/govuk-puppet/development-vm$ cat /etc/govuk/content-performance-manager/env.d/FEATURE_AUDITING_ALLOCATION
true
```

and running the application on the VM with that variable set.

PR #287, opened 2017-08-18
Fix column widths on audit screen

chao-xian:
Although we hide the filter options in the sidebar when viewing the audit
screen, we had lost the logic to collapse the width of that column and
there was a big gap.

## Before

screenshot-content-performance-manager integration publishing service gov uk-2017-08-18-12-12-16

## After

screenshot-localhost-3000-2017-08-18-12-12-05

### DIFF
--- a/app/views/layouts/audits.html.erb
+++ b/app/views/layouts/audits.html.erb
@@ -32,7 +32,7 @@
     </div>
   </div>
   <div class="row">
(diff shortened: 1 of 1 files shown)


chao-xian:
[Fix column widths on Audit screen](https://trello.com/c/mq5xWRdz/454-fix-column-widths-on-audit-screen)

PR #288, opened 2017-08-18
Show assigned auditor

chao-xian:
When a content item is allocated to a person to audit, it needs to be reflected on the audit screen who it is.


chao-xian:
[(1) Show allocation details in the show page of an audit](https://trello.com/c/4WZFxYMJ/430-1-show-allocation-details-in-the-show-page-of-an-audit)

PR #289, opened 2017-08-18
Allow all content items to be assigned on a page

andrewgarner:


andrewgarner:
[(3) Allocate all content items](https://trello.com/c/B2byRQkl/428-3-allocate-all-content-items)

spec/features/audit/allocation_spec.rb
```diff
@@ -104,4 +104,19 @@
     expect(page).to have_select("allocate_to", selected: "No one")
     expect(page).to have_content("1 items unallocated")
   end
+
+  scenario 'Allocate all content within current page', :js do
+    create_list(:content_item, 100)
```
pmanrubia: Creating a list of 100 seems a bit too much. What about creating 27 items (page size + a couple of them more)?

app/assets/javascripts/select-all.js
```diff
@@ -0,0 +1,23 @@
+(function (Modules) {
+  "use strict";
+
+  Modules.SelectAll = function () {
+    var that = this;
+
+    that.start = function (element) {
```
pmanrubia: I would say that we should use `this` at this point, and use `that` inside the event handler.
But, as we are not using the reference to `this` in lines 13-15 we could get rid of `that`

chao-xian: Addressed in https://github.com/alphagov/content-performance-manager/pull/290/commits/d4fd166a4ef59282c85bf2d178a0748e0c77dbf3#diff-1bf7091d9f272b27a4976caf9081ac8cR5

spec/features/audit/allocation_spec.rb
chao-xian: Addressed in https://github.com/alphagov/content-performance-manager/pull/290/commits/d4fd166a4ef59282c85bf2d178a0748e0c77dbf3#diff-69c737006735f69541cec3ef0afaea49R109

PR #290, opened 2017-08-18
Audit select all on all pages

andrewgarner:
Allows all content items that match the current filters to be allocated, regardless of pagination.

I've added a dependency on https://github.com/zdennis/activerecord-import so that we can handle the potential to create thousands of allocation database records in a single trip to the server. This will need more serious consideration if the prototype tests well.

I've refactored the allocation service to use IDs and not objects to avoid hitting the database.

One potential UX improvement that we could introduce is to keep the checkboxes checked on subsequent pages if the user checks the "Select X items on all pages" then navigates to another page.

### DIFF
--- a/app/assets/javascripts/select-all.js
+++ b/app/assets/javascripts/select-all.js
@@ -2,9 +2,7 @@
   "use strict";
 
   Modules.SelectAll = function () {
-    var that = this;
-
-    that.start = function (element) {
+    this.start = function (element) {
       function addEventListener() {
         var selectAll = element.find('#select_all');
 
--- a/app/assets/javascripts/select-all.js
+++ b/app/assets/javascripts/select-all.js
@@ -5,12 +5,22 @@
     this.start = function (element) {
       function addEventListener() {
         var selectAll = element.find('#select_all');
+        var selectAllPages = element.find('#select_all_pages');
+        var selectAllPagesContainer = selectAllPages.closest('.if-js-hide');
 
         if (selectAll != undefined) {
           selectAll.on('change', function (e) {
             var checked = selectAll.prop('checked');
             var checkboxes = $('.select-content-item');
             checkboxes.prop('checked', checked);
+
+            if (selectAllPages != undefined) {
+              selectAllPages.prop('checked', false);
+            }
+
+            if (selectAllPagesContainer != undefined) {
+              selectAllPagesContainer.toggleClass('if-js-hide', !checked);
+            }
           });
         }
       }
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem 'govuk_sidekiq'
 gem 'plek'
 
 # Third party gems
+gem 'activerecord-import'
 gem 'draper'
 gem 'feature'
 gem 'google-api-client', '~> 0.9'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -51,6 +51,8 @@ GEM
       activemodel (= 5.1.2)
       activesupport (= 5.1.2)
       arel (~> 8.0)
+    activerecord-import (0.19.1)
+      activerecord (>= 3.2)
     activesupport (5.1.2)
       concurrent-ruby (~> 1.0, >= 1.0.2)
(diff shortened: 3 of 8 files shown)


andrewgarner:
[(3) Allocate all content items](https://trello.com/c/B2byRQkl/428-3-allocate-all-content-items)

app/assets/javascripts/select-all.js
```diff
@@ -5,12 +5,22 @@
     this.start = function (element) {
       function addEventListener() {
         var selectAll = element.find('#select_all');
+        var selectAllPages = element.find('#select_all_pages');
+        var selectAllPagesContainer = selectAllPages.closest('.if-js-hide');
```
chao-xian: `var selectAllPagesContainer = element.find('.checkbox.if-js-hide');` should just work actually so I might switch it to that.

chao-xian: Ah I've just discovered with @alecgibson that `.find()` is jQuery in action so the `.closest()` should be provided by jQuery.

app/assets/javascripts/select-all.js
```diff
@@ -2,17 +2,25 @@
   "use strict";
 
   Modules.SelectAll = function () {
-    var that = this;
-
-    that.start = function (element) {
+    this.start = function (element) {
       function addEventListener() {
         var selectAll = element.find('#select_all');
+        var selectAllPages = element.find('#select_all_pages');
```
alecgibson: Could you please prefix jQuery variables with `$`, as demonstrated in our [documentation](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#module-structure).

app/assets/javascripts/select-all.js
```diff
@@ -2,17 +2,25 @@
...
+        var selectAllPagesContainer = selectAllPages.closest('.if-js-hide');
 
         if (selectAll != undefined) {
           selectAll.on('change', function (e) {
             var checked = selectAll.prop('checked');
             var checkboxes = $('.select-content-item');
             checkboxes.prop('checked', checked);
+
+            if (selectAllPages != undefined) {
```
alecgibson: Is this _ever_`false`? I thought jQuery's [`.find()`](https://api.jquery.com/find/) _always_ returned an [`jQuery`](https://api.jquery.com/Types/#jQuery) object? Shouldn't you check `if (selectAllPages.length)` ?

app/assets/javascripts/select-all.js
```diff
@@ -2,17 +2,25 @@
...
+              selectAllPages.prop('checked', false);
+            }
+
+            if (selectAllPagesContainer != undefined) {
```
alecgibson: As above.

app/assets/javascripts/select-all.js
```diff
@@ -2,17 +2,25 @@
...
-    var that = this;
-
-    that.start = function (element) {
+    this.start = function (element) {
       function addEventListener() {
         var selectAll = element.find('#select_all');
+        var selectAllPages = element.find('#select_all_pages');
+        var selectAllPagesContainer = selectAllPages.closest('.if-js-hide');
 
         if (selectAll != undefined) {
```
alecgibson: This should also be changed to check `.length` (see below)

alecgibson:
@andrewgarner @chao-xian sorry I got around to reviewing this after it got merged, but I think we need to look at some of our `if` statements - I don't think they work properly.

chao-xian:
@alecgibson OK I'll open another PR

app/assets/javascripts/select-all.js
alecgibson: For example:
```javascript
$('body').find('non-existent-element') == undefined // false
```

andrewgarner:
@alecgibson thanks for the feedback. I built on the approach in https://github.com/alphagov/content-tagger/blob/master/app/assets/javascripts/select-all.js –
 aiming for good enough for research. Thanks for your recommendations though, perhaps we can apply them in Content Tagger too?

PR #292, opened 2017-08-22
Rake task to build term generation report

whoojemaflip:
Rake task and calculator function to generate and save a CSV for the "rate of return of page review"

[Trello](https://trello.com/c/x6GUpctA/173-generate-graph-showing-rate-of-term-generation-for-coming-to-the-uk)

### DIFF
--- a/app/domain/term_generation/rate_of_discovery.rb
+++ b/app/domain/term_generation/rate_of_discovery.rb
@@ -1,0 +1,46 @@
+module TermGeneration
+  module RateOfDiscovery
+    WINDOW_SIZE = 50
+
+    # Calculate the moving-window average of the absolute change
+    # in new term discovery for a project. Aims to give insight
+    # into the expected reward of reviewing a single extra page.
+
+    def self.calculate(project)
+      results = []
+
+      # Keeps track of previously seen unique terms.
+      terms = Set.new
+
+      # Iterate over previously completed Todos in completion order.
+      project_todos(project).each do |todo|
+        # Note how many terms have been encountered thus far.
+        count_of_previously_found_terms = terms.size
+
+        # Add the terms from this Todo to the set of all terms.
+        terms.merge Set.new(todo.terms)
+
+        # Record the absolute change in number of terms encountered for this Todo.
+        results << {
+          count_of_new_terms: terms.size - count_of_previously_found_terms
+        }
+
+        # Our window is the last 50 results.
+        moving_window = results.last(WINDOW_SIZE)
+
+        # Calculate the absolute change in terms found over the course of the window.
+        total_change = moving_window.sum { |r| r[:count_of_new_terms] }
+
+        # Average the result. Expected values are roughly between 0.3 and 1.5
+        results[-1][:moving_average] = total_change.to_f / moving_window.size
+      end
+
+      # Return an array of floating point values.
+      results.map { |r| r[:moving_average] }
+    end
+
+    def self.project_todos(project)
+      project.taxonomy_todos.done.order(completed_at: :asc).includes(:terms)
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


lib/tasks/term_generation/calculate_project_statistics.rake
```diff
@@ -0,0 +1,35 @@
...
+    terms = Set.new([])
+    project.taxonomy_todos.done.order(completed_at: :asc).each do |todo|
+      count_of_previously_found_terms = terms.size
+      terms.merge Set.new(todo.terms)
+      todo.update_attribute(:count_of_new_terms, terms.size - count_of_previously_found_terms)
+
+      previous50 = TaxonomyTodo.done
+                                .order(completed_at: :desc)
+                                .where("completed_at <= ?", todo.completed_at)
+                                .limit(50)
```
koetsier: could you make the '50' a constant/parameter? It seems a bit hard-coded like this.

lib/tasks/term_generation/calculate_project_statistics.rake
```diff
@@ -0,0 +1,35 @@
+namespace :term_generation do
+  desc "Calculate statistics for a Term Generation project"
+  task :calculate_project_statistics, [:project_id] => :environment do |_, args|
+    project = TaxonomyProject.find(args.project_id)
```
koetsier: You could extract the logic of updating the statistics to a separate service object or maybe the model itself so you can write unit tests

lib/tasks/term_generation/calculate_project_statistics.rake
whoojemaflip: I don't think it makes any sense as a parameter, unless we intend to build a generic "moving window average" algorithm.

app/domain/term_generation/rate_of_discovery.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    # Calculate the moving-window average of the absolute change
+    # in new term discovery for a project. Aims to give insight
+    # into the expected reward of reviewing a single extra page.
+
+    def self.calculate(project)
+      results = []
+
+      # Keeps track of previously seen unique terms.
+      terms = Set.new([])
```
koetsier: can you not just call Set.new()?

whoojemaflip: updated

app/domain/term_generation/rate_of_discovery.rb
```diff
@@ -0,0 +1,42 @@
...
+        # Add the terms from this Todo to the set of all terms.
+        terms.merge Set.new(todo.terms)
+
+        # Record the absolute change in number of terms encountered for this Todo.
+        results << {
+          count_of_new_terms: terms.size - count_of_previously_found_terms
+        }
+
+        # Our window is the last 50 results.
+        moving_window = results.reverse[0..WINDOW_SIZE - 1]
```
koetsier: can you not use results.last(WINDOW_SIZE)?

whoojemaflip: I had a total blind spot for `Enum#last` thanks for pointing that out!

app/domain/term_generation/rate_of_discovery.rb
```diff
@@ -0,0 +1,42 @@
...
+
+        # Calculate the absolute change in terms found over the course of the window.
+        total_change = moving_window.sum { |r| r[:count_of_new_terms] }
+
+        # Average the result. Expected values are roughly between 0.3 and 1.5
+        results[-1][:moving_average] = (total_change * 1.0) / moving_window.size
```
koetsier: is there an advantage over total_change.to_f?

whoojemaflip: Force of habit. Updated.

PR #296, opened 2017-08-23
Re-enable feature toggle control of theme filter

chao-xian:
Removes the temporary commenting out of the theme filters from the UI,
and re-enables use of the feature toggle. Also re-enables the associated
tests.

Addresses #295


chao-xian:
[Hide "Theme" filter options](https://trello.com/c/JWXZSDtI/456-hide-theme-filter-options)

PR #297, opened 2017-08-24
Remove model for Groups

pmanrubia:
[Trello card](https://trello.com/c/wWeN72af/474-remove-groups)

Groups are no longer in use. We rely on subscription rules to group
content items. 

There are no Groups in production, so I am confident that deleting them
won’t have an impact in the tool.

### DIFF
--- a/db/migrate/20170824142643_drop_groups.rb
+++ b/db/migrate/20170824142643_drop_groups.rb
@@ -1,0 +1,12 @@
+class DropGroups < ActiveRecord::Migration[5.1]
+  def change
+    drop_table :groups do |t|
+      t.string :slug
+      t.string :name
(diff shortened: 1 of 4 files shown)


db/migrate/20170824142643_drop_groups.rb
```diff
@@ -0,0 +1,12 @@
+class DropGroups < ActiveRecord::Migration[5.1]
+  def change
+    drop_table :groups do |t|
+      t.string :slug
```
chao-xian: What does specifying the columns in the block for `drop_table` do?

pmanrubia: It allows a rollback of the migration

PR #299, opened 2017-08-24
Replace Organisation filter with an auto-complete box

alecgibson:
This change progressively enhances the Organisation filter `select`
element with [`alphagov/accessible-autocomplete`](https://github.com/alphagov/accessible-autocomplete).

As the name suggests, this change gives us an accessible approach to
offering auto-complete on the long Organisations field.

In order to accommodate the change, JavaScript files must now be
explicitly included in the `application.js` manifest, because we now
have load-order dependency in our scripts.

The change adds JavaScript and CSS from the accessible autocomplete
project. However, the styling has been modified to fit in better with
surrounding Boostrap styling.

A polyfill for `.bind` is also included, because this function is not
supported by older browser, including PhantomJS, which we use to run
feature tests.

Finally, this change also bumps our version of `govuk-lint`, in order to
allow BEM Sass styling, which is used by the autocomplete component.


alecgibson:
[(8) Search as you type for organisations to filter by](https://trello.com/c/HEzdGpyZ/469-8-search-as-you-type-for-organisations-to-filter-by)

alecgibson:
Closed in favour of: https://github.com/alphagov/content-performance-manager/pull/302 (renamed branch)

PR #300, opened 2017-08-24
Fix failing migration

alecgibson:
Following a refactor that renamed the `ContentItem` model, this
migration no longer works.

For exactly this reason, models should not be referenced in migrations.
This change fixes the broken migration by executing raw SQL rather than
referring to the content model.


alecgibson:
@chao-xian duly noted - great tip.

PR #306, opened 2017-08-29
[DO NOT MERGE] Script to enhance screaming frog reports

MatMoore:
This code probably won't be merged - it's to help the search team analyse some data on titles and meta descriptions. Once we've done this we'll be able to make some recommendations from an SEO perspective. Hopefully this can inform future content audits.

That said, if it would be useful to integrate any of this with the app in some way I'm happy to chat about this further.

Trello: https://trello.com/c/PrJ2QXir/259-analyse-crawler-output

The rake task takes CSV output from the Screaming Frog SEO Spider, and cross references it with data in content performance manager.

This will hopefully help us prioritise the most problematic content from an SEO perspective.

For example, using the document type we can prioritise improving guidance content.

If content has very low pageviews and Screaming Frog has also flagged issues with titles/descriptions, we could flag those for review. On the other hand, if content is relatively popular, and Screaming Frog still flags it, small in provements to SEO could affect a large number of users.

I've also include "Primary Publishing Organisation" which is sometimes an arbitrary guess, because this field isn't populated for all content. I just added this to see if there are any patterns we might want to look into.

There is still some content that is not linked to any organisation at all, but most content has some organisation information.

### DIFF
--- a/lib/screaming_frog.rb
+++ b/lib/screaming_frog.rb
@@ -1,0 +1,135 @@
+class ScreamingFrog
+  INPUT_HEADERS = [
+    "URL",
+    "Title",
+    "Character Length",
+    "Pixel Length",
+    "Description"
+  ].freeze
+
+  def initialize(input_stream, output_stream: STDOUT)
+    @input_csv = CSV.new(
+      input_stream
+    )
+
+    # The first 3 lines are garbage, the 4th is the real header
+    @input_csv = @input_csv.drop(4)
+
+    @output_csv = CSV.new(output_stream, headers: true)
+
+    @output_csv << INPUT_HEADERS + [
+      "Pageviews Last 6 Months",
+      "Document Type",
+      "Primary Publishing Organisation"
+    ]
+
+    @content_lookup = {}
+
+    fetch_all_the_things
+  end
+
+  def write
+    progressbar = ProgressBar.create
+    progressbar.total = 145377
+
+    input_csv.each do |row|
+      progressbar.increment
+
+      row = INPUT_HEADERS.zip(row).to_h
+      uri = URI(URI.encode(row["URL"]))
+      next if uri.host.present? && uri.host != 'www.gov.uk'
+
+      # Ignore withdrawn content for now
+      next if /\[Withdrawn\]/ =~ row["Title"]
+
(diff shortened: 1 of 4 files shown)


lib/screaming_frog.rb
```diff
@@ -0,0 +1,130 @@
...
+    # Use these as a fallback for when primary publishing organisation isn't set
+    organisation_links = Content::Link
+      .where(link_type: [:organisations, :worldwide_organisations])
+      .joins("JOIN content_items ON (links.target_content_id=content_items.content_id)")
+      .pluck(:source_content_id, :target_content_id, "content_items.title")
+
+    organisation_links.each do |source, target, title|
+      prolificness = organisations[target]
+
+      organisations_lookup[source] ||= []
```
dwhenry: can we just do this once outside the loop

dwhenry: or better yet, just sort the maximum rather than storing all of them and the doing a lookup later

lib/screaming_frog.rb
```diff
@@ -0,0 +1,130 @@
...
+
+    input_csv.each do |row|
+      progressbar.increment
+
+      row = INPUT_HEADERS.zip(row).to_h
+      uri = URI(URI.encode(row["URL"]))
+      next if uri.host.present? && uri.host != 'www.gov.uk'
+
+      # Ignore withdrawn content for now
+      next if /\[Withdrawn\]/ =~ row["Title"]
```
dwhenry: FYI: this is not currently a definitive way to ignore all withdrawn content. :(

lib/screaming_frog.rb
```diff
@@ -0,0 +1,130 @@
...
+
+      base_path = uri.path
+      binding.pry if base_path.nil?
```
dwhenry: assume this was a testing thing and we can now remove

MatMoore:
@pmanrubia I'm not sure! Do you think it's worth keeping?

pmanrubia:
Hi @MatMoore, I thought you were going to merge it because @dwhenry approved it. I would be happy with retrieving it from git history if needed.

MatMoore:
OK, I'm going to close the PR for now.

PR #309, opened 2017-08-29
Sort audits by multiple criteria

pmanrubia:
[Trello card](https://trello.com/c/iTfk3R38/468-2-sort-by-title)

Depends on https://github.com/alphagov/content-performance-manager/pull/307

### User needs

**As** a content auditor 
**I want to** see similar content together 
**So that** that I can deal with similarity efficiently.

**As** a content auditor 
**I want to** order content items in alphabetical order 
**So that** I can see similarly name content items, viewed as a batch and audit them more efficiently. 

### Changes

image


app/domain/audits/filter.rb
```diff
@@ -19,6 +21,15 @@ def audit_status=(value)
                       end
     end
 
+    def sort_by=(value)
+      if value.present?
```
andrewgarner: Perhaps we should have something a bit more robust here? The sort by value is only valid if it is a string containing an underscore followed by 'asc' or 'desc' aka `/\A[a-z]+_(asc|desc)\z/`. We've maybe overlooked that because at the moment we're concentrating on how this object is used by the controller upstream.

pmanrubia: Addressed in 499c54ad01373402047ea0225fd59545d5fc3f7d

pmanrubia: Addressed in 499c54ad01373402047ea0225fd59545d5fc3f7d

spec/domain/audits/filter_spec.rb
```diff
@@ -0,0 +1,21 @@
+module Audits
+  RSpec.describe Filter do
+    describe "#sort_by" do
+      it "does nothing is sort_by is not present" do
```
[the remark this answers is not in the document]
pmanrubia: Addressed in 499c54ad01373402047ea0225fd59545d5fc3f7d

spec/domain/audits/filter_spec.rb
```diff
@@ -0,0 +1,21 @@
+module Audits
+  RSpec.describe Filter do
+    describe "#sort_by" do
```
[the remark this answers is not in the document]
pmanrubia: Addressed in 499c54ad01373402047ea0225fd59545d5fc3f7d

pmanrubia:
@andrewgarner thank you for your review. I have addressed your comments.

PR #310, opened 2017-08-30
Show only auditable formats

pmanrubia:
[Trello card](https://trello.com/c/Affxvnkf/467-2-only-show-auditable-formats-in-the-audit-tool)

Depends on #307 

Not all formats are auditable by content auditors, so we need to exclude
them. We have decided to define a `Plan` and audit that determines what
subset of content items are auditable.

As of now, we are only excluding by format (document type), but in the
long run we plan to extend the plan definition with subscriptions and
individual content items.

### Changes

image

### DIFF
--- a/app/domain/audits/plan.rb
+++ b/app/domain/audits/plan.rb
@@ -1,0 +1,46 @@
+module Audits
+  class Plan
+    def self.document_type_ids
+      values.values
+    end
+
+    def self.document_types
+      values
+    end
+
+    def self.values
+      @values ||= {
+         "Answer" => :answer,
+         "Case Study > Case Study" => :case_study,
+         "Consultation > Closed Consultation" => :closed_consultation,
+         "Consultation > Consultation Outcome" => :consultation_outcome,
+         "Publication > Corporate Report" => :corporate_report,
+         "Publication > Correspondence" => :correspondence,
+         "Publication > Decision" => :decision,
+         "Detailed Guidance > Detailed Guide" => :detailed_guide,
+         "Document Collection > Collection" => :document_collection,
+         "Publication > Form" => :form,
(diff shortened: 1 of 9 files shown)


app/domain/content/query.rb
```diff
@@ -160,7 +163,12 @@ def set(instance_variable, value)
     end
 
     def builder(verify_presence:)
-      yield if verify_presence.present?
+      if verify_presence.is_a?(Array)
+        yield unless verify_presence.empty?
+      else
+        yield if verify_presence.present?
```
andrewgarner: `verify_presence` is present if it’s not blank. To be blank is to be false, empty, or a whitespace string. Therefore we don't need the condition:

```rb
def builder(verify_presence:)
  yield if verify_presence.present?
end
```

pmanrubia: Addressed in ca4c313e67ee6c298405c6c65eeea9c899e33b99

pmanrubia:
@andrewgarner: thank you for your review. I have addressed your comments.

app/domain/audits/plan.rb
```diff
@@ -0,0 +1,46 @@
...
+  class Plan
+    def self.document_type_ids
+      values.values
+    end
+
+    def self.document_types
+      values
+    end
+
+    def self.values
```
andrewgarner: It looks like the ordering of the values is significant? If so, I don't think that is a domain concern, but something we should do in the dropdown helper:

```rb
Audits::Plan.document_types.sort_by { |key, value| key.split(/\s>\s/) }
```

pmanrubia: Addressed in 75f0ed0

pmanrubia:
Rebased and force pushed in preparation for merging.

PR #315, opened 2017-09-01
Fix allocate/unallocate when nothing is selected

pmanrubia:
[Trello card](https://trello.com/c/nIYtldbE/491-fix-empty-allocations-issue)

We are getting server errors when the user is trying to allocate or
unallocated an empty selection of content items. To fix it we just need
to use a default (empty array) when nothing is selected.

### After

image

image

### DIFF
--- a/app/controllers/audits/allocations_controller.rb
+++ b/app/controllers/audits/allocations_controller.rb
@@ -28,7 +28,7 @@ module Audits
       if params[:select_all_pages]
         FindContent.all(build_filter).pluck(:content_id)
       else
-        params[:content_ids]
+        params.fetch(:content_ids, [])
       end
     end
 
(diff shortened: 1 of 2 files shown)


chao-xian:
:shipit:

app/controllers/audits/allocations_controller.rb
```diff
@@ -28,7 +28,7 @@ def content_ids
       if params[:select_all_pages]
         FindContent.all(build_filter).pluck(:content_id)
       else
-        params[:content_ids]
+        params.fetch(:content_ids) { [] }
```
andrewgarner: How about passing in a default instead of supplying a block?

```rb
params.fetch(:content_ids, [])
```

pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/315/commits/8c412454cbfb3a9b509dc5a23a543cc59afae50f

PR #318, opened 2017-09-04
Add support for title search in Audit Tool

pmanrubia:
[Trello card](https://trello.com/c/7JOmgvRl/495-3-add-search-to-audit-tool)

### User need

As a Content Auditor
I want to search content by text
So that I can audit and allocate content easily

### Changes

image

### DIFF
--- a/app/views/audits/common/_title.html.erb
+++ b/app/views/audits/common/_title.html.erb
@@ -1,0 +1,4 @@
+<div class="form-group">
+  <%= label_tag 'query', 'Search term:', class: '' %>
+  <%= text_field_tag 'query', params[:query], placeholder: 'Enter search term...', class: 'form-control' %>
+</div>
--- a/spec/features/audit/allocation/allocation_spec.rb
+++ b/spec/features/audit/allocation/allocation_spec.rb
@@ -17,7 +17,9 @@ RSpec.feature "Content Allocation", type: :feature do
     visit audits_allocations_path
 
     expect(page).to have_selector(".nav")
+    expect(page).to have_selector("li.active", text: "Assign content")
     expect(page).to have_selector("#sort_by")
+    expect(page).to have_selector("#query")
 
     expect(page).to have_content("content item 1")
     expect(page).to have_content("content item 2")
--- a/spec/features/audit/allocation/allocation_spec.rb
+++ b/spec/features/audit/allocation/allocation_spec.rb
@@ -1,0 +1,142 @@
+RSpec.feature "Content Allocation", type: :feature do
+  around(:each) do |example|
+    Feature.run_with_activated(:auditing_allocation) { example.run }
+  end
+
+  let!(:content_item) { create :content_item, title: "content item 1" }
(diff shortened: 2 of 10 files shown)


app/views/audits/common/_title.html.erb
```diff
@@ -0,0 +1,4 @@
+<div class="form-group">
+  <%= label_tag 'query', 'Search term:', class: '' %>
```
chao-xian: I would probably make this `Search in title...` as otherwise it might be misunderstood that it searches across all text (or maybe check with content designer).

spec/features/audit/lists/filter_spec.rb
```diff
@@ -153,6 +153,27 @@
     end
   end
 
+  context "filtering by title" do
+    scenario "the user enters a text in the search box and retrieves a filtered list" do
+      create :content_item, title: "some text"
+      create :content_item, title: "another text"
```
chao-xian: Is there any way to make the search case insensitive? When I search for `Driving` and `driving` I get different sets of results.

spec/features/audit/allocation/allocation_spec.rb
```diff
@@ -14,7 +14,7 @@
     create(:allocation, content_item: content_item, user: current_user)
     create(:allocation, content_item: another_content_item, user: another_user)
 
-    visit audits_path
+    visit audits_allocations_path
 
     expect(page).to have_selector(".nav")
```
chao-xian: If this test was wrong before we should probably then check that on the page. So maybe check that the page renders the `Assign to...` label for the dropdown, which doesn't appear for any of the other tabs.

app/views/audits/common/_title.html.erb
pmanrubia: I just reused the same label as in the CPM that was user researched. I would prefer to stick to existing patterns for the UI

spec/features/audit/lists/filter_spec.rb
pmanrubia: Address in e390f6c

app/views/audits/common/_title.html.erb
chao-xian: CPM is a slightly different context though. For example it doesn't ever display the summary for a piece of content, whereas the Audit Tool does. So for CPM `text` is always title but here we have two fields with text from a content item. I think it's worth asking a third party what they think. Or we can leave this until the next round of user research... It's only a label.

spec/features/audit/allocation/allocation_spec.rb
pmanrubia: Addressed in ad69271

app/views/audits/common/_title.html.erb
pmanrubia: Yeah, I would prefer to stick to what we have. Worth discussing with Joe, though.

pmanrubia:
Thank you for your review @chao-xian.
I have fixed up, rebased and force pushed in preparation for merging.

PR #321, opened 2017-09-07
Allow filtering by multiple organisations

andrewgarner:


andrewgarner:
[(5) Filter content for more than one organisation](https://trello.com/c/iLxEWCXQ/466-5-filter-content-for-more-than-one-organisation)

app/assets/javascripts/organisation-autocomplete.js
```diff
@@ -5,16 +5,118 @@ window.GOVUKAdmin.Modules = window.GOVUKAdmin.Modules || {};
...
+      function createAutocompleteTemplate() {
+        var $select = organisationSelects().first();
+        $select.removeAttr('multiple');
+        var $wrapper = wrapper($select);
+        var $template = $wrapper.clone();
+        $template.find('[selected]').removeAttr('selected');
+        return $template.prop('outerHTML');
+      }
+
+      function addOrganisationButton() {
```
pmanrubia: Why is the organisation button created via JS?

app/assets/javascripts/organisation-autocomplete.js
```diff
@@ -5,16 +5,118 @@ window.GOVUKAdmin.Modules = window.GOVUKAdmin.Modules || {};
...
-        showAllValues: true,
-        autoselect: false,
-        dropdownArrow: dropdownArrow
-      });
-    };
 
-    function dropdownArrow() {
-      return '<span class="caret"></span>';
-    }
+      var autocompleteTemplateHtml = createAutocompleteTemplate();
```
pmanrubia: I would rename `autocompleteTemplateHtml` to `$template`. Autocomplete is inferred from the module, and HTML is not adding much. I would also add $ as part of the name as this is a `JQuery` object.

app/assets/javascripts/organisation-autocomplete.js
```diff
@@ -5,16 +5,118 @@ window.GOVUKAdmin.Modules = window.GOVUKAdmin.Modules || {};
...
+            $(autocompleteTemplateHtml)
+              .insertAfter(lastSelectWrapper())
+              .find('option[value="' + $(option).attr('value') + '"]')
+              .attr('selected', 'selected')
+          }
+        });
+
+      enableAutocomplete();
+
+      function createAutocompleteTemplate() {
```
pmanrubia: Have you considered embedding the template within the HTML, hide it via CSS and clone it with `JQuery`?
I find this approach more readable and maintainable.

app/assets/javascripts/organisation-autocomplete.js
```diff
@@ -5,16 +5,118 @@ window.GOVUKAdmin.Modules = window.GOVUKAdmin.Modules || {};
...
+        return $(
+          '<button type="button" ' +
+          '        id="add-organisation" ' +
+          '        class="add-organisation js-add-organisation">' +
+          '  Add another organisation' +
+          '</button>'
+        ).click(addOrganisation);
+      }
+
+      function removeOrganisationButton() {
```
pmanrubia: As I commented above, I would prefer not to have HTML embedded in the JS if possible.

app/domain/content/query.rb
```diff
@@ -163,7 +163,7 @@ def set(instance_variable, value)
     end
 
     def builder(verify_presence:)
-      yield if verify_presence.present?
+      yield if [verify_presence].flatten.reject(&:blank?).present?
```
pmanrubia: Can you think of a way for not parsing the input on the domain layer? Ideally `flatten` and `reject` could be done in the controller?

app/assets/javascripts/organisation-autocomplete.js
alecgibson: I don't think we should have the `$` prefix - that denotes a jQuery element, not a raw HTML string.

app/assets/javascripts/organisation-autocomplete.js
alecgibson: I've done this sort of approach in the past, and it works pretty well. It sort of just moves the mess from JS to Ruby/CSS, but I don't mind either way.

app/assets/javascripts/organisation-autocomplete.js
alecgibson: Because we don't want it there when JS is disabled (progressive enhancement). We could alternatively have it present in the original HTML with `display: none` and unhide it in JS (similar to your suggestion for the template HTML).

app/assets/javascripts/organisation-autocomplete.js
```diff
@@ -5,16 +5,118 @@ window.GOVUKAdmin.Modules = window.GOVUKAdmin.Modules || {};
...
-        dropdownArrow: dropdownArrow
-      });
-    };
 
-    function dropdownArrow() {
-      return '<span class="caret"></span>';
-    }
+      var autocompleteTemplateHtml = createAutocompleteTemplate();
+
+      $element
```
alecgibson: If you're not moving this into the HTML, could we please break up this slightly monolithic nested jQuery chain? It would be nice to move it into a series of single-responsibility functions that describe what they're doing, eg `addAddOrganisationButton` (definitely want a better name than that); `addSelectElementsForSelectedOptions`

app/assets/javascripts/organisation-autocomplete.js
```diff
@@ -5,16 +5,118 @@ window.GOVUKAdmin.Modules = window.GOVUKAdmin.Modules || {};
...
+        .append(addOrganisationButton())
+        .find('[selected]')
+        .each(function (index, option) {
+          if (index > 0) {
+            $(autocompleteTemplateHtml)
+              .insertAfter(lastSelectWrapper())
+              .find('option[value="' + $(option).attr('value') + '"]')
```
alecgibson: There's got to be a nicer way of doing this. Probably using [`.val()`](https://api.jquery.com/val/).

app/assets/javascripts/organisation-autocomplete.js
```diff
@@ -5,16 +5,118 @@ window.GOVUKAdmin.Modules = window.GOVUKAdmin.Modules || {};
...
+          '  <span class="sr-only">Remove organisation</span>' +
+          '</button>'
+        ).click(removeOrganisation);
+      }
+
+      function enableAutocomplete() {
+        organisationSelects().each(function (index, select) {
+
+          // When no option is selected, the selectedIndex property will return
+          // -1. We have a blank option at index 0. Selecting the blank option
```
alecgibson: It might be worth mentioning this is specifically for `multiple`.

app/assets/javascripts/organisation-autocomplete.js
```diff
@@ -5,16 +5,118 @@ window.GOVUKAdmin.Modules = window.GOVUKAdmin.Modules || {};
...
+        var $organisationSelectWrapper = wrapper($removeButton);
+
+        if (organisationSelects().length === 1) {
+          clearAutocompleteAndSelect($organisationSelectWrapper);
+        } else {
+          $organisationSelectWrapper.remove();
+        }
+      }
+
+      function clearAutocompleteAndSelect($organisationSelectWrapper) {
```
alecgibson: Sorry, I forgot to remove this method when we got rid of the "X" on the first select, so I believe this is unused now? However, we could repurpose it to fire on `change` for the `input` fields, to address the bug where backspacing the `input` doesn't select the blank option?

app/assets/javascripts/organisation-autocomplete.js
pmanrubia: Ok. I would then build a `JQuery` object around it and use the `$`, and only get the `HTML` when needed. I have also scanned the code, and it feels that the variable [is only used once][1], and it basically [undoing this code][2]. 

[1]: https://github.com/alphagov/content-performance-manager/blob/f9e897b0ed61a658d0546cbc3d5861656803ba98/app/assets/javascripts/organisation-autocomplete.js#L64-L64
[2]: https://github.com/alphagov/content-performance-manager/blob/f9e897b0ed61a658d0546cbc3d5861656803ba98/app/assets/javascripts/organisation-autocomplete.js#L18-L18

app/assets/javascripts/organisation-autocomplete.js
pmanrubia: Unless you or @andrewgarner have strong opinions about using JS, I would prefer to use `HTML` via the `ERB` template.

app/assets/javascripts/organisation-autocomplete.js
andrewgarner: 👍

app/assets/javascripts/organisation-autocomplete.js
andrewgarner: 👍

app/assets/javascripts/organisation-autocomplete.js
alecgibson: But when we're creating a new HTML element, we don't want to accidentally copy around references to the template object. I think serialising it to a string is the safest way to do this.

app/assets/javascripts/organisation-autocomplete.js
alecgibson: Fine by me, but I think we should pick one method and do that. So if you want this in HTML, I think _everything_ should be in HTML (including the template for creating new `select` elements, as you suggested elsewhere).

app/assets/javascripts/organisation-autocomplete.js
pmanrubia: mmm not sure about it. Let's move the HTML out of the JS, use a template and discuss form there.

Hope this is ok with you @andrewgarner

app/assets/javascripts/organisation-autocomplete.js
andrewgarner: Personally I think the HTML we have in the JavaScript module is correct. On the server we render only the HTML we need for the feature to work and then progressively we add additional behaviour and markup through the JavaScript module when needed and supported.

This is the same approach taken by the [accessible-autocomplete](https://github.com/alphagov/accessible-autocomplete) module and would likely be the approach taken if this behaviour were extracted into a shared GOV.UK component or if we decided to add multiple selection to [accessible-autocomplete](https://github.com/alphagov/accessible-autocomplete).

alecgibson: Yeah, it's hard to say. Arguably the accessible autocomplete adds HTML through JavaScript, because that's how React and JSX is meant to be used, whereas it could be said that jQuery is intended mainly to manipulate existing DOM elements rather than create new ones.

If you wanted to be _really_ purist about it, we could share our HTML templates between JS and Ruby using [mustache](https://github.com/alphagov/shared_mustachehttps://github.com/alphagov/shared_mustache). That would let us define an HTML template once, but without having "useless" HTML floating around on the page, or declaring HTML markup in JS.

andrewgarner: We should decide what capabilities we gain or loose instead of thinking in terms of whether HTML should or should not belong in the JavaScript module.

As it stands now:

- The additional elements are only required when we progressively add the behaviour.
- We're not sharing the templates with the server so we're not duplicating markup.

pmanrubia: Let's not block the PR for this reason. Let's keep the HTML in the JS for now.

app/assets/javascripts/organisation-autocomplete.js
andrewgarner: I think we should fix that with https://trello.com/c/AvtSdooL/528-organisations-cant-be-de-selected

app/domain/content/query.rb
andrewgarner: This is a bit tricky. It's essentially defensive programming so that we're not relying on how `verify_presence` is provided to us.

I think this is necessary given the way the application is currently architected:

- `BaseController` constructs a `Filter` by passing whatever `params` values we have directly with no validation.
- `Filter` is a value object and sets no expectations about what a valid `Filter` attribute value is.
- `FindContent` sets no expectations about what a valid `Filter` is.
- `Content::Query` sets no expectations on the values provided to it's chain-able methods.

pmanrubia: `Filter` is currently validating the `sort_by` attribute, for example. A `ValueObject`, as any other object, can set a contract with their peers.  When we set any attribute on a Filter, we can/should transform it to something that is valid.

andrewgarner:
@pmanrubia @alecgibson comments addressed the best I can!

PR #324, opened 2017-09-08
Change Audit Status filter to radio buttons

alecgibson:
Radio buttons are more accessible, and generally more apparent than
dropdown menus. This change replaces the old `select` element with a set
of radio buttons for selecting the audit status.

This change also reorders the audit status options, and uses “Not
audited” as the default, both in the view, and in the controller.

### Before

audit-status-before

### After

audit-status-after

### DIFF
--- a/app/domain/audits/filter.rb
+++ b/app/domain/audits/filter.rb
@@ -15,11 +15,11 @@ module Audits
       :title
 
     def audit_status=(value)
-      @audit_status = if value.blank?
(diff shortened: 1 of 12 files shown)


alecgibson:
[Replace select box for audit status with a radio button](https://trello.com/c/ZInqU8Ct/512-replace-select-box-for-audit-status-with-a-radio-button)

PR #325, opened 2017-09-08
Present all organisation users in dropdowns

alecgibson:
When picking colleagues from dropdowns, we currently only show users
with content allocated to them. This can make for a confusing user
experience when trying to - for example - allocate content to users
who have never had content allocated to them, or when trying to view
content allocated to that user.

This change removes that restriction. The users in our database are only
the users who have logged into the application, which will also be
restricted by Signon permissions. This will naturally limit the number
of other users presented in the dropdowns, so it is assumed that we
do not need to restrict users in this way.

### DIFF
--- a/app/domain/audits/find_team_auditors.rb
+++ b/app/domain/audits/find_team_auditors.rb
@@ -12,7 +12,6 @@ module Audits
 
     def call
       User
-        .joins("INNER JOIN allocations ON allocations.uid = users.uid")
         .where(organisation_slug: user.organisation_slug)
         .distinct
         .to_a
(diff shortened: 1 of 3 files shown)


alecgibson:
[Show Info Message when there is no content to audit assigned to user](https://trello.com/c/xZgn0y9x/514-show-info-message-when-there-is-no-content-to-audit-assigned-to-user)

alecgibson:
[Display all users in an organisation in dropdowns](https://trello.com/c/YGJMEx8h/526-display-all-users-in-an-organisation-in-dropdowns)

PR #326, opened 2017-09-08
Display a message when no unaudited content is assigned

alecgibson:
In order to help users orientate themselves on the site, this change
adds a message that explicitly tells them if they are filtering by
“Not audited” content (the default), but no content has been assigned
that is not audited.

A message is shown regardless of whose allocations are being viewed,
but the message is only shown on the “Audit content” tab, and notably
_not_ the “Assign content” tab. This is because the “Assign content”
tab will have a default view that will probably default to showing all
unassigned, unaudited content, and having no content there will be a
rare case.

Also notably, the change does _not_ show a message for “Audited”
content either - so if a user has not yet audited any content, the user
will be presented with a blank table.

## Screenshots

me-no-content

someone-else-no-content

### DIFF
--- a/app/controllers/audits/audits_controller.rb
+++ b/app/controllers/audits/audits_controller.rb
@@ -4,10 +4,10 @@ module Audits
 
     def index
       respond_to do |format|
-        format.html { @content_items = FindContent.paged(build_filter) }
+        format.html { @content_items = FindContent.paged(filter) }
         format.csv do
           send_data(
-            Report.generate(build_filter, request.url),
+            Report.generate(filter, request.url),
             filename: "Transformation_audit_report_CSV_download.csv"
           )
         end
@@ -16,13 +16,13 @@ module Audits
 
     def show
       @content_item = Content::Item.find_by!(content_id: params.fetch(:content_item_content_id))
-      @next_content_item = FindNextItem.call(@content_item, build_filter)
+      @next_content_item = FindNextItem.call(@content_item, filter)
       @audit = Audit.find_or_initialize_by(content_item: @content_item)
     end
 
     def save
       @content_item = Content::Item.find_by!(content_id: params.fetch(:content_item_content_id))
-      @next_content_item = FindNextItem.call(@content_item, build_filter)
+      @next_content_item = FindNextItem.call(@content_item, filter)
       @audit = Audit.find_or_initialize_by(content_item: @content_item)
       @audit.user = current_user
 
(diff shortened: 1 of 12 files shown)


alecgibson:
[Show Info Message when there is no content to audit assigned to user](https://trello.com/c/xZgn0y9x/514-show-info-message-when-there-is-no-content-to-audit-assigned-to-user)

app/controllers/audits/audits_controller.rb
```diff
@@ -59,5 +63,26 @@ def audit_params
           )
         )
     end
+
```
pmanrubia: The controller seems to be doing too much. Could you think on ways to move this code out of here?

app/domain/audits/policies/allocated.rb
```diff
@@ -1,6 +1,8 @@
 module Audits
   module Policies
     class Allocated
+      NO_ONE = :no_one
```
pmanrubia: Not sure about this constant nor its relationship with the drop downs.

app/views/audits/audits/no_content_to_audit.html.erb
```diff
@@ -0,0 +1,18 @@
...
+
+<% content_for :header do %>
+  <%= render 'audits/common/navigation' %>
+<% end %>
+
+<div class="alert alert-info" role="alert">
+  <p>
+    <%= message -%>
+  </p>
+  <% unless params[:allocated_to].blank? || params[:allocated_to] == Audits::Policies::Allocated::NO_ONE.to_s %>
```
pmanrubia: I would extract this into a business rule. Some that can be reuse. It seems related to the 'Filter'

app/domain/audits/policies/allocated.rb
alecgibson: What do you mean by "not sure"? As in, you want me to find-replace? Or delete it?

alecgibson:
@pmanrubia can you please re-review?

app/controllers/audits/base_controller.rb
```diff
@@ -1,10 +1,10 @@
 module Audits
   class BaseController < ApplicationController
     layout "audits"
-    helper_method :filter_params, :primary_org_only?
+    helper_method :filter, :filter_params, :primary_org_only?
 
-    def build_filter(override = {})
-      options = {
+    def filter(override = {})
+      @filter_options ||= {
```
pmanrubia: Not sure about memoizing it. [Sometimes we modify it](https://github.com/alphagov/content-performance-manager/blob/5574c63f8051b70bb88917e1244f7fe7bd26db95/app/domain/audits/find_next_item.rb#L4-L6)

We might need to think in a more robust API for the filter.

alecgibson: Yeah fine - I think we should make this immutable in the future. I've undone that change.

PR #327, opened 2017-09-08
Switch default select option to 'Me'

chao-xian:
To help users orientation of the tool, and recognise the 'audit content'
tab as their area to do work assigned to them, we change 'audited by'
default option to 'Me' when first coming onto the tool.


chao-xian:
[Change 'audited by' default to 'me' when first coming onto the tool](https://trello.com/c/3YsLuzyS/522-change-audited-by-default-to-me-when-first-coming-onto-the-tool)

PR #329, opened 2017-09-11
Batch allocation of content items

pmanrubia:
[Trello card](https://trello.com/c/d4aMshlf/516-add-allocation-of-n-content-items-to-a-content-auditor)

### Overview

The previous mechanism to select content items to allocate had a few issues during user research. We have agreed to change the mechanism for multiple selection, so the user can enter the number of items that he wishes to allocate.

### Changes

image


app/views/audits/allocations/_toolbar.html.erb
```diff
@@ -1,31 +1,15 @@
...
-        <%= select_tag "allocate_to",
-                       allocation_options_for_select(params[:allocate_to]),
-                       class: "form-control",
-                       id: "allocate_to" %>
-        <%= filter_to_hidden_fields(:allocate_to) %>
-        <%= submit_tag "Go", class: "btn btn-default" %>
-      </div>
+    <div class="form-group form-inline pull-left">
+      <%= label_tag :allocate_to, "Assign", class: "allocate" %>
+      <%= text_field_tag :batch_size, "", class: "form-control" %>
```
andrewgarner: This needs a label to be accessible.

app/views/audits/allocations/_toolbar.html.erb
```diff
@@ -1,31 +1,15 @@
...
+      <%= label_tag :allocate_to, "items to", class: "items"  %>
```
andrewgarner: I don't believe this is an accessible label.

app/views/audits/allocations/index.html.erb
```diff
@@ -8,21 +8,21 @@
...
+          <td>Title</td>
+          <td>Page views (6mth)</td>
+        </tr>
+        </thead>
+        <tbody>
+        <%= render collection: content_items, partial: 'content_item' %>
+        </tbody>
+      </table>
+  <% end %>
+  <%= paginate content_items, :window => 2 %>
```
andrewgarner:
```rb
<%= paginate content_items, window: 2 %>
```

app/assets/javascripts/audits/batch-allocation.js
```diff
@@ -0,0 +1,38 @@
+window.GOVUKAdmin = window.GOVUKAdmin || {};
+window.GOVUKAdmin.Modules = window.GOVUKAdmin.Modules || {};
+
+(function (Modules) {
+  "use strict";
+
+  Modules.BatchAllocation = function () {
```
andrewgarner: This module is only about allocation when it is used in that context. Maybe we could call this `SelectBatches`?

app/controllers/audits/allocations_controller.rb
```diff
@@ -1,5 +1,7 @@
 module Audits
   class AllocationsController < BaseController
+    before_action :set_batch_value, only: %w(destroy create), if: -> { batch_size > content_ids.length }
```
andrewgarner: `content_ids.size` (alias of length) might read better here?

app/assets/javascripts/audits/batch-allocation.js
```diff
@@ -0,0 +1,38 @@
+window.GOVUKAdmin = window.GOVUKAdmin || {};
```
alecgibson: Given that we keep doing this, I think we should extract this and call it once in some sort of `initialize.js` or similar.

app/assets/javascripts/audits/batch-allocation.js
```diff
@@ -0,0 +1,38 @@
...
+window.GOVUKAdmin.Modules = window.GOVUKAdmin.Modules || {};
+
+(function (Modules) {
+  "use strict";
+
+  Modules.BatchAllocation = function () {
+    this.start = function ($root) {
+      this._initElements($root);
```
alecgibson: What do underscores in variable names mean in this context? I know they're traditionally meant to indicate `private`, but surely there's no need to do this with the GOV.UK module setup, because `.start` is the only function that's ever called externally?

app/assets/javascripts/audits/batch-allocation.js
```diff
@@ -0,0 +1,38 @@
...
+      this._bindEventHandlers();
+    };
+
+    this._initElements = function ($root) {
+      this.$batchSizeInput = $root.find("#batch_size");
+      this.$checkboxes = $root.find("input[type=checkbox]");
+    };
+
+    this._bindEventHandlers = function () {
+      this.$batchSizeInput.keyup($.proxy(this.onBatchSizeChange, this));
```
alecgibson: The fact that `this` is used three times on this line is a code smell to me. Any reason we don't just wrap this entire thing in `this.start`, and declare internal variables at the top of that function? Then we can stop saying `this` so much (which - in my experience - is generally a good thing).

app/assets/javascripts/audits/batch-allocation.js
```diff
@@ -0,0 +1,38 @@
...
+      this.$checkboxes.click($.proxy(this.onCheckboxClick, this));
+    };
+
+    this.onBatchSizeChange = function () {
+      this.$checkboxes
+        .prop('checked', false)
+        .slice(0, this.$batchSizeInput.val())
```
alecgibson: (NB: users can input things like `-10` and this will select everything except the last 10 elements - probably not a common occurrence, but that's the behaviour we'd get here)

app/assets/javascripts/audits/batch-allocation.js
pmanrubia: The module is a [constructor object][1], which means that it returns an object overtime it is instantiated. It is also a Singleton, as per the `Module` definition, with a public interface, by contract: the method `start`.

So, it is awkward to me to define everything int the start method, and not used the `this` object to define functions and instance variables. 

I am addressing your changes so we get this feature out, and ask in the frontend channel for best practices around Modules within GOV.UK.

[1]: https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/modules.js#L35-L37

app/assets/javascripts/audits/batch-allocation.js
pmanrubia: I find defining public/private methods very useful, as in the module the only public method, by contract is `start`, hence is the only one that we need to test. By following the convention os using `_` for private methods we make explicit that we can change them without breaking anything, and that we don't need to explicitly test them.

I am addressing your changes so we get this feature out, and ask in the frontend channel for current conventions around using `_` for private methods.

app/views/audits/allocations/_toolbar.html.erb
pmanrubia: I have created a [Trello][1] card to address this issue. I think it is better to split the card in the same way we have done for the multiple organisation selection.

[1]: https://trello.com/c/NVzzXbUL/529-make-batch-selection-accessible

app/views/audits/allocations/_toolbar.html.erb
pmanrubia: I have created a [Trello][1] card to address this issue. I think it is better to split the card in the same way we have done for the multiple organisation selection.

[1]: https://trello.com/c/NVzzXbUL/529-make-batch-selection-accessible

app/assets/javascripts/audits/batch-allocation.js
alecgibson: > in the module the only public method, by contract is `start`

Yes, that's what I meant - by definition everything that's not `start` is inferred to be `private`. I was just confused as to why you would mark some methods `private`, but not others (also it's kind of pointless to mark them, given that we have this convention).

app/assets/javascripts/audits/batch-allocation.js
alecgibson: Yes, happy to have discussion around this. My experience is that the JavaScript `this` keyword usually leads to more pain than it's worth, and should be avoided where possible. 

I'm happy to not define everything inside `this.start`; if you want to do that, then I'd prefer avoiding instance variables. Instead we should explicitly passing them around in as arguments, where `this.start` is like a "Main" method, which orchestrates all variable assignment and function calls.

app/assets/javascripts/application.js
```diff
@@ -14,8 +14,9 @@
 //= require bootstrap-tagsinput
 //= require cable
 //= require google-tag-manager
+//= require namespaces
 //= require inventory_toggle
-//= require select-all
 //= require selectize
 //= require accessible-autocomplete.min
 //= require organisation-autocomplete
+//= require audits/batch-selection.js
```
andrewgarner: We don't need an `audits` folder. This behaviour can be used anywhere.

app/views/audits/allocations/_toolbar.html.erb
andrewgarner: We know this is fundamentally not accessible, with the multiple organisation selection we made our best efforts to make it accessible and wrote a card to look into it further.

If we want to get this merged we should think about a simpler solution, with a conventional form structure and then write a card to explore a more sophisticated approach that remains accessible.

pmanrubia:
@andrewgarner I dismissed your review after having addressed all our comments. Please, let me know if you want to tackle any other thing and I will have a look at it.

PR #333, opened 2017-09-12
Update guide styling

alecgibson:
This change makes a couple of styling changes, primarily affecting the guidance, but also affecting the default page width of the audit tool:

- Default page to "full-width" (12 columns) if there is no sidebar present
- Increase guidance page width to 2/3 width
- Increase guidance copy font size to 19px

## Before

guidance-before

## After

guidance-after


alecgibson:
[Revise page width of in tool guidance](https://trello.com/c/HAO3rran/509-revise-page-width-of-in-tool-guidance)

PR #335, opened 2017-09-12
Add counter of content items

pmanrubia:
[Trello card](https://trello.com/c/6bCpfgfN/518-add-the-number-of-content-items-listed-in-the-allocation-tab-and-audit-content-tab)

Add the number of content items listed as a result of applying
the filter.

### Changes

image

image

### DIFF
--- a/app/views/audits/allocations/index.html.erb
+++ b/app/views/audits/allocations/index.html.erb
@@ -11,6 +11,8 @@
 <div class="allocations" data-module="batch-selection">
   <%= form_tag audits_allocations_path do %>
       <%= render 'toolbar' %>
+      <%= render 'audits/common/counter', locals: { content_items: content_items } %>
+
       <table class="table table-bordered table-hover">
         <thead>
         <tr class="table-header">
--- a/app/views/audits/audits/_content_items.html.erb
+++ b/app/views/audits/audits/_content_items.html.erb
@@ -1,3 +1,5 @@
+<%= render 'audits/common/counter', locals: { content_itemsitems: content_items } %>
+
 <table class="table table-bordered table-hover">
   <thead>
   <tr class="table-header">
--- a/app/views/audits/common/_counter.html.erb
+++ b/app/views/audits/common/_counter.html.erb
@@ -1,0 +1,3 @@
+<div class="content-items-counter">
+  <strong><%= format_number content_items.total_count %> <%='item'.pluralize(content_items.total_count) %></strong>
+</div>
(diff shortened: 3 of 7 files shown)


app/views/audits/allocations/index.html.erb
```diff
@@ -11,6 +11,8 @@
 <div class="allocations" data-module="batch-selection">
   <%= form_tag audits_allocations_path do %>
       <%= render 'toolbar' %>
+      <%= render 'audits/common/counter' %>
```
andrewgarner: We should provide the number to the partial as a local variable not rely on the partial being used in a context where `content_items` is present.

app/views/audits/common/_counter.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class="content-items-counter">
+  <strong><%= format_number content_items.total_count %> items</strong>
```
andrewgarner: If you do `<% 'item'.pluralize(content_items.total_count) %>` it will read "1 item" not "1 items".

app/views/audits/allocations/index.html.erb
pmanrubia: This is very true 😊

pmanrubia:
@andrewgarner thank you for your review. I have addressed your comments

app/views/audits/allocations/index.html.erb
```diff
@@ -11,6 +11,8 @@
 <div class="allocations" data-module="batch-selection">
   <%= form_tag audits_allocations_path do %>
       <%= render 'toolbar' %>
+      <%= render 'audits/common/counter', locals: { content_items: content_items } %>
```
andrewgarner: All the partial needs is the count. We can just do:

```rb
<%= render 'audits/common/counter', count: content_items.total_count %>
```

app/views/audits/audits/_content_items.html.erb
```diff
@@ -1,3 +1,5 @@
+<%= render 'audits/common/counter', locals: { content_itemsitems: content_items } %>
```
andrewgarner: This is still working because the partial is falling back to using the `content_items` helper method because we've misspelled the local variable. That would be revealed if we only gave the partial the data it needs.

pmanrubia: Sorry I tried to finish it in a rush before leaving. Will do it properly once I get back.

pmanrubia:
Superseded by #339

PR #340, opened 2017-09-13
Fix showing items belonging to "Me" as default

chao-xian:
When default filtering for allocation was switched from "Anyone" to
"Me", the default view of items should have displayed content items
for the current user.

### DIFF
--- a/app/controllers/audits/base_controller.rb
+++ b/app/controllers/audits/base_controller.rb
@@ -7,6 +7,7 @@ module Audits
       Filter.new(
         allocated_to: params[:allocated_to],
         audit_status: params[:audit_status],
+        current_user_id: current_user.uid,
         document_type: params[:document_type],
         organisations: params[:organisations],
         page: params[:page],
--- a/app/domain/audits/filter.rb
+++ b/app/domain/audits/filter.rb
@@ -4,6 +4,7 @@ module Audits
     attr_accessor :after,
       :allocated_to,
       :audit_status,
+      :current_user_id,
       :document_type,
       :organisations,
       :page,
@@ -19,7 +20,7 @@ module Audits
     end
 
     def audit_status
-      @audit_status || Audit::NON_AUDITED
+      @audit_status || Audit::ALLOCATED_TO
     end
 
     def sort_by=(value)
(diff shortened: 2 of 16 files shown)


chao-xian:

app/controllers/audits/base_controller.rb
```diff
@@ -7,6 +7,7 @@ def build_filter
       Filter.new(
         allocated_to: params[:allocated_to],
         audit_status: params[:audit_status],
+        current_user_id: current_user.uid,
```
andrewgarner: I feel like the concern of how to build the filter belongs in the controller. The controller can decide that if the `allocated_to` param is missing then we should apply a default.

Something like this:

```rb
Filter.new(allocated_to: params.fetch(:allocated_to, current_user.uid))
```

app/domain/audits/filter.rb
```diff
@@ -33,10 +34,13 @@ def sort_by=(value)
     end
 
     def allocated_policy
-      if allocated_to == 'no_one'
-        Policies::Unallocated
-      elsif allocated_to.blank?
+      if allocated_to.blank?
```
andrewgarner: If we move this concern to the controller we can negate this diff.

spec/features/audit/allocation/allocate_spec.rb
```diff
@@ -10,7 +10,7 @@
     second = create(:content_item, title: "content item 2")
     first = create(:content_item, title: "content item 3")
 
-    visit audits_allocations_path
+    visit audits_allocations_path(allocated_to: "anyone")
```
andrewgarner: I feel like we should be able to handle the absence of the `allocated_to` param and default to `anyone`.

spec/features/audit/allocation/allocate_spec.rb
```diff
@@ -38,7 +38,7 @@
   scenario 'Allocate all content within current page', :js do
     create_list(:content_item, 26)
 
-    visit audits_allocations_path
+    visit audits_allocations_path(allocated_to: "anyone")
```
andrewgarner: I think these scenarios should start with the default state which is to visit the allocations page without any filters. We should then describe what happens initially, then how the behaviour changes when we filter.

chao-xian:
Thanks for your feedback @andrewgarner. Closing this off as superseded by #342.

PR #341, opened 2017-09-13
Let users clear the first organisation filter with an “X” button

alecgibson:
We currently hide the “X” remove organisation button if there is only
one organisation filter. However, users cannot clear this selection, so
we should add this functionality back.

This change unhides the “X” button if there is only one organisation,
and fixes a bug that means clearing the input opens the options
without being able to dismiss them.


alecgibson:
[Organisations can't be de-selected](https://trello.com/c/AvtSdooL/528-organisations-cant-be-de-selected)

PR #342, opened 2017-09-14
[WIP] Change tab defaults

alecgibson:
The state of a user's filtering is currently shared between all three tabs of the Audit Tool, which is done by persisting the query string when changing tabs.

However, we now want to change the filters in each view, tailoring them to the task that is to be performed on that tab. As such, we want different default values when loading each tab, and we should stop sharing the filter state when switching tabs.

In order to do this, this change removes all notion of defaults and blank values from the `filter` class, and moves the defaults up into the controllers.

The sidebar filter values are also driven by the `filter` object, rather than the query parameters, which are not indicative of the current state of the filter (since they may be blank, but a default filter will apply).

### DIFF
--- a/app/controllers/audits/base_controller.rb
+++ b/app/controllers/audits/base_controller.rb
@@ -4,6 +4,16 @@ module Audits
     helper_method :filter, :filter_params, :primary_org_only?
 
     def filter(override = {})
+      options = default_filter
+        .merge(query_string_filter)
+        .merge(override)
+
+      Filter.new(options)
+    end
+
+  private
+
+    def query_string_filter
       options = {
         allocated_to: params[:allocated_to],
         audit_status: params[:audit_status],
@@ -13,12 +23,16 @@ module Audits
         primary_org_only: primary_org_only?,
         sort_by: params[:sort_by],
         title: params[:query],
-      }.merge(override)
+      }
 
-      Filter.new(options)
+      options.delete_if do |_, v|
+        v == "" || v.nil?
+      end
     end
 
-  private
+    def default_filter
+      {}
+    end
 
     def filter_params
       request
--- a/spec/features/audit/lists/list_content_items_spec.rb
+++ b/spec/features/audit/lists/list_content_items_spec.rb
@@ -6,20 +6,20 @@ RSpec.feature "List Content Items to Audit", type: :feature do
   end
 
   scenario "List Content Items to Audit" do
-    create(:content_item, title: "item1", six_months_page_views: 10_000, content_id: "content-id")
-    create(:content_item, title: "item2")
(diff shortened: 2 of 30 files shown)


alecgibson:
[Create a default filter for 'Audit content' and 'assign content' tabs](https://trello.com/c/Zg9BNa7N/533-create-a-default-filter-for-audit-content-and-assign-content-tabs)

alecgibson:

chao-xian:

spec/features/audit/lists/list_content_items_spec.rb
```diff
@@ -6,29 +6,29 @@
...
+    content_item1 = create(:content_item, six_months_page_views: 10_000, allocated_to: @current_user)
+    content_item2 = create(:content_item, allocated_to: @current_user)
 
     visit audits_path
 
-    expect(page).to have_content("item1")
+    expect(page).to have_content(content_item1.title)
     expect(page).to have_content("10,000")
-    expect(page).to have_link("item1", href: "/content_items/content-id/audit")
+    expect(page).to have_link(content_item1.title, href: "/content_items/#{content_item1.content_id}/audit")
```
pmanrubia: I'd rather not writing expectations based on fixture methods.

pmanrubia: First commit is doing lots of things. It is easier to review if this is broken down into smaller steps that describe a story.

alecgibson: It would've been impossible to make this an atomic commit without these test fixes - all the tests make now-broken assumptions about the state of the page when landing on it.

app/controllers/audits/base_controller.rb
```diff
@@ -13,12 +23,16 @@ def filter(override = {})
...
+      }
 
-      Filter.new(options)
+      options.delete_if do |_, v|
+        v == "" || v.nil?
+      end
     end
 
-  private
+    def default_filter
```
pmanrubia: Using a template method here is very handy. But default_filter is only used in the index action. Can you think of a way to make it more explicit?

spec/features/audit/lists/list_content_items_spec.rb
pmanrubia: Yes, tests should accompany code changes. But we could have created a commit per tab. Plus other commit for small refactorings? No need to change it now.  It is usually better to have small commits as the review will be done more safely.

alecgibson:
Superceded by https://github.com/alphagov/content-performance-manager/pull/345

alecgibson:
Superseded by https://github.com/alphagov/content-performance-manager/pull/345

PR #344, opened 2017-09-14
Remove automatic user setup from feature testing

alecgibson:
We currently "hide" the setup of our logged-in user in a `before`
block. This can obscure logic based on the current user, and hides
the assumptions made about them.

It also makes it harder to set the current user up with - say - a
particular organisation, which may be needed for testing.

This change removes the automatic setup, forcing each feature test to
declare its own logged-in user, along with any assumptions and extra
setup required (eg a particular name, which might be checked in banner
messages, or a particular organisation, which may be used for content
or selected from dropdowns, etc.).

### DIFF
--- a/spec/features/audit/allocation/filter_spec.rb
+++ b/spec/features/audit/allocation/filter_spec.rb
@@ -1,5 +1,9 @@
 RSpec.feature "Filter content by allocated content auditor", type: :feature do
-  let!(:current_user) { User.first }
+  let!(:me) do
+    create(
+      :user,
+    )
+  end
 
   scenario "List is unfiltered" do
     visit audits_path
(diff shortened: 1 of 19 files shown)


spec/features/audit/allocation/filter_spec.rb
```diff
@@ -1,5 +1,10 @@
 RSpec.feature "Filter content by allocated content auditor", type: :feature do
-  let!(:current_user) { User.first }
+  let!(:me) do
```
pmanrubia: 👍 to using `:me`

spec/features/taxonomy_generation/content_preview_modal_spec.rb
```diff
@@ -1,6 +1,13 @@
 require "rails_helper"
 
 RSpec.feature "Content Preview Modal", type: :feature, js: true do
+  let!(:me) do
```
pmanrubia: Love the names 👍 . Have you considered only adding names on the tests where are used?

pmanrubia:
Great work on being explicit on the tests!

spec/features/taxonomy_generation/content_preview_modal_spec.rb
alecgibson: Yeah agreed - will take them out where they're not used.

PR #345, opened 2017-09-15
Set different default filters on each tab

alecgibson:
## Overview

The primary aim of this Pull Request is to show a different "default" set of filters on each tab (discarding the previous filter when switching tabs). The motivation for this is that each tab has a particular purpose, and we should present the user with a view that lends itself to that purpose when they arrive.

## Defaults

The defaults are as follows:

### Audit content

- Audit status: unallocated
- Allocated to: me

### Assign content

- Audit status: unallocated
- Allocated to: no one
- Organisation: my organisation

### Audit status

- Audit status: all
- Allocated to: me
- Organisation: my organisation

## Other fixes

In order to reach this goal, some work had to be done to get our filters in good enough shape to cope with this change. These have all been isolated in their own commits.

### Move defaults out of the filter class

The `Filter` class has an understanding of defaults. In order to have different defaults on a per-view basis, the Filter class needs to be completely unaware of any default values, and should not assign special meaning to "blank" values.

### Render allocation hidden fields directly from the filter

When allocating content, the current filter values are submitted with the form as a set of hidden `input` fields. These were previously generated from the query string, but the query string will not always hold the filter status (ie when changing tabs). We now update these hidden filters directly from the filter. In order to do this, the filter has been made to extend a `Struct` in order to have use of the `to_h` method.

### Sidebar filter values from the filter object

Similarly to the hidden inputs, the sidebar used to pull the current filter values from the query string, but this is unreliable. Instead, they are changed to read from the filter object directly.

### Replace `Struct`

We may want to think about using something other than a `Struct` for the `Filter`, since `Struct` has several ways of accessing attributes (`[:myattribute]`, `.my_attribute`, `.to_h[:my_attribute]`). As such, it's possible to have an inconsistent API with a `Struct`, which is not ideal.

However, it is considered "good enough" for this change, and we should iterate from here.

### Add multiple-organisation filter tests for allocation

As part of the refactoring, this change addresses a bug where we could not bulk-allocate content when filtering by multiple organisations. However, this was previously untested, and adding a test is considered out-of-scope for this change.


alecgibson:
[Create a default filter for 'Audit content' and 'assign content' tabs](https://trello.com/c/Zg9BNa7N/533-create-a-default-filter-for-audit-content-and-assign-content-tabs)

alecgibson:

alecgibson:
[When allocating content, the organisation is not persisted in the lelft side filter.](https://trello.com/c/bnNIt8bR/543-when-allocating-content-the-organisation-is-not-persisted-in-the-lelft-side-filter)

pmanrubia:
Thank you for the time you have put into rewriting this PR in smaller commits. It is very easy to follow. 

I would like to work a bit more on:

1. How defaults are being set
2. The Filter object role within the view and domain layer
3. Explicitness of the tests

Having said this, let's merge it. Good job! 👍

PR #349, opened 2017-09-18
Limit Sidekiq retries and only log final failure

alecgibson:
We previously used Sidekiq's default retry behaviour, which is an
exponential backoff strategy limited to 25 retries, which will span
approximately three weeks.

Since our jobs rerun daily, we do not need to keep retrying beyond one
day at the most.

This change therefore limits the number of retries to 14, which is
approximately 12 hours according to the formula documented in [the
Sidekiq documentation](https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry):

```
(retry_count ** 4) + 15 + (rand(30) * (retry_count + 1))
```

Further, our jobs are highly dependent on external parties' APIs. We do
not need to log errors that are beyond our control (for example,
timeouts).

This being the case, this change also updates the Airbrake configuration
to ignore all errors whose `retry_count` is below the maximum retries.

In order to achieve these changes, the workers must now use the
`Sidekiq::Worker` mixin, rather than extending `ActiveJob::Base`,
because `ActiveJob::Base` does not allow configuration of
`sidekiq` options, as noted in [the documentation](https://github.com/mperham/sidekiq/wiki/Active-Job).

### DIFF
--- a/app/domain/content/importers/all_content_items.rb
+++ b/app/domain/content/importers/all_content_items.rb
@@ -10,7 +10,7 @@ module Content
       content_items = content_items_service.fetch_all_with_default_locale_only
 
       content_items.each do |content_item|
-        ImportItemJob.perform_later(content_item[:content_id], content_item[:locale])
+        ImportItemJob.perform_async(content_item[:content_id], content_item[:locale])
       end
     end
   end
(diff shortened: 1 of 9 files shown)


alecgibson:
[Only log an error in Errbit/Sentry when the Sidekiq job has no retries left](https://trello.com/c/5goqTLpz/550-only-log-an-error-in-errbit-sentry-when-the-sidekiq-job-has-no-retries-left)

PR #350, opened 2017-09-19
Upgrade to sentry

alecgibson:
We currently use Errbit for error reporting, but we are [migrating all applications to use Sentry instead](https://trello.com/b/aZ0OD1qJ/sentry-migration).

This change removes Airbrake, which was used to report errors to Errbit, and adds the `govuk_app_config` gem, which sets up the application to use Raven for reporting to Sentry.

A test error was triggered, and the output can be seen [here](https://sentry.io/govuk/app-content-performance-manager/issues/352306091/).


alecgibson:
[Move CPM from Errbit to Sentry](https://trello.com/c/wZhqA82N/532-move-cpm-from-errbit-to-sentry)

PR #352, opened 2017-09-20
Content Audit Tool is now in BETA

andrewgarner:


andrewgarner:
[Add feedback link to Beta banner](https://trello.com/c/GZAcO0DK/542-add-feedback-link-to-beta-banner)

alecgibson:
Given that this is one huge `if` block, wouldn't it better to just have two separate Alpha and Beta templates? Things would be a lot more readable that way.

app/views/layouts/audits.html.erb
```diff
@@ -17,7 +17,7 @@
 <% end %>
 
 <% content_for :content do %>
-  <%= render 'application/phase_banner' %>
+  <%= render 'application/phase_banner', phase: :beta %>
```
pmanrubia: Have you thought in splitting `phase_banner` into two different files?. We are passing a variable and then [checking it inside the partial with no big benefit](https://github.com/alphagov/content-performance-manager/pull/352/files#diff-211f3c6952675e67d0f1f1bdf4f9c8dcR7).

In the layout, we know what phase we are in, so we should be able to pick up the right one.

andrewgarner:
@pmanrubia @alecgibson – changes made.

PR #353, opened 2017-09-21
Log certain Sidekiq errors only on their final retry

alecgibson:
Some of our Sidekiq errors arise from interactions with 3rd party APIs,
and failures beyond our control are to be expected from time to time.
We do not want to be notified every time we get one of these errors;
we just want to be notified the final time the job fails.

This change updates our `ApplicationJob` to perform its tasks inside a
block that will be rescued when certain errors raise. It will then
re-raise a `RetryableError`, which will be ignored by Sentry.

When the retries on a job are exhausted, the error will finally be sent
to Sentry.

Raven is Sentry's API offering. Documentation on the Ruby implementation
can be found [here][raven].

We thought about solving this using Raven's `should_capture` method,
which would enable us to filter errors base on arbitrary conditions,
but that would have require some error parsing. The approach taken here
is more explicit about the expected errors, and requires fewer moving
parts.

[raven]: https://docs.sentry.io/clients/ruby/

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,9 +1,36 @@
+require 'gds_api/exceptions'
+require 'google/apis/analyticsreporting_v4'
+
 class ApplicationJob
   include Sidekiq::Worker
+
+  # List of errors to retry without alerting Sentry until their final retry
+  ERRORS_TO_RETRY_WITHOUT_ALERTS = [
+    Google::Apis::ServerError,
+    Google::Apis::TransmissionError,
+    GdsApi::HTTPBadGateway,
+    GdsApi::TimedOutException,
+  ].freeze
 
   # Retry for ~12 hours with exponential backoff, according to the default Sidekiq formula:
   # https://github.com/mperham/sidekiq/wiki/Error-Handling#automatic-job-retry
   MAX_RETRIES = 14
 
   sidekiq_options retry: MAX_RETRIES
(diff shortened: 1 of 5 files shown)


alecgibson:
[Only log an error in Errbit/Sentry when the Sidekiq job has no retries left](https://trello.com/c/5goqTLpz/550-only-log-an-error-in-errbit-sentry-when-the-sidekiq-job-has-no-retries-left)

app/jobs/application_job.rb
```diff
@@ -1,9 +1,37 @@
+require 'gds_api/exceptions'
+require 'google/apis/analyticsreporting_v4'
+
 class ApplicationJob
   include Sidekiq::Worker
 
+  # List of errors to retry without alerting Sentry until their final retry
+  ERRORS_TO_RETRY_WITHOUT_ALERTS = [
+    Google::Apis::RateLimitError,
```
pmanrubia: I would not include here `Google::Apis::RateLimitError`. It could have implications in our GA quota.

config/raven.rb
```diff
@@ -0,0 +1,3 @@
+Raven.configure do |config|
```
pmanrubia: Could you please write a brief description in the commit notes about what `raven` is, and where we can find documentation about it?

alecgibson:
@pmanrubia done.

PR #356, opened 2017-09-21
Update labelling

andrewgarner:


andrewgarner:
[(3) Make changes to labelling of audit tool (as recommended in ticket #506)](https://trello.com/c/FiWkJSmM/507-3-make-changes-to-labelling-of-audit-tool-as-recommended-in-ticket-506)

app/controllers/audits/allocations_controller.rb
```diff
@@ -17,13 +17,13 @@ def index
     def create
       allocation = AllocateContent.call(user_uid: user_uid, content_ids: content_ids)
 
-      redirect_to audits_allocations_url(redirect_params), notice: allocation.message
+      redirect_to audits_allocations_url(redirect_params), notice: allocation.notice, alert: allocation.alert
```
pmanrubia: This seems a bit odd to me. Using `notice:` and `alert:` int he response. Any thoughts on how to improve it?

```ruby
notice: allocation.notice, alert: allocation.alert
```

alecgibson: I did this originally. `alert` and `notice` are Railsy things. We could alternatively send back some sort of symbol that says if the banner is `:success` or `:danger` or similar, but we'd still need to send back a second argument.

andrewgarner: I've attempted to address this with b4cdaa0497fd3436ce9f199b58d20c085ee1a977.

I think this is a better approach because the messaging is concern of the view. All we need the controller to do is make the data available.

Thoughts?

andrewgarner: I've attempted to address this with b4cdaa0.

I think this is a better approach because the messaging is concern of the view. All we need the controller to do is make the data available.

Thoughts?

pmanrubia: I would encapsulate the result of the allocation within the Operation/Interaction.
If we have an operation that is called `AllocateContent`, I would expect to be able to do something like this:

```ruby
# DRY-up needed
allocation = AllocateContent.call
if allocation.success?
  flash[:notice] = allocation.message
else
  flash[:alert] = allocation.message
end
```

I would not extract another class to model the result of the allocation. I would inline it within the operation itself. It is ok for me to wrap the message in the `allocation`, in the same way that Exceptions carry a human-readable-message with them.

If we embed in a view...

```ruby
if flash[:count] 
  if flash[:count].zero? 
```

...then other views would need to carry out with them a similar logic; I would say that it fits a bit better in the `AllocateContent` class.

Testing would also be better.

Any other ideas?

andrewgarner: I've followed your approach in 117929c579fcdf1f3d7ec8b0078bd3c73d9082b5

app/models/audits/report_row.rb
```diff
@@ -29,7 +29,6 @@ def precompute
         other_organisations,
         content_type,
         last_major_update,
-        whitehall_url,
```
pmanrubia: `whitehall_url` is currently being displayed on the report (CSV). Could you please double check with Ben that this is no longer needed? It seems important to be able to do the changes quickly, but I might be missing something.

PR #360, opened 2017-09-21
Track allocation batch size

alecgibson:
When allocating content, we want to be able to track the number of
content items users are assigning.

This change adds a tracking module that will submit this information to
Google Tag Manager when the form is submitted.


alecgibson:
[(2) push allocation 'batch_size' to the data layer when content is assigned](https://trello.com/c/oJ4Ai6VE/505-2-push-allocation-batchsize-to-the-data-layer-when-content-is-assigned)

PR #361, opened 2017-09-22
Add new Sidekiq processes to the Procfile

alecgibson:
This change updates the Procfile to define the two new Sidekiq processes
we need to run.

The old process definition will be removed once Puppet is updated and
deployed.


alecgibson:
[Use different concurrency sidekiq policy for GA and for PublishingAPI](https://trello.com/c/0TgXtx7w/549-use-different-concurrency-sidekiq-policy-for-ga-and-for-publishingapi)

PR #362, opened 2017-09-22
Improve multiple organisation test waiting

alecgibson:
The tests involving the JavaScript-enhanced multiple organisation
select fail intermittently. In order to try to address this issue, this
change:

- increases the amount of time Capybara will wait for an element to 5
  seconds
- adds more explicit selector assertions, which should ensure that the
  page is in the expected state before continuing with the rest of the
  test

### DIFF
--- a/spec/features/audit/lists/filter_spec.rb
+++ b/spec/features/audit/lists/filter_spec.rb
@@ -155,6 +155,10 @@ RSpec.feature "Filter Content Items to Audit", type: :feature do
       end
 
       context "filtering by organisation" do
+        around(:each) do |example|
+          Capybara.using_wait_time(5) { example.run }
+        end
+
         scenario "organisations are in alphabetical order" do
           within("#organisations") do
             options = page.all("option")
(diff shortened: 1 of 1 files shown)


alecgibson:
[Fix intermittent test failure](https://trello.com/c/pHaOGHl7/560-fix-intermittent-test-failure)

spec/features/audit/lists/filter_spec.rb
```diff
@@ -164,34 +164,47 @@
         end
 
         scenario "using autocomplete", js: true do
-          expect(page.current_url).not_to include("organisations%5B%5D=#{hmrc.content_id}")
```
pmanrubia: What about?

```ruby
context "filtering by organisation" do
  around(:each) do |example|
    Capybara.using_wait_time 5 do
      example.run
    end
  end
  #
end
```

Or, I use the scenario `metadata` data and configure it globally in `rails_helper.rb`?

alecgibson: @pmanrubia done.

PR #363, opened 2017-09-27
Handle `nil` rows in Google Analytics responses

alecgibson:
When we fetch pageview data from Google Analytics, it returns a response
object. If none of the queried base paths have any pageviews, then
the response object's rows will be `nil`, rather than an empty array.

This case may arise if we query a batch of uncommonly-accessed documents,
such as a [batch of announcments][sentry].

In order to guard against `nil` errors, this change checks that the
`rows` field exists, and - if it isn't - will provide an empty array.

[sentry]: https://sentry.io/govuk/app-content-performance-manager/issues/353541553/events/7483935478/

### DIFF
--- a/app/services/google_analytics/responses/page_views_response.rb
+++ b/app/services/google_analytics/responses/page_views_response.rb
@@ -8,7 +8,10 @@ module GoogleAnalytics
 
       def parse(response)
         report = response.reports.first
-        report.data.rows.map do |row|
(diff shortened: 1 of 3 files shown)


alecgibson:
[Handle `nil` rows in Google Analytics job](https://trello.com/c/1rsMpaky/566-handle-nil-rows-in-google-analytics-job)

PR #364, opened 2017-09-27
Only process Google Analytics for truthy base paths

alecgibson:
When requesting the Google Analytics data for a `nil` base path, we
surprisingly get results.

We are currently creating a filter in `page_views_request` by ignoring
all `nil` base paths, but in the corner case where **all** base paths
are `nil`, this results in an empty filter (ie **all** results are
returned!).

Processing all results then also results in the return of every single
path processed by GA, which includes paths who have no content item.
This in itself results in attempting to fetch a non-existent content
item, which is `nil`, and causes a `NoMethodError` when trying to call
`update!` on it, as seen in [this error report][sentry].

This change makes a couple of remedial actions. First of all, we no
longer filter out blank base paths when creating a filter. The behaviour
of creating an empty filter is confusing, so instead we should fail
loudly instead of swallowing the error at this low level.

Instead, the guard for `nil` base paths is moved up to the service layer
in `GoogleAnalyticsService`, which should be responsible for cleansing
the input it receives.

A second guard is also added at the layer above the service, in the
`AllGoogleAnalyticsMetrics` importer.

Note that here we **do not** check for the presence of a content item
when trying to retrieve it by base path. We expect to **never** try
to retrieve a content item with a non-existent base path, so we should
not swallow that error, and should indeed continue to fail loudly as we
have done in this case.

[sentry]: https://sentry.io/govuk/app-content-performance-manager/issues/353542557/events/7483853175/

### DIFF
--- a/app/domain/content/importers/all_google_analytics_metrics.rb
+++ b/app/domain/content/importers/all_google_analytics_metrics.rb
@@ -8,7 +8,9 @@ module Content
 
     def run
       Content::Item.find_in_batches(batch_size: batch_size) do |content_items|
-        base_paths = content_items.map(&:base_path)
+        base_paths = content_items
+          .map(&:base_path)
+          .select(&:present?)
         ImportPageviewsJob.perform_async(base_paths)
       end
     end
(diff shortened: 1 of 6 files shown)


alecgibson:
[We're getting `nil` errors when trying to fetch base paths that are all `nil`](https://trello.com/c/JwTjYJmt/567-were-getting-nil-errors-when-trying-to-fetch-base-paths-that-are-all-nil)

PR #368, opened 2017-10-05
Track user's organisation in the Audit Tool

alecgibson:
In order to better understand the data we are receiving from Google Tag
Manager (GTM), we would like to know the organisation of the user. This
will help us to filter out "noise" from users at Government Digital
Service (GDS). It will also help us to see how organisations work when
using the tool.

In order to track the user's organisation, this change introduces a
`<script>` tag on the Audit Tool only (ie not the Content Performance
Manager). This `<script>` tag is populated with JSON by a Rails template,
which fetches information from a helper method, where we have access to
the user's organisation from Signon.

Since we are in full control of the JSON, and it is all rendered server-
side, this method is considered safe.
inject the JSON onto the page in this way.

### DIFF
--- a/app/helpers/google_tag_manager_helper.rb
+++ b/app/helpers/google_tag_manager_helper.rb
@@ -1,0 +1,7 @@
+module GoogleTagManagerHelper
+  def google_tag_manager
+    @google_tag_manager ||= GoogleAnalytics::GoogleTagManager.new(
+      current_user,
+    )
+  end
+end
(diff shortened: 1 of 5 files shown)


alecgibson:
[GTM tracking - Push the 'organisation' of an auditor to the data layer](https://trello.com/c/wnzrQgix/568-gtm-tracking-push-the-organisation-of-an-auditor-to-the-data-layer)

PR #369, opened 2017-10-05
Allow clearing of default organisation

alecgibson:
This change addresses a bug we have:

1. Visit the "Allocation" page
2. The organisation filters should default to your own organisation
3. Clear the organisation filter
4. Apply filters
5. The page is still filtered by your own organisation

This is because we have no way of determining the difference between a user who has just landed on the page "cold" (ie with no filters set), or a user who has actively deselected all the organisations (or selected the blank organisation).

In order to fix this, this change moves "blankable" filters (ie filters who can override defaults when explicitly `blank` eg `[]` or `false`) into their own method, and adds a `hidden` `<input>` field to the organisations multi-select. See individual commits for further details.

### DIFF
--- a/app/controllers/audits/allocations_controller.rb
+++ b/app/controllers/audits/allocations_controller.rb
@@ -9,6 +9,7 @@ module Audits
         allocated_to: :no_one,
         audit_status: Audits::Audit::NON_AUDITED,
         organisations: [current_user.organisation_content_id],
+        primary_org_only: true,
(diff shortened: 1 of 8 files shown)


alecgibson:
['Clear an organisation' in allocation is not working.](https://trello.com/c/6b7n4cJE/564-clear-an-organisation-in-allocation-is-not-working)

PR #370, opened 2017-10-06
Add test for allocating content from multiple organisations

alecgibson:
We recently [fixed a bug][change] where allocating to organisations
wasn't working because the organisations are an array.

This change adds test coverage for that case, which we didn't previously
have.

[change]: https://github.com/alphagov/content-performance-manager/commit/311e4c805cf7992b277fb06d07a8c9946474e3ef#diff-a1d9a8be563b9698e6c917b6b4de816c


alecgibson:
[Add test for allocating to multiple organisations](https://trello.com/c/CXPR0K4x/586-add-test-for-allocating-to-multiple-organisations)

PR #372, opened 2017-10-10
Ensure filter defaults are applied on first interaction

andrewgarner:
When you first visit one of the audit tabs, the querystring does not contain any filter state. This means default filters are only applied if you interact with an element on the page that has URLs appended with the filter defaults.

This fixes that so that the defaults are applied correctly.

### DIFF
--- a/app/controllers/audits/base_controller.rb
+++ b/app/controllers/audits/base_controller.rb
@@ -44,9 +44,9 @@ module Audits
     end
 
     def filter_params
-      request
-        .query_parameters
-        .deep_symbolize_keys
+      Audits::SerializeFilterToQueryParameters
+        .new(filter)
+        .call
     end
 
     def primary_org_only?
--- a/app/services/audits/serialize_filter_to_query_parameters.rb
+++ b/app/services/audits/serialize_filter_to_query_parameters.rb
@@ -1,0 +1,19 @@
+module Audits
+  class SerializeFilterToQueryParameters
+    def initialize(filter)
+      @filter = filter
+    end
+
+    def call
+      {
+        allocated_to: @filter.allocated_to,
+        audit_status: @filter.audit_status,
+        document_type: @filter.document_type,
+        organisations: @filter.organisations,
+        primary: @filter.primary_org_only,
+        query: @filter.title,
+        sort_by: @filter.sort_by,
+      }
+    end
+  end
+end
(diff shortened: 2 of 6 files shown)


andrewgarner:
[Bug: Filter context missing from audit.](https://trello.com/c/NU10JeNq/591-bug-filter-context-missing-from-audit)

app/domain/audits/serialize_filter_to_query_parameters.rb
```diff
@@ -0,0 +1,21 @@
+module Audits
+  class SerializeFilterToQueryParameters
```
pmanrubia: Would you say that `Query` parameters belong to the domain layer.
It is a good idea to extract into a serializer the code to create a filter, but I would expect this code to live out of the domain layer. When we talk about query parameters in the domain layer it is likely to think that they are database related?

Any thoughts?

app/domain/audits/serialize_filter_to_query_parameters.rb
```diff
@@ -0,0 +1,21 @@
...
+    attr_accessor :filter
```
alecgibson: Why have we exposed this accessor?

app/domain/audits/serialize_filter_to_query_parameters.rb
andrewgarner: Agreed. I've moved into the services directory.

spec/domain/audits/serialize_filter_to_query_parameters_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  RSpec.describe SerializeFilterToQueryParameters do
+    let(:allocated_to) { double }
+    let(:audit_status) { double }
+    let(:document_type) { double }
+    let(:organisations) { double }
+    let(:primary_org_only) { double }
+    let(:sort_by) { double }
+    let(:title) { double }
+
+    let(:filter) {
```
alecgibson: This is a matter of personal preference, but this test basically checks that we take a hash and reproduce exactly the same hash, which doesn't seem hugely valuable. I know it adds some coupling, but it might be nice to actually instantiate a `Filter` object and pass that in. I think the coupling is okay, because this class' sole purpose in life is to operate on `Filter` objects (as pointed out by its name).

app/domain/audits/serialize_filter_to_query_parameters.rb
andrewgarner: No deliberate reason. I'll remove.

spec/domain/audits/serialize_filter_to_query_parameters_spec.rb
andrewgarner: I used a double in the spirit of duck typing. We're testing that when given an object with a suitable interface we return a suitable hash. We test the integration elsewhere.

PR #373, opened 2017-10-11
Fix persistence of sort by filter

andrewgarner:
The sort by filter was not being persisted due to a regression within the Filter class.
Our Filter object is a struct and so we need to use the accessor methods not instance variables.


andrewgarner:
[Bug: "Sort by" filter gets reset](https://trello.com/c/YwJjQDCG/592-bug-sort-by-filter-gets-reset)

PR #375, opened 2017-10-19
Reorder questions

alecgibson:
Users want to decide if content is a duplicate before deciding if it should be removed. This change moves the duplicate question before the question asking if the content should be removed.


alecgibson:
[(2) Users want to decide whether an item is a duplicate before deciding if it needs to be removed -- (also) Include the word 'duplicate' in the question about similar content](https://trello.com/c/DBzDbQRH/605-2-users-want-to-decide-whether-an-item-is-a-duplicate-before-deciding-if-it-needs-to-be-removed-also-include-the-word-duplicate)

PR #376, opened 2017-10-19
Show 100 items per page

alecgibson:
This change shows 100 items per page on both the audit and allocation pages. This is to address a user need to be able to quickly find the content they are looking for. Showing more items should reduce the amount of paging they have to do, and make it easier to find their content.


alecgibson:
[(2) Increase items on a page to 100 items](https://trello.com/c/W9EhP4pa/613-2-increase-items-on-a-page-to-100-items)

PR #377, opened 2017-10-19
Look up Users' organisation titles from Content Items

alecgibson:
We currently try to infer the title of a user's organisation by [`titleize`ing][1] their organisation slug. However, this doesn't work in all cases. For example, HMRC has the slug `/hm-revenue-customs`, which titleizes to "Hm Revenue Customs", when the actual title of the organisation is "HM Revenue & Customs" (note the incorrect capitalization of "HM" and the missing ampersand).

This change removes the titleization, and instead explicitly looks up the organisation title from the organisation's Content Item. This is done for both auditors and assignees.

[1]: http://api.rubyonrails.org/v5.1/classes/ActiveSupport/Inflector.html#method-i-titleize

### DIFF
--- a/app/decorators/audits/audit_decorator.rb
+++ b/app/decorators/audits/audit_decorator.rb
@@ -6,7 +6,6 @@ module Audits
 
     decorates_association :responses
     decorates_association :content_item
-    decorates_association :user
 
     def incomplete?
       object.new_record?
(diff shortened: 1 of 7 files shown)


alecgibson:
[(3) The item page view has a typo in the summary column: 'Hm Revenue and Customs'](https://trello.com/c/jgY02Tin/612-3-the-item-page-view-has-a-typo-in-the-summary-column-hm-revenue-and-customs)

PR #378, opened 2017-10-19
Add Navigation to the Guidance

alecgibson:
This change adds a list of anchor links that point to all the `h2` elements in the guidance.

guidance-nav


alecgibson:
[(2) Users need an easy navigation through the guidance page](https://trello.com/c/ugEWsAxT/604-2-users-need-an-easy-navigation-through-the-guidance-page)

app/controllers/audits/guidances_controller.rb
```diff
@@ -1,7 +1,21 @@
...
   class GuidancesController < BaseController
     def show
-      @content = Support.get
+      @content = Govspeak::Document.new(Support.get).to_html
+      @headings = h2_from_html(@content)
+    end
+
+  private
+
+    def h2_from_html(html)
```
pmanrubia: I would prefer not having HMLT related code in a controller. 
If the headings are important, we could extract them within:

```ruby
@content = Support.get
heading = @content.headings
```
Although the best option to me would be to use a custom template renderer. 
Like: `guidance.html.markdown`

alecgibson: @pmanrubia I'm not sure I understand the "custom template renderer" part of your comment, but I've moved the HTML logic into the (renamed) `Guidance` class.

alecgibson:
@pmanrubia as requested, I've made this a little bit more magical. The application should now handle arbitrary `.html.md` files, giving them all the appropriate sidebar, and without the need to define a wrapping `.html.erb` partial for them. The flip-side is that there's a bit more machinery around it, and it's a little bit more obscure.

PR #387, opened 2017-10-27
Add multiple users support for prototyping in Heroku

pmanrubia:
Adds multiple users support in Heroku deployments

### Background

The support for login is provided by [GDS-SSO][1] which is a gem that provides everything needed to integrate with [Signon][2]

In #384, we created a `rake` task to:

1. Deploy a branch to `Heroku`
2. Populate `Heroku` with a subset of the `Content::Item`s and `Link`s (about 1000)

Once a branch is deployed to `Heroku` it runs in `development` mode, so [GDS::SSO][1] mocks [Signon][2] and [authenticates automatically to the first user of the database][3]. This means, that only one user will exist in Heroku, and we have no easy way to authenticate other users.

This PR overcomes this issue by providing a simple authentication mechanism based on an URL parameter: `heroku_user_id` and the user session. This is obviously intended to be used only in `Heroku`.

### About the solution

1. I'd rather not creating a PR to support multiple users in [GDS-SSO][1] as this would be a temporary solution that would need to change once we move to PAAS.
2. I have created a helper class `lib/heroku.rb` that detects when the code runs in Heroku: checks for the existence of the environment variable `RUNNING_IN_HEROKU` and that we are not in `production` mode.
3. I have added an initialiser `config/initializers/heroku.rb` that overrides the default security mechanisms that are in place. 
4. I've removed the login information from the admin template, and I have added support for a select combobox to choose the user.

### After

image

[1]: https://github.com/alphagov/gds-sso
[2]: https://github.com/alphagov/signon
[3]: https://github.com/alphagov/gds-sso/blob/82ed8e549739a5ac807565ac49123e5dffff158d/lib/gds-sso/warden_config.rb#L83

### DIFF
--- a/config/initializers/heroku.rb
+++ b/config/initializers/heroku.rb
@@ -17,4 +17,8 @@ if Heroku.enabled?
       end
     end
   end
+
+  GovukAdminTemplate.configure do |c|
+    c.show_signout = false
+  end
 end
--- a/config/initializers/heroku.rb
+++ b/config/initializers/heroku.rb
@@ -1,6 +1,7 @@
 if Heroku.enabled?
   ApplicationController.class_eval do
     before_action :force_development!
+    before_action :set_default_user
 
     def authenticate_user!
       user_id = params.dig(:heroku, :user_id)
@@ -10,6 +11,10 @@ if Heroku.enabled?
 
     def current_user
       User.find(session[:heroku_user_id])
+    end
+
+    def set_default_user
+      session[:heroku_user_id] ||= User.first.id
     end
 
     private
--- a/config/initializers/heroku.rb
+++ b/config/initializers/heroku.rb
@@ -1,19 +1,22 @@
 if Heroku.enabled?
   ApplicationController.class_eval do
-    before_action :ensure_environment
+    before_action :force_development!
 
     def authenticate_user!
+      user_id = params.dig(:heroku, :user_id)
+
+      session[:heroku_user_id] = user_id if user_id.present?
     end
 
     def current_user
-      User.last
+      User.find(session[:heroku_user_id])
     end
 
     private
 
-    def ensure_environment
+    def force_development!
       if Rails.env.production?
-        raise "This code should never run in production"
+        raise 'This code should never run in production'
       end
     end
   end
--- a/config/initializers/heroku.rb
+++ b/config/initializers/heroku.rb
@@ -11,6 +11,9 @@ if Heroku.enabled?
 
     def current_user
       User.find(session[:heroku_user_id])
+    rescue
+      session.clear
+      User.first
     end
 
     def set_default_user
--- a/config/initializers/heroku.rb
+++ b/config/initializers/heroku.rb
@@ -1,0 +1,20 @@
+if Heroku.enabled?
+  ApplicationController.class_eval do
+    before_action :ensure_environment
+
+    def authenticate_user!
+    end
+
+    def current_user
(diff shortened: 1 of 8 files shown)


lib/heroku.rb
```diff
@@ -0,0 +1,13 @@
+class Heroku
+  def self.create_users(organisation_id)
+    organisation = Content::Item.find(organisation_id)
+
+    User.create!(
+      name: 'AT user',
+      email: 'uat@gds.com',
```
chao-xian: `.com`??? `.gov.uk` surely 😉

lib/heroku.rb
```diff
@@ -0,0 +1,13 @@
...
+      uid: SecureRandom.uuid,
```
mgrassotti: Wouldn't this eventually fail in the (very rare) situation in that the same uuid is generated twice?
We could add a loop to check if a uuid was already generated. I mean something like:
```
def generate_uid
  loop do
     random_uid = SecureRandom.urlsafe_base64(nil, false)
     break random_uid unless User.exists?(uid: random_uid)
   end
 end
```

spec/lib/heroku_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    context 'in development' do
+
+      it 'returns true if `RUNNING_IN_HEROKU` is present' do
+        expect(ENV).to receive(:[]).with("RUNNING_IN_HEROKU").and_return("TRUE")
+
+        expect(described_class.enabled?).to be true
+      end
+      it 'returns false otherwise' do
+        expect(ENV).to receive(:[]).with("RUNNING_IN_HEROKU").and_return(nil)
```
chao-xian: Is this supposed to have `.with("RUNNING_IN_HEROKU")`?

config/initializers/heroku.rb
```diff
@@ -1,24 +1,28 @@
 if Heroku.enabled?
   ApplicationController.class_eval do
-    before_action :ensure_environment
+    before_action :force_development!
```
chao-xian: Should this be a separate commit from the disabling of the GA?

lib/tasks/heroku.rake
```diff
@@ -43,12 +43,12 @@ namespace :heroku do
     system %{heroku pg:psql -c "DELETE FROM CONTENT_ITEMS" --app #{app_name}}
     system %{heroku pg:psql -c "DELETE FROM USERS" --app #{app_name}}
 
-    # Creates default user
-    system %{heroku run rails runner "Heroku.create_users(#{organisation.id})" --app #{app_name}}
-
     #Import Contetn Items and Links
```
chao-xian: Small typo

spec/lib/heroku_spec.rb
pmanrubia: I would say so. If the environment variable is not set then it always should return false?

lib/heroku.rb
pmanrubia: @mgrassotti ok, I will try with the following then:

```ruby
uid: "uid-#{index}"
```

pmanrubia:
@mgrassotti, @chao-xian thank you for reviewing it 👍

spec/lib/heroku_spec.rb
chao-xian: I was confused as I didn't know that `expect` can also implicitly behave like an `allow` as well as assert. Thanks for clarifying!

PR #390, opened 2017-10-30
Remove logged-in constraint for iframe proxy

alecgibson:
The iframe proxy requires that the user is logged in before serving
content. Since our Heroku deployment doesn't have a logged-in user,
this breaks the proxy.

This change removes this constraint, so the proxy should work on Heroku.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -37,11 +37,5 @@ Rails.application.routes.draw do
     mount GovukAdminTemplate::Engine, at: "/style-guide"
   end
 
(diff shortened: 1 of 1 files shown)


alecgibson:
[Prototype: All content](https://trello.com/c/MxGBtcet/621-prototype-all-content)

chao-xian:
[Prototype: Auditing a content item](https://trello.com/c/qxmGeXRS/622-prototype-auditing-a-content-item)

PR #392, opened 2017-10-30
Prototype DWP: add my content tab

pmanrubia:
[Trello card](https://trello.com/c/QdB4CUDS/624-prototype-my-content)

Add explicit support for content assigned to a user.
With this feature we want to user test if the user is
better placed within the application.

This is a prototype that should **never go into production.**

### After

image

### DIFF
--- a/app/views/audits/audits/index.html.erb
+++ b/app/views/audits/audits/index.html.erb
@@ -9,6 +9,6 @@
 
 <%= render "no_content_to_audit" if content_items.empty? %>
 <% if @first_item %>
-    <%= link_to "Start audit of first item", content_item_audit_path(@first_item), class: "btn btn-success", style: 'margin-bottom:10px;' %>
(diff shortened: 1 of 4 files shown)


app/views/audits/audits/index.html.erb
```diff
@@ -8,4 +8,7 @@
 <% end %>
 
 <%= render "no_content_to_audit" if content_items.empty? %>
+<% if @first_item %>
+    <%= link_to "Start audit of first item", content_item_audit_path(@first_item), class: "btn btn-success", style: 'margin-bottom:10px;' %>
```
theKHutDeveloper: This is not passing in the filter parameters so when user clicks to go to next item it is not the next item in the list

pmanrubia: Nice one! I have fixed it 👍

PR #395, opened 2017-10-31
[WIP] Fix - Precompute Report Rows

mgrassotti:
Trello: https://trello.com/c/wSruwud5/627-csv-export-broken

The CSV export was broken due to an unexpected Array object instead of
a Hash, returned by the Audits::Report `rows` method.

A new method of retrieving reports rows was introduced, but this is not
supported by the old values anymore.

This fix aims to set data for all the item involved in the reports
generations, triggering their `after_save` callbacks.

### DIFF
--- a/app/domain/audits/report.rb
+++ b/app/domain/audits/report.rb
@@ -39,9 +39,11 @@ module Audits
 
     def rows
       @rows ||= content_items
-        .joins(:report_row)
+        .includes(:audit)
+        .includes(:linked_primary_publishing_organisation)
+        .includes(:linked_organisations)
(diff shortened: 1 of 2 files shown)


chao-xian:
@mgrassotti after discussion with @pmanrubia, am going to close this PR for now and attach it to https://trello.com/c/lgJVfNB6 as the original bug is fixed by the data cleanup I did. We may not need to do this work at all as "improvement of content" stream of work may supersede the report generation aspect of the Audit Tool in the future. We can always come back to this PR if we get to work on this tech debt. Hope that's ok.

PR #397, opened 2017-10-31
Updates to rake task to upload to heroku

pmanrubia:
Addresses the following changes to the rake task
to deploy a branch to Heroku

* Use an identifier instead of a `pr_number`
* Use a default for an organisation
* Upload topics to `Heroku`


pmanrubia:
@chao-xian, Thank you for your review; I have updated the documentation.

PR #398, opened 2017-11-06
Refactor filtering feature tests into declarative style

chao-xian:
- Refactors the feature specs for filtering by using a semantic DSL for describing the filter page elements using the Page Object Model pattern.
- Changes the style the feature specs are written in to add clear separation between the Arrange, Act, Assert parts of the test.

See:
  - https://docs.publishing.service.gov.uk/manual/testing.html
  - https://martinfowler.com/bliki/PageObject.html
  - https://robots.thoughtbot.com/better-acceptance-tests-with-page-objects
  - https://github.com/natritmeyer/site_prism
  - https://about.futurelearn.com/blog/how-we-write-readable-feature-tests-with-rspec

### DIFF
--- a/app/views/audits/audits/_content_items.html.erb
+++ b/app/views/audits/audits/_content_items.html.erb
@@ -1,6 +1,6 @@
 <%= render 'audits/common/counter', count: content_items.total_count  %>
 
-<table class="table table-bordered table-hover" data-test="filter-list">
+<table class="table table-bordered table-hover" data-test-id="filter-list">
   <thead>
   <tr class="table-header">
     <td>Title</td>
--- a/app/views/audits/audits/_content_items.html.erb
+++ b/app/views/audits/audits/_content_items.html.erb
@@ -1,6 +1,6 @@
 <%= render 'audits/common/counter', count: content_items.total_count  %>
 
-<table class="table table-bordered table-hover filter-list">
+<table class="table table-bordered table-hover" data-test="filter-list">
   <thead>
   <tr class="table-header">
     <td>Title</td>
@@ -8,7 +8,7 @@
   </tr>
   </thead>
   <tbody>
-  <%= render collection: content_items, partial: "content_item", class: "filter-list" %>
+  <%= render collection: content_items, partial: "content_item" %>
   </tbody>
 </table>
 
--- a/app/views/audits/audits/_content_items.html.erb
+++ b/app/views/audits/audits/_content_items.html.erb
@@ -1,6 +1,6 @@
 <%= render 'audits/common/counter', count: content_items.total_count  %>
 
-<table class="table table-bordered table-hover">
+<table class="table table-bordered table-hover filter-list">
   <thead>
   <tr class="table-header">
     <td>Title</td>
@@ -8,7 +8,7 @@
   </tr>
   </thead>
   <tbody>
-  <%= render collection: content_items, partial: 'content_item' %>
+  <%= render collection: content_items, partial: "content_item", class: "filter-list" %>
   </tbody>
 </table>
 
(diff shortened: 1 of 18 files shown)


chao-xian:
[(5) Use declarative style for feature tests related to Filtering content and pagination](https://trello.com/c/l5q8Qloj/644-5-use-declarative-style-for-feature-tests-related-to-filtering-content-and-pagination)

spec/support/pages/audit/content_audit_tool.rb
```diff
@@ -4,4 +4,8 @@ class ContentAuditTool
   def audit_content_item
     AuditContentItemPage.new
   end
+
+  def filter_audit_list_page
```
andrewgarner: I think we should try and adopt terms closest to how our users would describe the pages of the tool. This page has the title 'Audit content' in the active tab? Maybe this could be `audit_content_page`?

spec/features/audit/lists/filter_spec.rb
```diff
@@ -33,14 +29,17 @@
...
+      filter_audit_list.filter_form do |form|
+        expect(form).to have_select("allocated_to", selected: "Me")
+        expect(form.audit_status).to have_checked_field("Not audited")
+      end
 
-      expect(page).to have_content("The Famous Five")
-      expect(page).to have_no_content("The Secret Seven")
-      expect(page).to have_no_content("The Wishing Chair")
+      expect(filter_audit_list).to have_list(text: "The Famous Five")
+      expect(filter_audit_list.list).to have_no_content("The Secret Seven")
```
andrewgarner: Can you do the following here:

```rb
expect(filter_audit_list).to have_no_list(text: "The Secret Seven")
```

(I think that is the documented approach – https://github.com/natritmeyer/site_prism#testing-that-an-element-does-not-exist)

spec/features/audit/lists/filter_spec.rb
```diff
@@ -91,166 +90,222 @@
...
+          form.audit_status.choose "All"
+          form.apply_filters.click
 
-        click_on "Apply filters"
+          expect(form.primary_orgs).to have_checked_field
 
-        expect(page).to have_content("VAT")
-        expect(page).to have_no_content("Tree felling")
+          form.organisations.select "HMRC"
+          # select "HMRC", from: "Organisations"
```
andrewgarner: What are we alluding to with this comment that we can't infer from the line of code above it?

app/views/audits/audits/_content_items.html.erb
```diff
@@ -1,14 +1,14 @@
 <%= render 'audits/common/counter', count: content_items.total_count  %>
 
-<table class="table table-bordered table-hover">
+<table class="table table-bordered table-hover filter-list">
```
KyleMacPherson: [Minor] Could potentially use a data-test attribute to identify elements in the tests rather than changing the class

chao-xian: Ah I didn't know about `data-test` as a convention. I prefer that so I'll make some changes. Thanks!

KyleMacPherson: No worries!

spec/support/pages/audit/content_audit_tool.rb
chao-xian: Addressed in https://github.com/alphagov/content-performance-manager/pull/398/commits/64467801e89fd50bc662f181cb73bbd93e6d0728

spec/features/audit/lists/filter_spec.rb
```diff
@@ -1,259 +1,461 @@
...
+    and_filtering_to_unaudited_content_allocated_to_me_by_default
+    then_the_filtered_list_shows_content_allocated_to_me
+    and_unaudited_content_not_allocated_to_me_is_not_shown
+  end
+
+  scenario "filtering audited content" do
+    given_content_belonging_to_departments
+    when_viewing_content_to_audit
+    and_filtering_to_audited_content_allocated_to_anyone
+    then_the_filter_list_shows_audited_item
```
andrewgarner: What's the difference between a "filter list" and a "list" and between an "item" (I think that should be "items") and "content"?

PR #402, opened 2017-11-08
Fix second page allocations

andrewgarner:
This fixes the logic we have to calculate which content items to allocate when a batch size is specified that is greater than the 100 items we display on a page and when the assignment is instigated from page 2 or greater.

We calculate an offset value which is the total number of content items that would have been displayed on all previous pages. We then assign X number of content items from there onwards.

### DIFF
--- a/app/controllers/audits/allocations_controller.rb
+++ b/app/controllers/audits/allocations_controller.rb
@@ -1,7 +1,7 @@
 module Audits
   class AllocationsController < BaseController
     before_action :set_default_parameters, only: :index
-    before_action :set_batch_value, only: %w(destroy create), if: -> { batch_size > content_ids.size }
+    before_action :set_batch_values, only: %w(destroy create), if: -> { batch_size > content_ids.size }
 
     decorates_assigned :content_items
 
@@ -42,11 +42,13 @@ module Audits
       params.fetch(:allocate_to)
     end
 
-    def set_batch_value
-      filter = params_to_filter
-      filter.per_page = batch_size
-
-      params[:content_ids] = FindContent.paged(filter).pluck(:content_id)
+    def set_batch_values
+      params[:content_ids] = FindContent
+                               .batch(
+                                 params_to_filter,
+                                 batch_size: batch_size,
+                                 from_page: params_to_filter.page,
+                               ).pluck(:content_id)
     end
 
     def batch_size
--- a/spec/features/audit/allocation/allocate_spec.rb
+++ b/spec/features/audit/allocation/allocate_spec.rb
@@ -126,7 +126,7 @@ RSpec.feature "Allocate multiple content items", type: :feature do
     let!(:books) do
       create_list(
         :content_item,
-        20,
+        60,
         primary_publishing_organisation: novelists,
       )
     end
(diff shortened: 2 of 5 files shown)


andrewgarner:
[Content allocation from second page broken](https://trello.com/c/PqURHXRc/631-content-allocation-from-second-page-broken)

spec/features/audit/allocation/allocate_spec.rb
```diff
@@ -126,36 +126,83 @@
...
+
+      expect(page).to have_content('210 items')
+
+      # Collect the titles that span each page, 100 per page
+      page_1_titles = content_item_titles
+      go_to_page 3
+      page_3_titles = content_item_titles
+      go_to_page 2
+
+      select 'Me', from: 'allocate_to'
```
pmanrubia: Isn't this a good opportunity to move into the new style of testing? We are adding new tests with the old way of doing things.

spec/features/audit/allocation/allocate_spec.rb
```diff
@@ -126,36 +126,83 @@
...
+        60,
         primary_publishing_organisation: novelists,
       )
     end
 
     let!(:paintings) do
       create_list(
         :content_item,
-        20,
+        60,
```
pmanrubia: Why have we incremented to 60 the number of content items?

spec/features/audit/allocation/allocate_spec.rb
```diff
@@ -126,36 +126,83 @@
...
       click_on "Assign"
 
-      expect(page).to have_content("26 items assigned to Jane Austen")
-      expect(page).to have_content("14 items")
+      expect(page).to have_content("101 items assigned to Jane Austen")
+      expect(page).to have_content("19 items")
+    end
+  end
+
+  context "There three pages of content items" do
```
pmanrubia: This feels more like a unit test for allocation, and not a feature test. 
It is not describing a feature from a user point of view, but just exercising an edge case.

spec/features/audit/allocation/allocate_spec.rb
andrewgarner: We previously updated our pages to have 100 content items per page, but didn't update this test to span multiple pages (as it used to). This change creates more items so that it spans multiple pages.

app/controllers/audits/allocations_controller.rb
```diff
@@ -37,13 +38,20 @@ def destroy
 
   private
 
+    PAGE_SIZE = 100
```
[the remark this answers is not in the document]
andrewgarner: I've partially addressed this with 88158a7fbeb54d7fa188b2b67e17a954cfd83568 by moving the default value to the filter object so that it is accessible in this context.

I'm generally uncomfortable with the concern of paging being mixed in with filtering but I didn't want to address that here and instead think about it with https://trello.com/c/seNXaaRp

app/controllers/audits/allocations_controller.rb
```diff
@@ -37,13 +38,20 @@ def destroy
...
     end
 
     def set_batch_value
+      page = params[:page].to_i
+      offset = page * PAGE_SIZE - PAGE_SIZE if page > 1
+
       params[:content_ids] = FindContent
-                               .paged(filter(per_page: batch_size))
+                               .paged(filter)
+                               .limit(batch_size)
```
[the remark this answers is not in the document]
andrewgarner: Again, I've partially addressed that with 5a8cfdfb6b69cb2e752bcdd0ff8a9451bc079085.

`FindContent` returns a scope, we can then apply our offset and limit to fetch the content item batch we need to allocate. I think this could be further extracted into a use case outside of the controller but again I wanted to wait until we've thought more about the concern of filtering with https://trello.com/c/seNXaaRp

spec/features/audit/allocation/allocate_spec.rb
andrewgarner: The page objects are not available to work with in master yet so I chose not to step on the shoes of other branches.

spec/features/audit/allocation/allocate_spec.rb
andrewgarner: I agree, but I didn't want to go too far with refactoring here. I think we should look to extract a use case for allocation that we can then test without the interface. I wondered if that might come up with https://trello.com/c/uBOWGdpD and https://trello.com/c/seNXaaRp

spec/features/audit/allocation/allocate_spec.rb
pmanrubia: I have added: https://github.com/alphagov/content-performance-manager/pull/402/commits/7a665d3459b1e21e7afbd28989dcc21db90c252c, as a temporary commit with some `draft` code to move the logic into the `FindContent` object.
It will also allow us to test this method in isolation without having a feature test.

We still need to sort out the pagination issue (we are 100 set up in many places), but this would be out of the scope of this PR. 

Any thoughts?

andrewgarner: I've re-worked c58412c50204ef5745b2e5fa4743437801a2a7ec incorporating your suggestions. That's allowed me to move the test coverage down the pyramid. The testing is a bit clunky due to the behaviour that is contained within `Content::Query` but I've left that for another day!

PR #374, opened 2017-10-17
Refactor audit feature specs

andrewgarner:
- Refactors the feature specs by introducing a semantic DSL for describing the audit page elements using the Page Object Model pattern.
- Changes the style the feature specs are written in to add clear separation between the Arrange, Act, Assert parts of the test.

See:
  - https://docs.publishing.service.gov.uk/manual/testing.html
  - https://martinfowler.com/bliki/PageObject.html
  - https://robots.thoughtbot.com/better-acceptance-tests-with-page-objects
  - https://github.com/natritmeyer/site_prism
  - https://about.futurelearn.com/blog/how-we-write-readable-feature-tests-with-rspec

PR #406, opened 2017-11-10
Refactor audit report feature tests

KyleMacPherson:
* Refactors the audit report feature specs by introducing a semantic DSL for describing the report page elements using the Page Object Model pattern.
* Changes the style the feature specs are written in to add clear separation between the Arrange, Act, Assert parts of the test.

See:

https://docs.publishing.service.gov.uk/manual/testing.html
https://martinfowler.com/bliki/PageObject.html
https://robots.thoughtbot.com/better-acceptance-tests-with-page-objects
https://github.com/natritmeyer/site_prism
https://about.futurelearn.com/blog/how-we-write-readable-feature-tests-with-rspec

Following on from these previous PRs which refactor other specs to follow a similar pattern:
https://github.com/alphagov/content-performance-manager/pull/374
https://github.com/alphagov/content-performance-manager/pull/398

### DIFF
--- a/spec/features/audit/report/allocation_spec.rb
+++ b/spec/features/audit/report/allocation_spec.rb
@@ -1,34 +1,38 @@
 RSpec.feature "Content Allocation", type: :feature do
   scenario "Filter allocated content" do
-    given_i_am_auditing_a_content_item
-    then_i_see_all_content_items
-    given_that_i_filter_by_attributes_that_do_not_apply_to_any_content_items
-    then_i_see_no_content_items
+    given_i_am_an_auditor_belonging_to_an_organisation
+    when_i_visit_the_report_page
+    then_i_can_see_the_content_item_i_have_been_allocated
+    when_i_select_no_one_in_the_allocated_to_filter
+    then_i_cannot_see_the_content_item_i_have_been_allocated
   end
 
-  def given_i_am_auditing_a_content_item
+  def given_i_am_an_auditor_belonging_to_an_organisation
     organisation = create(:organisation)
     user = create(:user, organisation: organisation)
-    content_item = create(
+    @content_item = create(
       :content_item,
       title: "Do Androids Dream of Electric Sheep",
       allocated_to: user,
       primary_publishing_organisation: organisation,
     )
-    @audit_report_page = ContentAuditTool.new.audit_report_page
-    @audit_report_page.load(content_id: content_item.content_id)
   end
 
-  def then_i_see_all_content_items
+  def when_i_visit_the_report_page
+    @audit_report_page = ContentAuditTool.new.audit_report_page
+    @audit_report_page.load(content_id: @content_item.content_id)
+  end
+
+  def then_i_can_see_the_content_item_i_have_been_allocated
     expect(@audit_report_page).to have_report_section(text: '1')
   end
 
-  def given_that_i_filter_by_attributes_that_do_not_apply_to_any_content_items
+  def when_i_select_no_one_in_the_allocated_to_filter
     @audit_report_page.allocated_to.select 'No one'
     @audit_report_page.apply_filters.click
   end
 
-  def then_i_see_no_content_items
+  def then_i_cannot_see_the_content_item_i_have_been_allocated
     expect(@audit_report_page).to have_report_section(text: '0')
   end
 end
--- a/spec/features/audit/report/allocation_spec.rb
+++ b/spec/features/audit/report/allocation_spec.rb
@@ -1,31 +1,34 @@
 RSpec.feature "Content Allocation", type: :feature do
-  let!(:my_organisation) do
-    create(
-      :organisation,
-    )
-  end
-
-  let!(:me) do
-    create(
-      :user,
-      organisation: my_organisation,
-    )
-  end
-
   scenario "Filter allocated content" do
-    create(
+    given_i_am_auditing_a_content_item
+    then_i_see_all_content_items
+    given_that_i_filter_by_attributes_that_do_not_apply_to_any_content_items
+    then_i_see_no_content_items
+  end
+
+  def given_i_am_auditing_a_content_item
+    organisation = create(:organisation)
+    user = create(:user, organisation: organisation)
+    content_item = create(
       :content_item,
       title: "Do Androids Dream of Electric Sheep",
-      allocated_to: me,
-      primary_publishing_organisation: my_organisation,
+      allocated_to: user,
+      primary_publishing_organisation: organisation,
     )
+    @audit_report_page = ContentAuditTool.new.audit_report_page
+    @audit_report_page.load(content_id: content_item.content_id)
+  end
 
-    visit audits_report_path
+  def then_i_see_all_content_items
+    expect(@audit_report_page).to have_report_section(text: '1')
+  end
 
-    expect(page).to have_selector(".report-section", text: "1")
+  def given_that_i_filter_by_attributes_that_do_not_apply_to_any_content_items
+    @audit_report_page.allocated_to.select 'No one'
+    @audit_report_page.apply_filters.click
+  end
 
-    select "No one", from: "allocated_to"
-    click_on "Apply filters"
-    expect(page).to have_selector(".report-section", text: "0")
+  def then_i_see_no_content_items
+    expect(@audit_report_page).to have_report_section(text: '0')
   end
 end
--- a/spec/features/audit/report/allocation_spec.rb
+++ b/spec/features/audit/report/allocation_spec.rb
@@ -24,7 +24,7 @@ RSpec.feature "Content Allocation", type: :feature do
   end
 
   def then_i_can_see_the_content_item_i_have_been_allocated
-    expect(@audit_report_page).to have_report_section(text: '1')
+    expect(@audit_report_page).to have_content_item_count(1)
   end
 
   def when_i_select_no_one_in_the_allocated_to_filter
--- a/spec/features/audit/report/allocation_spec.rb
+++ b/spec/features/audit/report/allocation_spec.rb
@@ -24,7 +24,7 @@ RSpec.feature "Content Allocation", type: :feature do
   end
 
   def then_i_can_see_the_content_item_i_have_been_allocated
-    expect(@audit_report_page).to have_content_item_count(1)
+    expect(@audit_report_page.content_item_count.text).to eq('1')
   end
 
   def when_i_select_no_one_in_the_allocated_to_filter
@@ -33,6 +33,6 @@ RSpec.feature "Content Allocation", type: :feature do
   end
 
   def then_i_cannot_see_the_content_item_i_have_been_allocated
-    expect(@audit_report_page).to have_report_section(text: '0')
+    expect(@audit_report_page.content_item_count.text).to eq('0')
   end
 end
--- a/spec/features/audit/report/csv_export_spec.rb
+++ b/spec/features/audit/report/csv_export_spec.rb
@@ -8,63 +8,36 @@ RSpec.feature "Exporting a CSV from the report page" do
   end
 
   scenario "Exporting a csv file as an attachment" do
-    given_i_am_an_auditor_belonging_to_an_organisation_with_audits
+    given_i_am_an_auditor_belonging_to_an_organisation
     when_i_export_an_audit_report
     then_i_receive_the_report_in_csv_format
   end
 
   scenario "Applying the filters to the export" do
-    given_i_am_an_auditor_belonging_to_an_organisation_with_audits
-    and_they_are_filtered
+    given_i_am_an_auditor_belonging_to_an_organisation
+    and_filter_is_applied_to_show_only_hmrc_content_items
     when_i_export_the_reports
-    then_i_receive_an_audit_progress_report_for_filtered_audits_in_csv_format
+    then_i_receive_only_hmrc_related_content_items_in_the_report
   end
 
   scenario "Multiple pages including_details_of_audit_progress_ of content items are in the database" do
-    given_i_am_an_auditor_belonging_to_an_organisation_with_audits
+    given_i_am_an_auditor_belonging_to_an_organisation
     and_there_are_multiple_pages_of_unfiltered_content_items
     when_i_export_all_the_reports
-    then_i_receive_an_audit_progress_report_for_all_audits_in_csv_format
+    then_i_receive_an_audit_progress_report_for_all_content_items_in_csv_format
   end
 
   scenario "Discard audit status filter when clicking from content view to report, and then exporting CSV" do
-    given_i_am_an_auditor_belonging_to_an_organisation_with_audits
-    when_i_apply_filters
-    then_i_see_the_two_content_items_which_match_the_filters
+    given_i_am_an_auditor_belonging_to_an_organisation
+    and_there_are_two_unassigned_content_items
+    when_i_filter_the_content_page_for_content_items_with_no_one_assigned_to_them
+    then_i_see_the_two_unassigned_content_items
     when_i_navigate_to_audits_report_page
-    then_i_see_one_content_item_which_matches_default_filters
+    then_i_do_not_see_the_two_unassigned_content_items
+    and_i_see_the_one_content_item_assigned_to_me
   end
 
-  def given_i_am_exporting_an_audit_report
-    user = create(:user)
-    hmrc = create(
-      :content_item,
-      title: "HMRC",
-      document_type: "organisation"
-    )
-    example1 = create(:content_item,
-                      title: "Example 1",
-                      base_path: "/example1",
-                      allocated_to: user)
-    create(:link,
-           source_content_id: example1.content_id,
-           target_content_id: hmrc.content_id,
-           link_type: "primary_publishing_organisation")
-    create(:content_item,
-           title: "Example 2",
-           base_path: "/example2",
-           allocated_to: nil)
-    create(:content_item,
-           title: "Example 3",
-           base_path: "/example3",
-           allocated_to: nil)
-    create(:audit, content_item: example1)
-    @audit_report = ContentAuditTool.new.audit_report_page
-    @audit_report.load(content_id: Content::Item.last.content_id)
-    @audit_report.export_to_csv.click
-  end
-
-  def given_i_am_an_auditor_belonging_to_an_organisation_with_audits
+  def given_i_am_an_auditor_belonging_to_an_organisation
     user = create(:user)
     hmrc = create(
       :content_item,
@@ -80,24 +53,16 @@ RSpec.feature "Exporting a CSV from the report page" do
            source_content_id: example1.content_id,
            target_content_id: hmrc.content_id,
            link_type: "primary_publishing_organisation")
-    create(:content_item,
-           title: "Example 2",
-           base_path: "/example2",
-           allocated_to: nil)
-    create(:content_item,
-           title: "Example 3",
-           base_path: "/example3",
-           allocated_to: nil)
     create(:audit, content_item: example1)
   end
 
-  def then_i_receive_an_audit_progress_report_for_filtered_audits_in_csv_format
+  def then_i_receive_only_hmrc_related_content_items_in_the_report
     expect(@audit_report).to have_content("Title,URL")
     expect(@audit_report).to have_content("Example 1,https://gov.uk/example1")
     expect(@audit_report).to have_no_content("Example 2,https://gov.uk/example2")
   end
 
-  def then_i_see_the_two_content_items_which_match_the_filters
+  def then_i_see_the_two_unassigned_content_items
     expect(@audit_content_page).to have_no_content("Example 1")
     expect(@audit_content_page).to have_content("Example 2")
     expect(@audit_content_page).to have_content("Example 3")
@@ -107,10 +72,13 @@ RSpec.feature "Exporting a CSV from the report page" do
     @audit_content_page.audits_progress_tab.click
   end
 
-  def then_i_see_one_content_item_which_matches_default_filters
+  def then_i_do_not_see_the_two_unassigned_content_items
     expect(@audit_report_page).to be_displayed
-    expect(@audit_report_page).to have_content('1 Content items')
     expect(@audit_report_page).to have_no_content('2 Content items')
+  end
+
+  def and_i_see_the_one_content_item_assigned_to_me
+    expect(@audit_report_page).to have_content('1 Content items')
   end
 
   def when_i_export_an_audit_report
@@ -129,10 +97,9 @@ RSpec.feature "Exporting a CSV from the report page" do
     expect(@audit_report).to have_content("Example 1,https://gov.uk/example1")
   end
 
-  def and_they_are_filtered
+  def and_filter_is_applied_to_show_only_hmrc_content_items
     @audit_report = ContentAuditTool.new.audit_report_page
     @audit_report.load(content_id: Content::Item.last.content_id)
-    @audit_report.allocated_to.select 'Anyone'
     @audit_report.organisations.select 'HMRC'
     @audit_report.apply_filters.click
   end
@@ -152,13 +119,13 @@ RSpec.feature "Exporting a CSV from the report page" do
     click_link "Export filtered audit to CSV"
   end
 
-  def then_i_receive_an_audit_progress_report_for_all_audits_in_csv_format
+  def then_i_receive_an_audit_progress_report_for_all_content_items_in_csv_format
     csv = CSV.parse(page.body)
     number_of_metadata_rows = 1
     expect(csv.count).to eq(Content::Item.count + number_of_metadata_rows)
   end
 
-  def when_i_apply_filters
+  def when_i_filter_the_content_page_for_content_items_with_no_one_assigned_to_them
     @audit_content_page = ContentAuditTool.new.audit_content_page
     @audit_report_page = ContentAuditTool.new.audit_report_page
     @audit_content_page.load
@@ -166,5 +133,14 @@ RSpec.feature "Exporting a CSV from the report page" do
       form.allocated_to.select "No one"
       form.apply_filters.click
     end
+  end
+
+  def and_there_are_two_unassigned_content_items
+    create(:content_item,
+           title: "Example 2",
+           base_path: "/example2")
+    create(:content_item,
+           title: "Example 3",
+           base_path: "/example3")
   end
 end
--- a/spec/features/audit/report/csv_export_spec.rb
+++ b/spec/features/audit/report/csv_export_spec.rb
@@ -8,25 +8,31 @@ RSpec.feature "Exporting a CSV from the report page" do
   end
 
   scenario "Exporting a csv file as an attachment" do
-    given_i_am_exporting_an_audit_report
-    then_i_receive_an_audit_progress_report_in_csv_format
+    given_i_am_an_auditor_belonging_to_an_organisation_with_audits
+    when_i_export_an_audit_report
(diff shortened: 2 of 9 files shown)


KyleMacPherson:
[(3) Use declarative style for feature tests related to Reporting](https://trello.com/c/DR4Utkno/645-3-use-declarative-style-for-feature-tests-related-to-reporting)

spec/features/audit/report/allocation_spec.rb
```diff
@@ -1,31 +1,34 @@
...
-
-  let!(:me) do
-    create(
-      :user,
-      organisation: my_organisation,
-    )
+  scenario "Filter allocated content" do
+    given_i_am_auditing_a_content_item
+    then_i_see_all_content_items
+    given_that_i_filter_by_attributes_that_do_not_apply_to_any_content_items
```
alecgibson: According to standard [Cucumber parlance](https://github.com/cucumber/cucumber/wiki/Given-When-Then), you should use `given` to "put the system in a known state before the user (or external system) starts interacting with the system".

You should use `when` here, as this step is here "to describe the key action the user performs".

As a rough rule of thumb, I'd only ever expect to see "Given" at the start of a test.

In general, this test is written a little vaguely, too. I'd probably expect a bit more detail about what exactly it is we're testing. For example, our scenario is filtering *allocated* content, but nowhere in your test do you mention the concept of allocation. I'd expect maybe:

> Given I have been allocated a content item
> When I visit the report page
> Then I can see the content item I have been allocated
> When I select "no one" in the "allocated to" filter
> Then I cannot see the content item I have been allocated

spec/features/audit/report/allocation_spec.rb
```diff
@@ -1,31 +1,34 @@
...
+    then_i_see_no_content_items
   end
 
-  scenario "Filter allocated content" do
-    create(
+  def given_i_am_auditing_a_content_item
+    organisation = create(:organisation)
```
alecgibson: I find it a bit odd hiding all this setup in this step. For example, this step has nothing to do with me belonging to an organisation (or existing for that matter!). I'd expect this to exist either as some sort of common setup step like`given_i_am_an_auditor_belonging_to_an_organisation`

alecgibson: As a corollary to this, you may well need to make `user` and `organisation` global variables, unless we've come up with a smarter way of doing that somewhere else?

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,145 @@ def content_disposition
...
-      visit audits_report_path
-      click_link "Export filtered audit to CSV"
-
-      expect(content_type).to eq("text/csv")
-      expect(content_disposition).to start_with("attachment")
-      expect(content_disposition).to include(
-        'filename="Transformation_audit_report_CSV_download.csv"',
-      )
+  scenario "Exporting a csv file as an attachment" do
+    given_i_am_exporting_an_audit_report
```
alecgibson: See above comment. This should be `when_i_export...`

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,145 @@ def content_disposition
...
+    then_i_receive_an_audit_progress_report_in_csv_format
+  end
 
-      expect(page).to have_content("Title,URL")
-      expect(page).to have_content("Example 1,https://gov.uk/example1")
-    end
+  scenario "Applying the filters to the export" do
+    given_i_have_applied_filters_and_exported_audits
```
alecgibson: `when_i_apply...`

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,145 @@ def content_disposition
...
+    then_i_receive_an_audit_progress_report_for_filtered_audits_in_csv_format
+  end
 
-    scenario "Applying the filters to the export" do
-      visit audits_report_path
-      select "Anyone", from: "allocated_to"
+  scenario "Multiple pagesincluding_details_of_audit_progress_ of content items are in the database" do
+    given_i_have_multiple_pages_of_content_items_and_do_not_filter_them
```
alecgibson:
```
given_i_have_multiple_pages_of_content_items
and_they_are_not_filtered
when_i_generate_a_csv_report
then_i_receive_an_audit...
```

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,145 @@ def content_disposition
...
+    then_i_receive_an_audit_progress_report_for_all_audits_in_csv_format
+  end
 
-      select "HMRC", from: "Organisations"
-      click_on "Apply filters"
+  scenario "Discard audit status filter when clicking from content view to report, and then exporting CSV" do
+    given_i_apply_filters_to_the_audit_content_page
```
alecgibson: `when_i_apply...`

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,145 @@ def content_disposition
...
+    then_i_see_filtered_audits
+    given_that_i_navigate_to_audits_report_page
```
alecgibson: `when_i_navigate...`

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,145 @@ def content_disposition
...
+    then_i_see_an_unfiltered_audit_progress_report
+  end
 
-      click_link "Export filtered audit to CSV"
-      expect(page).to have_content("Example 1,https://gov.uk/example1")
-      expect(page).to have_no_content("Example 2,https://gov.uk/example2")
-    end
+  def given_i_am_exporting_an_audit_report
```
alecgibson: My above comments also apply to this setup step - there's too much happening in here that isn't hinted at by the name of the method.

spec/features/audit/report/allocation_spec.rb
```diff
@@ -1,31 +1,34 @@
...
+    content_item = create(
       :content_item,
       title: "Do Androids Dream of Electric Sheep",
-      allocated_to: me,
-      primary_publishing_organisation: my_organisation,
+      allocated_to: user,
+      primary_publishing_organisation: organisation,
     )
+    @audit_report_page = ContentAuditTool.new.audit_report_page
+    @audit_report_page.load(content_id: content_item.content_id)
```
alecgibson: This is a user action - the user visiting the page. It belongs in its own `when` step.

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,145 @@ def content_disposition
...
+    expect(content_disposition).to start_with("attachment")
+    expect(content_disposition).to include(
+      'filename="Transformation_audit_report_CSV_download.csv"',
+    )
+    expect(@audit_report).to have_content("Title,URL")
+    expect(@audit_report).to have_content("Example 1,https://gov.uk/example1")
+  end
 
-      choose "Audited"
+  def given_i_have_applied_filters_and_exported_audits
```
alecgibson: This method also contains too much setup and user actions.

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,145 @@ def content_disposition
...
+    expect(@audit_report).to have_content("Title,URL")
+    expect(@audit_report).to have_content("Example 1,https://gov.uk/example1")
+    expect(@audit_report).to have_no_content("Example 2,https://gov.uk/example2")
+  end
 
-      click_link "Audit progress"
+  def given_i_have_multiple_pages_of_content_items_and_do_not_filter_them
+    create(:user)
+    create_list(:content_item, 110)
+    visit audits_report_path
```
alecgibson: User actions don't belong in `given` setup steps.

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,145 @@ def content_disposition
...
+  def then_i_receive_an_audit_progress_report_for_all_audits_in_csv_format
+    click_link "Export filtered audit to CSV"
+    csv = CSV.parse(page.body, headers: true)
+    number_of_metadata_rows = 1
+    expect(csv.count).to eq(Content::Item.count + number_of_metadata_rows)
   end
 
-  context "Multiple pages of content items are in the database" do
-    let!(:content_items) { create_list(:content_item, 110) }
+  def given_i_apply_filters_to_the_audit_content_page
```
alecgibson: Too much setup.

spec/features/audit/report/progress_spec.rb
```diff
@@ -1,72 +1,77 @@
 RSpec.feature "Reporting on audit progress" do
-  let!(:my_organisation) do
-    create(
-      :organisation
-    )
+  scenario "Displaying the number of items included in the audit" do
+    given_i_have_been_allocated_audited_and_unaudited_content_items_belonging_to_my_organisation
```
alecgibson: See previous comments on Cucumber-style naming.

spec/support/pages/audit/content_audit_tool/audit_report_page.rb
```diff
@@ -0,0 +1,14 @@
+require 'site_prism/page'
+
+class AuditReportPage < SitePrism::Page
+  set_url '/audits/report'
+
+  element :report_section, '[data-test-id=report-section]'
+  element :allocated_to, '[data-test-id=allocated-to]'
+  element :apply_filters, '[data-test-id=apply-filters]'
+  element :content_type, '[data-test-id=document-type]'
+  element :audit_progress_bar, '[data-test-id=audit_progress_bar]'
```
alecgibson: Why do some of our test IDs use hyphens and others use underscores?

spec/features/audit/report/allocation_spec.rb
```diff
@@ -1,34 +1,38 @@
...
+
+  def when_i_visit_the_report_page
     @audit_report_page = ContentAuditTool.new.audit_report_page
-    @audit_report_page.load(content_id: content_item.content_id)
+    @audit_report_page.load(content_id: @content_item.content_id)
   end
 
-  def then_i_see_all_content_items
+  def then_i_can_see_the_content_item_i_have_been_allocated
     expect(@audit_report_page).to have_report_section(text: '1')
```
alecgibson: Why are we expecting this text? It's a bit obscure. What does it mean? If you're looking for the count of items, it would be more descriptive to have a method that looks like:

```ruby
expect(@audit_report_page).to have_content_item_count(1)
```

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -8,25 +8,31 @@ def content_disposition
...
     then_i_receive_an_audit_progress_report_for_all_audits_in_csv_format
   end
 
   scenario "Discard audit status filter when clicking from content view to report, and then exporting CSV" do
-    given_i_apply_filters_to_the_audit_content_page
-    then_i_see_filtered_audits
-    given_that_i_navigate_to_audits_report_page
-    then_i_see_an_unfiltered_audit_progress_report
+    given_i_am_an_auditor_belonging_to_an_organisation_with_audits
+    when_i_apply_filters
```
alecgibson: I still find this step a bit vague. What filters are you applying? What in your setup makes me expect that anything should be filtered by these filters (or not)? Maybe:

> Given I am an auditor belonging to an organisation
> And there are (unaudited, unallocated?) content items
> When I filter by (bla)
> Then I do (or don't) see them

spec/features/audit/report/progress_spec.rb
```diff
@@ -1,43 +1,34 @@
...
+    given_i_am_an_auditor_belonging_to_an_organisation_with_content_items
+    when_i_visit_the_report_page
     then_i_see_all_content_items
-    given_that_select_a_document_type_that_does_not_apply_to_any_content_item
+    when_i_select_a_document_type_that_does_not_apply_to_any_content_item
     then_i_see_no_content_items
-    given_that_select_a_document_type_that_applies_to_all_content_items
-    then_i_see_all_content_items
+    when_i_select_a_document_type_that_does_apply_to_content_items
+    then_i_see_how_many_content_items_it_applies_to
```
alecgibson: Again, don't shy away from being explicit. We're being very specific in the actual things we test, so we should describe those things at the high level:

> Given I ... belong to an organisation with **3** content items
> ...
> Then I see 3 content items

spec/features/audit/report/progress_spec.rb
```diff
@@ -1,43 +1,34 @@
...
   end
 
   scenario "Displaying the number of items audited/not audited" do
-    given_i_have_been_allocated_audited_and_unaudited_content_items_belonging_to_my_organisation
-    then_i_see_a_percentage_breakdown_of_audited_tasks_and_a_progress_bar
+    given_i_am_an_auditor_belonging_to_an_organisation_with_content_items
```
alecgibson: Again, please describe setup and what you're actually testing. Maybe:

> Given I belong to an organisation with 3 content items
> And 2 of those items have been audited
> When I visit the report page
> Then I see 67% of my audit has been completed
> And I see 33% of my audit is left to complete

spec/features/audit/report/allocation_spec.rb
```diff
@@ -1,31 +1,38 @@
...
+  def when_i_select_no_one_in_the_allocated_to_filter
+    @audit_report_page.allocated_to.select 'No one'
+    @audit_report_page.apply_filters.click
+  end
 
-    select "No one", from: "allocated_to"
-    click_on "Apply filters"
-    expect(page).to have_selector(".report-section", text: "0")
+  def then_i_cannot_see_the_content_item_i_have_been_allocated
+    expect(@audit_report_page).to have_report_section(text: '0')
```
alecgibson: `expect(@audit_report_page).to have_content_item_count(0)` ?

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,140 @@ def content_disposition
...
+    when_i_export_an_audit_report
+    then_i_receive_the_report_in_csv_format
+  end
 
-    scenario "Exporting a csv file as an attachment" do
-      visit audits_report_path
-      click_link "Export filtered audit to CSV"
+  scenario "Applying the filters to the export" do
+    given_i_am_an_auditor_belonging_to_an_organisation
+    and_filter_is_applied_to_show_only_hmrc_content_items
```
alecgibson: (`and_a_filter...`)

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,140 @@ def content_disposition
...
+      :content_item,
+      title: "HMRC",
+      document_type: "organisation",
+      allocated_to: user
+    )
+    example1 = create(:content_item,
+                      title: "Example 1",
+                      base_path: "/example1",
+                      allocated_to: user)
+    create(:link,
```
alecgibson: You shouldn't create these explicit links; it's more expressive (and quicker!) to do:

```ruby
create(
  :content_item,
  primary_publishing_organisation: hmrc,
  ...
)
```

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,140 @@ def content_disposition
...
+           source_content_id: example1.content_id,
+           target_content_id: hmrc.content_id,
+           link_type: "primary_publishing_organisation")
+    create(:audit, content_item: example1)
+  end
 
-      select "HMRC", from: "Organisations"
-      click_on "Apply filters"
+  def then_i_receive_only_hmrc_related_content_items_in_the_report
```
alecgibson: It might be nicer to name your examples a bit more descriptively:

```ruby
def then_i_receive_onlky_hmrc_related_content_items_in_the_report
  ...
  expect(@audit_report).to have_content("HMRC content...")
  expect(@audit_report).to have_no_content("Non-HMRC content...")
end
```

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,140 @@ def content_disposition
...
+    expect(@audit_report).to have_content("Example 1,https://gov.uk/example1")
+    expect(@audit_report).to have_no_content("Example 2,https://gov.uk/example2")
+  end
 
-      click_link "Export filtered audit to CSV"
-      expect(page).to have_content("Example 1,https://gov.uk/example1")
-      expect(page).to have_no_content("Example 2,https://gov.uk/example2")
-    end
+  def then_i_see_the_two_unassigned_content_items
+    expect(@audit_content_page).to have_no_content("Example 1")
```
alecgibson: As above, may be nice to call these things like "Assigned content #1", "Unassigned content #1", etc. which gives the person reading the test a bit more context than "Example 1" - and should highlight the key assumption about that content item in the construction.

For example, if I was being a bit trigger-happy with my refactoring of "Example 1", "Example 2", etc., I might accidentally set all the items to "unassigned", but if I see "Assigned content 1", maybe I'd think twice before changing that property.

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,140 @@ def content_disposition
...
-      visit audits_path
-      select "Anyone", from: "allocated_to"
+  def when_i_navigate_to_audits_report_page
+    @audit_content_page.audits_progress_tab.click
+  end
 
-      choose "Audited"
+  def then_i_do_not_see_the_two_unassigned_content_items
+    expect(@audit_report_page).to be_displayed
+    expect(@audit_report_page).to have_no_content('2 Content items')
```
alecgibson: This is a dangerous assertion to rely on. This test could incorrectly pass if:
- we have 0 content items
- we have 1 content item
- we have 3 content items
- (you see where I'm going)
- and _also_ if we **do** have 2 content items, but we've reworded the copy to read eg "2 items"

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,140 @@ def content_disposition
...
+  end
 
-      click_on "Apply filters"
-      expect(page).to have_content("Example 1")
-      expect(page).to have_no_content("Example 2")
+  def and_i_see_the_one_content_item_assigned_to_me
+    expect(@audit_report_page).to have_content('1 Content items')
```
alecgibson: Although this doesn't fall into the "not seeing" traps like the assertion above, it still falls into the trap of just asserting that you see one item, but not exactly caring what that item is. That's _not_ what your step is describing. In fact you care a lot about what that one item is (the one assigned to you) - so you should assert that that is the case.

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,140 @@ def content_disposition
...
+    expect(content_disposition).to include(
+      'filename="Transformation_audit_report_CSV_download.csv"',
+    )
+    expect(@audit_report).to have_content("Title,URL")
+    expect(@audit_report).to have_content("Example 1,https://gov.uk/example1")
+  end
+
+  def and_filter_is_applied_to_show_only_hmrc_content_items
+    @audit_report = ContentAuditTool.new.audit_report_page
+    @audit_report.load(content_id: Content::Item.last.content_id)
```
alecgibson: What does this line do? Why are we loading a particular content item for the report?

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,140 @@ def content_disposition
...
+    @audit_report.export_to_csv.click
+  end
 
-    scenario "Exporting an unfiltered audit to CSV with all the content items" do
-      visit audits_report_path
-      select "Anyone", from: "allocated_to"
-      click_on "Apply filters"
-      click_link "Export filtered audit to CSV"
+  def and_there_are_multiple_pages_of_unfiltered_content_items
+    create_list(:content_item, 100)
```
alecgibson: I think we show 100 items per page, so this is just 1 page?

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,140 @@ def content_disposition
...
-      expect(csv.count).to eq(content_items.count + number_of_metadata_rows)
+  def then_i_receive_an_audit_progress_report_for_all_content_items_in_csv_format
+    csv = CSV.parse(page.body)
+    number_of_metadata_rows = 1
+    expect(csv.count).to eq(Content::Item.count + number_of_metadata_rows)
+  end
+
+  def when_i_filter_the_content_page_for_content_items_with_no_one_assigned_to_them
+    @audit_content_page = ContentAuditTool.new.audit_content_page
+    @audit_report_page = ContentAuditTool.new.audit_report_page
```
alecgibson: Why do we need `@audit_report_page` here?

spec/features/audit/report/csv_export_spec.rb
```diff
@@ -13,99 +7,140 @@ def content_disposition
...
+    @audit_content_page.load
+    @audit_content_page.filter_form do |form|
+      form.allocated_to.select "No one"
+      form.apply_filters.click
     end
   end
+
+  def and_there_are_two_unassigned_content_items
+    create(:content_item,
+           title: "Example 2",
```
alecgibson: Would be nice to name this "Unassigned item 1" or similar.

spec/features/audit/report/progress_spec.rb
```diff
@@ -1,72 +1,102 @@
...
+  scenario "Audit status filter is not present" do
+    given_i_am_an_auditor_belonging_to_an_organisation_with_three_content_items
+    when_i_visit_the_report_page
+    then_i_do_not_see_audit_status_filter
+  end
 
-    scenario "Displaying the number of items included in the audit" do
-      visit audits_report_path
+  def then_i_see_the_report_includes_all_three_content_items
+    expect(@audit_report_page).to have_report_section(text: '3 Content items')
```
alecgibson: `expect(@audit_report_page).to have_content_item_count(3)` ?

spec/features/audit/report/progress_spec.rb
```diff
@@ -1,72 +1,102 @@
...
+  end
 
-      expect(page).to have_content("3 Content items")
+  def then_i_see_the_report_includes_no_content_items
+    expect(@audit_report_page).to have_report_section(text: '0 Content items')
```
alecgibson: `expect(@audit_report_page).to have_content_item_count(0)` ?

spec/features/audit/report/progress_spec.rb
```diff
@@ -1,72 +1,102 @@
 RSpec.feature "Reporting on audit progress" do
-  let!(:my_organisation) do
-    create(
-      :organisation
-    )
+  scenario "Displaying the number of items included in the audit" do
+    given_i_am_an_auditor_belonging_to_an_organisation_with_three_content_items
```
alecgibson: Good change. Could probably be more specific still? Sorry, I should have pointed this out in my last review. You could do something like:

> Given I ... three transactions
> ...
> When I filter by Transaction document type
> ...
> When I filter by Guide document type
> ...

spec/features/audit/report/progress_spec.rb
```diff
@@ -1,72 +1,102 @@
...
+  def and_two_items_have_been_audited
+    create(:passing_audit, content_item: @content_items[0])
+    create(:failing_audit, content_item: @content_items[1])
+  end
 
-    scenario "Displaying the number of items audited/not audited" do
-      visit audits_report_path
+  def when_i_visit_the_report_page
+    @audit_report_page = ContentAuditTool.new.audit_report_page
+    @audit_report_page.load(content_id: Content::Item.last.content_id)
```
alecgibson: Again, not sure why we need a content ID for the report page?

spec/features/audit/report/progress_spec.rb
```diff
@@ -1,72 +1,102 @@
...
+  end
 
-      expect(page).to have_content("Audited 2 67%")
-      expect(page).to have_content("Still to audit 1 33%")
-      expect(width("#progress")).to eq("width: 66.66667%;")
-    end
+  def when_i_select_a_document_type_that_does_not_apply_to_any_content_item
+    @audit_report_page.allocated_to.select 'Anyone'
```
alecgibson: Why do we need to apply this filter? Either we don't need it, or it should be explicitly documented in its own step.

spec/features/audit/report/allocation_spec.rb
KyleMacPherson: As discussed have gone for `have_content_item_count(text: 0)

spec/features/audit/report/csv_export_spec.rb
KyleMacPherson: √

spec/features/audit/report/csv_export_spec.rb
KyleMacPherson: Thanks for explaining this.

spec/features/audit/report/csv_export_spec.rb
KyleMacPherson: Good idea! 👍

spec/features/audit/report/csv_export_spec.rb
KyleMacPherson: √

spec/features/audit/report/csv_export_spec.rb
KyleMacPherson: As discussed yesterday, this test is asserting that our current filter is not persisted when we navigate to a new page. I test that we have exactly `1 content item` on the next line but broke the two assertions down into individual method names to try to help a slightly weird test to be more easily understood.

spec/features/audit/report/csv_export_spec.rb
KyleMacPherson: As discussed yesterday, this assertion is based on the report page which does not show individual content item details, just a summary of how many content items are in the report, how many have been audited etc.

spec/features/audit/report/csv_export_spec.rb
KyleMacPherson: √

spec/features/audit/report/csv_export_spec.rb
KyleMacPherson: Good point 👍

spec/features/audit/report/csv_export_spec.rb
KyleMacPherson: It is needed because we navigate to the audit report page in this test, however I had this line in the wrong method 😔 Have moved it to the `when_i_navigate_to_audits_report_page` method.

spec/features/audit/report/csv_export_spec.rb
KyleMacPherson: √

spec/features/audit/report/progress_spec.rb
KyleMacPherson: Have added an additional step in which I update the content items to give them `document_type: 'transaction'`. I just did this to give the test an extra step to improve clarity without taking away from the re-usability of the `given...` method. What do you think?

spec/features/audit/report/progress_spec.rb
KyleMacPherson: √

spec/features/audit/report/progress_spec.rb
KyleMacPherson: √

spec/features/audit/report/progress_spec.rb
KyleMacPherson: √

spec/features/audit/report/progress_spec.rb
KyleMacPherson: You're right, no need. Was just a hangover from previous tests.

PR #409, opened 2017-11-20
Filter by status in report

pmanrubia:
[Trello card](https://trello.com/c/E2I3wbPe/639-3-filter-the-export-by-audit-status)

### User need

As a **content auditor**
I **need** filter the audit progress tab and csv export by audit status
**So that** I can efficiently prioritise audited items for improvement

### Background

This has been requested by HMRC. At present they can filter the audit 
progress report (and therefore the csv export) by `Assigned to`, `Organisation(s)`, 
`Content Type`. 

This PR will allow them to download the CSV that represent the status 
of the filter including the `Audit Status`.


alecgibson:
[(3) Filter the export by audit status](https://trello.com/c/E2I3wbPe/639-3-filter-the-export-by-audit-status)

app/models/audits/report_row.rb
```diff
@@ -30,12 +30,17 @@ def precompute
         "Content type" => content_type,
         "Last major update" => last_major_update,
         "Whitehall URL" => whitehall_url,
+        "Audited" => audited
```
alecgibson: (Trailing commas help to reduce future commit noise)

app/models/audits/report_row.rb
```diff
@@ -30,12 +30,17 @@ def precompute
...
       }
 
       save!
       self
     end
 
+    def audited
+      content_item.audit.present? ? 'Audited' : 'Non audited'
```
alecgibson: "Not audited" would probably be more correct

spec/domain/audits/report_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      expect(csv.lines.first).to start_with('Report URL,Report timestamp')
+    end
+
+    it 'outputs whether the item has been audited or not' do
+      csv = Report.generate(filter, url)
+      header = csv.lines.first
+      first_row = csv.lines.third
+
+      expect(header).to match(/Audited/)
+      expect(first_row).to match(/Non audited/)
```
alecgibson: It would be nice to explicitly set this up in the `before` block.

spec/domain/audits/report_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      expect { Report.generate(filter, url) }.not_to raise_error,
+        'Should not have tried to execute a query after initializing Report'
+    end
+
+    it 'returns all items regardless of the audit status' do
+      create(:passing_audit)
+      create(:failing_audit)
+
+      expect(Report.generate(filter, url).lines.length).to eq(5)
```
alecgibson: It's unclear why I'd be expecting 5 rows here. It might be clearer to:

- Contain the item set up in the `before` block in its own separate `describe` block, because it's easy to forget that it exists
- Then write an assertion like `expect(Report.generate(filter, url).lines.length).to eq(%w[metadata_row headers passing_audit failing_audit].count)`

spec/domain/audits/report_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    it 'returns items filtered by audit status' do
```
alecgibson: As above, I don't think these assertions of row counts are particularly clear.

pmanrubia:
@alecgibson, thank you for your review; I have addressed your comments.

PR #410, opened 2017-11-20
Store Content Metrics over time

alecgibson:
This change adds a database migration and rake task that will store the metrics for a given document type for the last 3 months.

The metrics are currently limited to the title length and the reading grade, as determined by the [SMOG readability index][1].

[1]: https://en.wikipedia.org/wiki/SMOG


alecgibson:
[Spike #2: (Management - Justify work) If we store content metrics through time then we will be able to track their relationship with content performance [2-3 people, 3 days, dev and PA]](https://trello.com/c/49inwhSM/656-spike-2-management-justify-work-if-we-store-content-metrics-through-time-then-we-will-be-able-to-track-their-relationship-with-c)

app/domain/content/importers/metrics.rb
```diff
@@ -0,0 +1,117 @@
...
+          .select { |version_number| version_number < content_item[:user_facing_version] }
+          .sort
+          .last
+      end
+
+      def body(content_item)
+        return "" unless content_item.dig(:details, :parts).present?
+        content_item[:details][:parts].reduce("") do |content_item_body, part|
+          content_item_body + if part[:body].present?
+                                part[:body].reduce("") do |part_body, body_fragment|
```
KyleMacPherson: What do you think about taking the second reduce function out into a seperate method to make this easier to read/understand?

`def concat_all_parts_of_content_item_body` or something like that?

app/domain/content/importers/metrics.rb
```diff
@@ -0,0 +1,117 @@
...
+        puts "Fetching metrics for #{content_id} ..."
+        versions = fetch_versions(content_id)
+
+        metrics_by_version = {}
+
+        metrics = (@start_date..Date.today).map do |date|
+          version = versions.find { |v| Date.parse(v[:updated_at]) <= date }
+
+          if version.nil?
+            nil
```
KyleMacPherson: What is the `nil` value for here? 

If `version == nil` then `version[:title].length` on line 42 could cause us some problems right?

app/domain/content/importers/metrics.rb
```diff
@@ -0,0 +1,117 @@
...
+          else
+            metric_for_version = metrics_by_version.fetch(version[:user_facing_version]) do |version_number|
```
KyleMacPherson: Sometimes the version is not the user facing version (it is a draft?) and so we use the reading age of the user facing version. Am I understanding that correctly?

app/domain/content/importers/metrics.rb
```diff
@@ -0,0 +1,117 @@
...
+        puts "Writing metrics to database..."
+        # TODO: on_duplicate_key_update
+        Content::Metric.import(metrics)
+        puts "Imported all in #{Time.now - fetched_metrics_end}s"
+      end
+
+    private
+
+      def fetch_metrics(content_id)
+        puts "Fetching metrics for #{content_id} ..."
```
KyleMacPherson: Nice work, I always forget to do this! 👌

app/domain/content/importers/metrics.rb
alecgibson: This `if...else` block is exactly here to stop problems with `version[:title]` (which is in the `else` bit). The `nil` is compacted out on line 54.

app/domain/content/importers/metrics.rb
alecgibson: I'm pretty sure that historical versions have always been published (the only way of creating a new draft is by either publishing or discarding the current draft in progress). There might be funniness around the current version being a draft, but I'm not sure that's that important for this spike? Certainly something we'd have to consider for production code.

app/domain/content/importers/metrics.rb
KyleMacPherson: Oops, I meant line 48 `version[:content_id]` which is outside of the `else` statement. :man_facepalming:

alecgibson: Line 48 is also inside the `else` block. I admit there's an icky amount of nesting here. If this was production code I would definitely tidy it up.

KyleMacPherson: Ha oh `code reading fail`. Sorry!

alecgibson:
[Back-import content metrics from Publishing API for Guides](https://trello.com/c/wSqIkQQQ/667-back-import-content-metrics-from-publishing-api-for-guides)

alecgibson:
@KyleMacPherson @pmanrubia I've tidied up the code a bit and thrown in some more metrics - could you please have another look?

PR #412, opened 2017-11-22
Ignore drafts when fetching content items

alecgibson:
We currently have a mismatch between the Audit Tool and the live GOV.UK
site when a draft of a content item exists.

If a draft exists, the Audit Tool will import the draft, and display
that, rather than the latest published edition (which is shown on
GOV.UK).

This change adds a method to our Publishing API client to fetch the
latest published edition, or to return `nil`, so that the content we
display in the Audit Tool will match that on GOV.UK.

### GOV.UK

govuk-summary

### Audit tool - before

cpm-summary-before

### Audit tool - after

cpm-summary-after

### DIFF
--- a/app/services/clients/publishing_api.rb
+++ b/app/services/clients/publishing_api.rb
@@ -26,8 +26,18 @@ module Clients
         .flatten
     end
 
-    def fetch(content_id, locale)
-      normalise(publishing_api.get_content(content_id, locale: locale))
+    def fetch(content_id, options = {})
+      normalise(publishing_api.get_content(content_id, options))
+    end
+
+    def fetch_latest_published(content_id, locale)
+      content_item = fetch(content_id, locale: locale)
+      return content_item if content_item[:publication_state] == "published"
+
+      preceding_version = content_item[:user_facing_version] - 1
+      return nil if preceding_version < 1
+
+      fetch(content_id, locale: locale, version: preceding_version)
     end
 
     def links(content_id)
--- a/app/domain/content/importers/single_content_item.rb
+++ b/app/domain/content/importers/single_content_item.rb
@@ -13,6 +13,7 @@ module Content
 
     def run(content_id, locale)
       content_item = content_items_service.fetch(content_id, locale)
+      return unless content_item.present?
       links = content_items_service.links(content_id)
 
       set_metrics(content_item)
--- a/app/services/content/items_service.rb
+++ b/app/services/content/items_service.rb
@@ -27,7 +27,8 @@ class Content::ItemsService
       publishing_app
       locale
     ]
-    all_attributes = client.fetch(content_id, locale)
+    all_attributes = client.fetch_latest_published(content_id, locale)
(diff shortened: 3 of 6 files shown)


alecgibson:
[Audit page represents Whitehall not GOV.UK](https://trello.com/c/EZP3OS59/641-audit-page-represents-whitehall-not-govuk)

app/services/clients/publishing_api.rb
```diff
@@ -26,8 +26,18 @@ def fetch_all(fields)
         .flatten
     end
 
-    def fetch(content_id, locale)
-      normalise(publishing_api.get_content(content_id, locale: locale))
+    def fetch(content_id, options = {})
+      normalise(publishing_api.get_content(content_id, options))
+    end
+
+    def fetch_latest_published(content_id, locale)
```
pmanrubia: I think `fetch_latest_published` should be implicit within the `fetch` method. Audit Tool and Content Performance Manager only care about `published` content items.
It seems that the issue is related with having a draft version of a `ContentItem` created after a Content Item has been published. 
I am not feeling comfortable returning `nil` to express that the content item is drafted y their latest edition.
I would prefer to keep the `fetch` method as it is, and inside it, get the latest published version.

app/services/clients/publishing_api.rb
```diff
@@ -26,8 +26,18 @@ def fetch_all(fields)
...
+      content_item = fetch(content_id, locale: locale)
+      return content_item if content_item[:publication_state] == "published"
+
+      preceding_version = content_item[:user_facing_version] - 1
```
pmanrubia: Why is this condition being checked? If the version is 0, shouldn't it be a `non-published` content item. Have you found this scenario in our current data?

app/services/clients/publishing_api.rb
pmanrubia: Would you know as well, if we can have more than one draft version for a Content Item? If this is the case, the way it is implemented would not work in all situations.

alecgibson: Fine by me. I kept `fetch`, because we still need to fetch draft / published stuff. But I can hide that implementation.

You can only have [one draft](https://github.com/alphagov/publishing-api/blob/master/doc/model.md#workflow):

> At any one time a document can contain:
>
> - 1 edition in a "draft" state
> - 1 edition in a "published" or "unpublished" state
> - any number of editions in a "superseded" state

app/services/clients/publishing_api.rb
alecgibson: `user_facing_version` [is 1-based](https://github.com/alphagov/publishing-api/blob/ad08eeb43b4e92d7823017812ff7d3007bddff42/db/migrate/20161214113229_pillar_column_defaults.rb#L5). If the `preceding_version` is 0, then the content item we have is actually the first version (and therefore there is no preceding version).

alecgibson:
Superseded by https://github.com/alphagov/content-performance-manager/pull/437

PR #413, opened 2017-11-28
[DO NOT MERGE] Spike content review

chao-xian:
This is a proof of concept prototype and must not be merged into master!

Store a `review_by` date field against each content item. Expose this for other apps to consume, such as Whitehall Publisher.

Goes with this change in Whitehall: https://github.com/alphagov/whitehall/pull/3567

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,6 @@
 class ApplicationController < ActionController::Base
   include GDS::SSO::ControllerMethods
 
-  protect_from_forgery with: :exception
+  protect_from_forgery unless: -> { request.format.json? }
   before_action :authenticate_user!
 end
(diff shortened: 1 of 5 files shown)


chao-xian:
[Spike #1: (Management - Life expectancy) If we set a review date on content then we know what we need to review (and audit) at right time and feed into archive process because we can use this to alert publishers and time it with key dates (e.g budget) [2 people, 2 days, dev and competitor analysis]](https://trello.com/c/A3cMuEY6/658-spike-1-management-life-expectancy-if-we-set-a-review-date-on-content-then-we-know-what-we-need-to-review-and-audit-at-right-tim)

pmanrubia:
@chao-xian I am closing this PR, as this was never intended to be merged into `master`. Good work 👍!

PR #415, opened 2017-11-30
Full-width, two-column layout

alecgibson:
In preparation for displaying a content preview to the right-hand side of the audit page, this change rearranges the page into a full-width, two-column layout.

The two panes are independently scrollable. The right-hand pane will be hidden on mobile and tablet devices.

## Dependencies

- [x] https://github.com/alphagov/govuk_admin_template/pull/157
- [x] https://github.com/alphagov/govuk_admin_template/pull/158

## Before

before

## After

after

### DIFF
--- a/app/views/layouts/audit_wrapper.html.erb
+++ b/app/views/layouts/audit_wrapper.html.erb
@@ -1,0 +1,20 @@
+<% content_for :page_title do %>
+  Content Audit Tool
+<% end %>
+
+<% content_for :app_title do %>
+  Content Audit Tool
+<% end %>
+
+<% content_for :app_home_path do %>
+  <%= audits_path %>
+<% end %>
+
+<% content_for :head do %>
+  <%= csrf_meta_tags %>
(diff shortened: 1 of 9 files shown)


alecgibson:
[(5) Add full-width 2-column layout to the Audit Page](https://trello.com/c/1kTWYL4J/693-5-add-full-width-2-column-layout-to-the-audit-page)

app/views/audits/audits/_audit.html.erb
```diff
@@ -0,0 +1,99 @@
+<div class="audit-content-item">
```
alecgibson: NB: This entire file is a copy-paste from `show.html.erb`

app/views/layouts/audit_wrapper.html.erb
```diff
@@ -0,0 +1,20 @@
+<% content_for :page_title do %>
```
alecgibson: This file was cut-and-pasted from `audits.html.erb`

PR #417, opened 2017-12-04
Upgrade to Ruby 2.4.2 & Bundle update

danielburnley:
Ahead of enabling Dependabot on the repository, we want to be using the latest
dependancies, to minimise the onslaught of incremental PRs.

https://trello.com/c/vX8qwh2X/46-ruby-upgrade-of-all-apps


Gemfile
```diff
@@ -50,7 +50,7 @@ group :development, :test do
   gem 'capybara'
   gem 'database_cleaner'
   gem 'factory_bot_rails'
-  gem 'govuk-lint', '~> 3.2'
+  gem 'govuk-lint', '3.2.0'
```
andrewgarner: Why do we have to specify a hard dependency on govuk-lint?

danielburnley: When govuk-lint upgrades it causes linting failures, these should probably be fixed in another PR

tijmenb:
@pmanrubia will merge this once ready.

Gemfile
pmanrubia: @danielburnley, thank you for your work on this PR. 

In our team, we'd rather not constraining versions in the `Gemfile`. In case we have `Lint` failures because of an update, we would do the following:

1. Bundle update `govuk-lint`
2. Fix `Lint` issues
3. Add commit (including `Gemfile.lock`)
4. run `Bundle update`
5. Add and commit `Gemfile.lock`

I hope this is ok with you.

PR #418, opened 2017-12-04
Add page counter to Audit tabs 'All content' and 'My content'

theKHutDeveloper:
Amend the format of the page counter and make it appear at both the
top and bottom of the 'All content' and 'My content' pages.

We are use the gem Kaminari for pagination, to format the display
of the page counter we customise the Kaminari helpers on page-entries_info.

[Trello](https://trello.com/c/S4OPeGYb/676-2-add-an-page-counter-x-y-of-z)

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -39,7 +39,7 @@ en:
     page_entries_info:
       one_page:
         display_entries:
-          one: "1 - %{count} of %{count} %{entry_name}"
-          other: "1 - %{count} of %{count} %{entry_name}"
+          one: "1&nbsp;-&nbsp;%{count} of %{count} %{entry_name}"
+          other: "1&nbsp;-&nbsp;%{count} of %{count} %{entry_name}"
       more_pages:
         display_entries: "%{first}&nbsp;-&nbsp;%{last} of %{total} %{entry_name}"
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -35,3 +35,11 @@ en:
   term_generation_states:
     dont-know: "Don't know"
     not-relevant: "Not relevant to theme"
+  helpers:
+    page_entries_info:
+      one_page:
+        display_entries:
+          one: "1 - %{count} of %{count} %{entry_name}"
+          other: "1 - %{count} of %{count} %{entry_name}"
+      more_pages:
+        display_entries: "%{first}&nbsp;-&nbsp;%{last} of %{total} %{entry_name}"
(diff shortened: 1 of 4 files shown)


alecgibson:
[(2) Add an page counter (x-y of z)](https://trello.com/c/S4OPeGYb/676-2-add-an-page-counter-x-y-of-z)

config/locales/en.yml
```diff
@@ -35,3 +35,11 @@ en:
...
     dont-know: "Don't know"
     not-relevant: "Not relevant to theme"
+  helpers:
+    page_entries_info:
+      one_page:
+        display_entries:
+          one: "1 - %{count} of %{count} %{entry_name}"
+          other: "1 - %{count} of %{count} %{entry_name}"
+      more_pages:
+        display_entries: "%{first}&nbsp;-&nbsp;%{last} of %{total} %{entry_name}"
```
alecgibson: Why have we got non-breaking spaces here, but not in `one_page:display_entries`? I don't really think we even need them, do we? When do we expect this text to be anywhere near a line break?

andrewgarner: 👍 for the non breaking spaces. When we define these translation strings we shouldn't think too much about the context where they'll be used. Maybe we should adopt this approach throughout?

alecgibson: Sure I don't mind if we keep them or not, but they should definitely be used in your other string to be consistent with this one.

andrewgarner: As an aside, we should use the key `display_entries_html`. The `_html` suffix tells Rails that out translation string (containing HTML markup) is HTML safe and negates us having to explicitly use `.html_safe` in our view code.

See http://guides.rubyonrails.org/i18n.html#using-safe-html-translations

theKHutDeveloper: @alecgibson, @andrewgarner

I’ve added nbsp to one_page:display_entries but when I change the key display_entries to display_entries_html the Kaminari gem displays the default page counter e.g. ‘Displaying items 1 - 100 of 500 in total’ instead of ‘1 - 100 of 500 items’.

Looking further into the [Kaminari helper methods](https://github.com/kaminari/kaminari/blob/508009c4fdc2305fdc3a5e1afb806720fcc49a84/kaminari-core/lib/kaminari/helpers/helper_methods.rb) I am wondering if Kaminari recognises  ‘display_entries_html’?  

 I would appreciate your thoughts.

andrewgarner: @theKHutDeveloper ahh okay. Kaminari [is explicitly calling `.html_safe` in this case](https://github.com/kaminari/kaminari/blob/508009c4fdc2305fdc3a5e1afb806720fcc49a84/kaminari-core/lib/kaminari/helpers/helper_methods.rb#L113) so my tip wasn't that useful!

PR #419, opened 2017-12-05
Content preview iframe

alecgibson:
This change:

- adds a preview of the content being audited to the right of the audit page
- uses JavaScript to add a CSS spinner while the iframe is loading
- changes the iframe proxy to open all links in a new tab, removing confusion about how to navigate around in an iframe

### Side-by-side preview

preview

### Loading spinner

preview-spinner

## Compatibility

This solution has been tested on:

- Chrome v62.0.3202.94 on Mac
- Firefox v57.0.1 on Mac
- Safari v11.0.1 on Mac
- Internet Explorer 8 on PC
- Internet Explorer 9 on PC
- Internet Explorer 11 on PC
- Edge 16 on PC

Some compatibility issues exist in versions of IE9 and below.

### Flexbox

The panels' height is set using flexbox. Setting height through flexbox means that we can make the iframe take up as much screen real estate as possible (without resorting to JavaScript). A similar effect could have been achieved by using `position: fixed` and anchoring the two panes to the viewport. However, this would have removed the responsiveness of the page (see below). In order to address this, the height of the page is set to a fixed height of 500px on older browsers.

preview-ie8

### CSS animations

IE9 and below also don't support CSS animations. In these cases, the CSS spinner is simply replaced with text: "Loading..."

preview-ie8-loading

## Accessibility

The iframe is inserted at the end of the page so to not be obstructive when using a screenreader, or navigating using the keyboard. It has a title that announces its purpose to screen readers, and can be navigated using the keyboard if the user prefers to stay on the same page, rather than open the content in a new tab.

### No script

The only difference users will see with JavaScript disabled is that they will not get the loading spinner.

### Responsiveness

The side-by-side view makes no sense on a narrower viewport, so the preview is hidden completely on mobile and tablet using Bootstrap's built-in classes.

preview-tablet

### DIFF
--- a/app/views/audits/audits/_content_preview.html.erb
+++ b/app/views/audits/audits/_content_preview.html.erb
@@ -1,0 +1,15 @@
+<p class="preview-url">
+  <%= link_to("https://www.gov.uk#{content_item.base_path}",
+    "https://www.gov.uk#{content_item.base_path}",
+    target: "_blank",
+    rel: "noopener noreferrer",
+  ) %>
+</p>
+
+<%= tag(:iframe,
+  id: "content-preview",
+  title: "Content as shown on GOV.UK",
+  src: content_item.proxy_url,
+  scrolling: "yes",
+  frameborder: "no",
+) %>
--- a/app/views/audits/audits/_content_preview.html.erb
+++ b/app/views/audits/audits/_content_preview.html.erb
@@ -12,4 +12,5 @@
   src: content_item.proxy_url,
   scrolling: "yes",
   frameborder: "no",
+  class: "if-js-hide",
 ) %>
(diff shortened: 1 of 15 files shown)


alecgibson:
[(5) Preview content in the split screen with the Iframe](https://trello.com/c/OyURrlrJ/694-5-preview-content-in-the-split-screen-with-the-iframe)

app/views/audits/audits/_content_preview.html.erb
```diff
@@ -12,4 +12,5 @@
   src: content_item.proxy_url,
   scrolling: "yes",
   frameborder: "no",
+  class: "if-js-hide",
```
pmanrubia: Out of curiosity, where is this class defined? Is it part of `govuk_admin_template`?

alecgibson: [Yup](https://github.com/alphagov/govuk_admin_template/blob/340e2e341b28783d9ee44a60ce4a0cf07af81c3b/CSS.md#hide-and-show-content)

andrewgarner:
🤟🏻

PR #420, opened 2017-12-06
Topics filter

andrewgarner:
screen shot 2017-12-06 at 09 58 25


andrewgarner:
[(5) Add a 'Topics' filter](https://trello.com/c/wpzNZH8P/675-5-add-a-topics-filter)

app/helpers/dropdown_helper.rb
```diff
@@ -15,6 +15,30 @@ def organisation_options_for_select(selected = nil)
     options_for_select(organisation_options, selected)
   end
 
+  def topic_options_for_select(selected = nil)
+    topic_option_attributes = ->(topic) do
```
alecgibson: (I personally think `lambda` is less confusing syntax, more in keeping with Ruby's block syntax)

app/helpers/dropdown_helper.rb
```diff
@@ -15,6 +15,30 @@ def organisation_options_for_select(selected = nil)
...
+      [
+        *topic.links.map { |link| link.target.title.squish },
```
alecgibson: Do we only ever expect 1 and only 1 link here? If so, I'd prefer something like `first` rather than the splat operator to make that clear.

alecgibson: Better yet, could we just ask for `topic.parent` if that's what we want? It's quite hard to understand what's going on in this lambda, especially before reading the scope from below.

app/helpers/dropdown_helper.rb
```diff
@@ -15,6 +15,30 @@ def organisation_options_for_select(selected = nil)
...
+
+    topic_options = Content::Item
+                      .includes(links: %i(target))
+                      .where(
+                        document_type: 'topic',
+                        links: { link_type: Content::Link::PARENT },
+                      )
+                      .map(&topic_option_attributes)
+                      .sort_by { |parent_title, title, _| [parent_title, title] }
+                      .map { |parent_title, title, content_id|
```
alecgibson: I think we should be using `do..end` for multi-line blocks; I'm surprised our linter didn't shout at you for this?

pmanrubia: Addressed in 5e545ac
cc/ @andrewgarner

app/helpers/dropdown_helper.rb
pmanrubia: Addressed in a0665f1
cc/ @andrewgarner

app/helpers/dropdown_helper.rb
pmanrubia: Addressed in 098a0615cb1972e6c27b6bdcdc2bedf6686a8d31
cc/ @andrewgarner

pmanrubia:
I have addressed @alecgibson's comments in @andrewgarner's PR. I have not fixed up the commits as the comments were code improvements.

app/views/audits/audits/_sidebar.html.erb
```diff
@@ -1,6 +1,7 @@
 <%= form_tag audits_path, method: :get do %>
   <%= render 'audits/common/title' %>
   <%= render 'audits/common/audit_status' %>
+  <%= render 'audits/common/topics' %>
```
alecgibson: We've merged https://github.com/alphagov/content-performance-manager/pull/424#issuecomment-349951337 before this, so can you please make sure we put the "Topics" filter in the right place. According to the [Trello card](https://trello.com/c/6h4bHTO1/679-1-filter-area-for-the-my-content-tab), it needs to go after "Primary org" and before "Content type"

pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/420/commits/c38fa8e0c2db09e404bff3e0a0dc34663e4465d5

PR #421, opened 2017-12-06
Content scale and performance data report for archiving team

mgrassotti:
Trello: https://trello.com/c/eOO9oPQD/787-content-scale-and-performance-data-for-archiving-team

In order to explore options for retaining all content on GOV.UK and
automatically down weighting low value, out of date and ephemeral
content, we would like to try and get an understanding of the scale and
spread of content with no views or minimal views

This change provides the following queries results:

- Number of content items with no views (in the last 6 months)
- Number of content items with less than 10 views
- Number of content items with no views categorised by format
- Number of content items with no views that have never been updated
(in the last 6 months)
- Number of content items with no views categorised by department

It can be run from the console with `rake report:content_items`

### DIFF
--- a/lib/tasks/report/content_items.rake
+++ b/lib/tasks/report/content_items.rake
@@ -1,0 +1,31 @@
+namespace :report do
+  task content_items: :environment do
+    puts "\n--- REPORT ---"
+    puts "--- views = Six months page views ---"
+
+    # Number of content items with no views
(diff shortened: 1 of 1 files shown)


pmanrubia:
@mgrassotti Thank you for your work on this PR. 
I would prefer to track these queries in the Trello card so we can get back to hem if needed.  In case this information is requested at intervals of times, we will then prepare a way to automate them.

PR #423, opened 2017-12-06
Add select all checkbox to allocations table

alecgibson:
This change adds a checkbox to the header row of the allocations table,
which - when selected - will select all the checkboxes in the table.
It will also update the batch input value to match the number of
checked items.

The checkbox is hidden when JavaScript is disabled, and it has a label
visible only to screenreaders for additional context.

This change also defaults the value in the batch input to "0" to be
consistent with when all items are deselected, and to provide a better
prompt that the input can accept a number.

select-all


alecgibson:
[(3) Add a 'select all' option to list table header box](https://trello.com/c/a30wBTIT/677-3-add-a-select-all-option-to-list-table-header-box)

pmanrubia:
👍

PR #425, opened 2017-12-07
Change name, order and paths of audits navigation tabs

KyleMacPherson:
- Rename `Assign content` tab to `All content` to more accurately reflect the
the contents of the page and re-order the tabs, so that the `All content` tab
is first, to improve user’s navigation and understanding of the appplication (as
validated by user research anaylsis).
- Change paths to reflect the new naming of the navigation tabs.
- Alter set up of tests and make changes to the urls of page object models to
ensure that tests navigate to the intended page after the changes that have
been made.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,8 +13,10 @@ Rails.application.routes.draw do
     end
   end
 
+
   namespace :audits do
-    get '/', to: "audits#index"
+    get '/', to: "allocations#index"
+    get '/my-content', to: "audits#index", as: 'my_content'
     resource :report, only: :show
     resource :guidance, only: :show
 
--- a/app/views/audits/audits/_sidebar.html.erb
+++ b/app/views/audits/audits/_sidebar.html.erb
@@ -1,4 +1,4 @@
-<%= form_tag audits_path, method: :get do %>
+<%= form_tag audits_my_content_path, method: :get do %>
   <%= render 'audits/common/title' %>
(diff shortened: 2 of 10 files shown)


KyleMacPherson:
[(2) Switch tab order and change labels](https://trello.com/c/hKl4VnCJ/673-2-switch-tab-order-and-change-labels)

config/routes.rb
```diff
@@ -13,8 +13,11 @@
     end
   end
 
+
   namespace :audits do
-    get '/', to: "audits#index"
+    get '/', to: "allocations#index"
+    get '/my_content', to: "audits#index", as: 'my_content'
```
alecgibson: The more "URL-like" thing to do is use spinal-case rather than snake_case.

config/routes.rb
```diff
@@ -13,8 +13,11 @@
...
+    get '/all_content', to: "allocations#index", as: 'all_content'
```
alecgibson: I'm not sure about having two routes leading to `allocations#index`. We should do one of:

- remove the `/all_content` endpoint altogether
- redirect the root path: `get "/", to: redirect("/all_content")`

config/routes.rb
KyleMacPherson: Thanks 👍

config/routes.rb
KyleMacPherson: 👍

PR #422, opened 2017-12-06
Side-by-side content preview on audit pages

alecgibson:
This change merges in the changes made in these separate (feature branch) pull requests:

- https://github.com/alphagov/content-performance-manager/pull/414
- https://github.com/alphagov/content-performance-manager/pull/415
- https://github.com/alphagov/content-performance-manager/pull/419

Please see separate PRs for further context.

PR #426, opened 2017-12-07
Collapsible metadata

alecgibson:
This change hides some of the audit metadata in a collapsible section. It makes use of the existing partial used to hide some of the filters in the sidebar. The partial has been renamed and modified to fit this use case too.

## Dependencies

- [ ] https://github.com/alphagov/content-performance-manager/pull/422

expand-metadata


alecgibson:
[(3) Collapse content item metadata into "More details" section](https://trello.com/c/WmRATkuI/696-3-collapse-content-item-metadata-into-more-details-section)

alecgibson:
Review app: https://cpm-prototype-426.herokuapp.com/audits

app/views/components/_expandable_content.html.erb
```diff
@@ -2,22 +2,33 @@
   start_expanded ||= false
   open_label ||= "More details"
   close_label ||= "Fewer details"
+  button_at_top ||= false
+
+  toggle_button = content_tag(:div, class: "form-group js-expand-button-wrapper hidden") do
```
pmanrubia: This might fit better in a helper?

alecgibson: @pmanrubia I'd actually thought about just moving it into a partial. The reason I didn't was that having a partial (or even a helper) would make this feel much less like a standalone component (which is what it's trying to be).

I have no strong feelings, though, so if you'd prefer, I can move this code into a partial?

PR #429, opened 2017-12-07
Fix content preview after saving an audit

alecgibson:
After saving an audit, if there is no content item to audit next, the
audit page collapses into a single-column layout.

This is because in the `save` action we manually render the `show`
view, bypassing the layout that gets set in the `show` action.

This change fixes this by correctly setting the layout in this case.


alecgibson:
[Audit split-screen does not show iframe when completing all audits](https://trello.com/c/o4L1M7i0/700-audit-split-screen-does-not-show-iframe-when-completing-all-audits)

PR #430, opened 2017-12-07
Disallow allocation to "Anyone"

alecgibson:
This change addresses [an error](https://sentry.io/govuk/app-content-performance-manager/issues/413345465/) we've been seeing, which arises from users assigning content to "Anyone". The fix:

- adds foreign keys to the Allocations table
- adds a uniqueness constraint to users' IDs (so that we can FK to it, apart from it also just being a Good Thing)
- removes "Anyone" from the choices for allocation

### DIFF
--- a/app/helpers/dropdown_helper.rb
+++ b/app/helpers/dropdown_helper.rb
@@ -31,8 +31,15 @@ module DropdownHelper
     options_for_select(document_type_options, selected)
   end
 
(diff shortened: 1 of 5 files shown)


alecgibson:
[BUG: Audit Tool: Some content items can't be audited](https://trello.com/c/1OYQv1xE/699-bug-audit-tool-some-content-items-cant-be-audited)

PR #431, opened 2017-12-08
Add `assigned to` column to content items table

KyleMacPherson:
- Add `assigned to` column to content items table on the `all content` page to
help user easily identify which team member, if any, each task is assigned to
as validated by user research findings.
- Add tests to assert that the column is populated with either the correct user's
name or `No one` and add data-test attributes to easily and explicitly query
the necessary elements


KyleMacPherson:
[(3) Include an 'Assigned to' column on the 'All content' page](https://trello.com/c/8m3C61xF/674-3-include-an-assigned-to-column-on-the-all-content-page)

PR #432, opened 2017-12-08
Allocate from an audit page

alecgibson:
This change makes it possible to allocate content from the audit page.
Further, it actually prevents users from auditing content that is not
allocated to them.

allocation

### DIFF
--- a/app/controllers/audits/allocations_controller.rb
+++ b/app/controllers/audits/allocations_controller.rb
@@ -75,7 +75,11 @@ module Audits
     end
 
     def redirect_url
-      audits_allocations_url(redirect_params)
+      if params[:redirect_to_content_item] && content_ids.count == 1
+        content_item_audit_path(content_ids.first)
+      else
+        audits_allocations_url(redirect_params)
+      end
     end
   end
 end
--- a/app/helpers/content/items_helper.rb
+++ b/app/helpers/content/items_helper.rb
@@ -55,4 +55,8 @@ module Content::ItemsHelper
       },
     ]
   end
+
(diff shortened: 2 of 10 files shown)


alecgibson:
[(5) Add 'This item is not assigned to anyone.' bar](https://trello.com/c/89y2uTHV/684-5-add-this-item-is-not-assigned-to-anyone-bar)

alecgibson:
Review app: https://cpm-prototype-432.herokuapp.com/audits

app/controllers/audits/allocations_controller.rb
```diff
@@ -74,7 +74,11 @@ def redirect_params
     end
 
     def redirect_url
-      audits_allocations_url(redirect_params)
+      if params[:redirect_to_content_item] && content_ids.count == 1
```
pmanrubia: What is the reason behind ```content_ids.count==1```?

alecgibson: In theory to avoid us ever trying to pass `redirect_to_conent_item` `true` but also passing either no content Ids (`500` error) or multiple Ids, because that wouldn't make any sense. No strong feelings on this logic. If you prefer I can instead check `content_ids.positive?`?

alecgibson: I suppose it would be most restful to return a `400` if someone sends `redirect...` `true` and a non-unity number of Ids?

alecgibson: As discussed, we'll leave this as-is for now.

PR #434, opened 2017-12-11
Restrict iframe proxy endpoint to SAMEORIGIN frames

alecgibson:
We currently have a constraint on our iframe proxy endpoint that means
a user must be logged in to access it. However, when we deploy our app
to Heroku, we don't have Signon, and this constraint fails.

We can't just enable the endpoint on Heroku, because this would allow
arbitrary people to use our proxy in their own iframes.

This change allows us to make this endpoint work on Heroku, whilst also
preventing people from using our proxy on their own sites. We achieve
this by:

- removing the logged-in constraint
- making our own proxy return an `X-FRAME-OPTIONS SAMEORIGIN` header,
  which is exactly the thing we were trying to circumvent on GOV.UK in
  the first place, and therefore should be adequate security when
  applied here, too

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -37,16 +37,10 @@ Rails.application.routes.draw do
     mount GovukAdminTemplate::Engine, at: "/style-guide"
   end
 
-  class ProxyAccessContraint
-    def matches?(request)
-      !request.env['warden'].try(:user).nil?
-    end
-  end
-
   # rack-proxy does not work with webmock, so disable it for tests: https://github.com/ncr/rack-proxy#warning
   if Rails.env.test?
     get "#{Proxies::IframeAllowingProxy::PROXY_BASE_PATH}*base_path", to: proc { [200, {}, ['Proxy disabled in Test environment']] }
   else
-    mount Proxies::IframeAllowingProxy.new => Proxies::IframeAllowingProxy::PROXY_BASE_PATH, constraints: ProxyAccessContraint.new
+    mount Proxies::IframeAllowingProxy.new => Proxies::IframeAllowingProxy::PROXY_BASE_PATH
   end
 end
--- a/lib/proxies/iframe_allowing_proxy.rb
+++ b/lib/proxies/iframe_allowing_proxy.rb
@@ -34,7 +34,7 @@ module Proxies
         result = body
       end
 
-      [status, headers.tap { |h| h['x-frame-options'] = 'ALLOWALL' }, result]
+      [status, headers.tap { |h| h['x-frame-options'] = 'SAMEORIGIN' }, result]
     end
 
     def rewrite_relative_link!(link)


alecgibson:
@tijmenb could you please clarify why you [asked for this constraint](https://github.com/alphagov/content-performance-manager/pull/230#discussion_r126897659) in the first place?

cbaines:
> When a user is not logged in to the application through Signon, they
> receive a 404 error, which means that we don't need this extra
> precaution in our routing.

I haven't tried this out, but where do you think this behaviour is coming from? I would expect it to work, without authentication, if you remove the requirement, as you're doing in this Pull Request.

alecgibson:
So it definitely shows me the homepage if I'm signed in, and I definitely get a 404 if I'm not. Here's the URL I've tried for reference:
https://content-performance-manager.integration.publishing.service.gov.uk/iframe-proxy/

I can try to hunt down this behaviour, but the Signon documentation isn't the greatest.

I'd also like to know why we're so worried about this? If it's about people being able to phish/generally trick people, can't those people also set up their own reverse proxy? And since we've got it written in our own open-source repo, we've already told them how to do it. Is it meant to be just about making it _harder_ for people to iframe the site? Because that seems a bit silly - we should prevent people from doing it completely if we're going to do it at all?

tijmenb:
The constraint is there to prevent the proxy app from being used by unauthenticated users. At the time I wasn't sure how the proxy would work and what special powers it could have (since the requests come from inside our network). Seems like a sensible precaution.

As for the larger point of iFrames, there's https://github.com/alphagov/government-frontend/pull/392, but I wasn't able to find someone to approve it for me. You're welcome to reopen. Other applications would need the same update, since no-iframe is the Rails default.

alecgibson:
@tijmenb if I can track down the behaviour in Signon that means unauthenticated users get a 404, would it be okay to drop this constraint?

tijmenb:
@alecgibson yeah sure, but are you sure that the behaviour is there? The proxy should be available to the public without the constraint. Afaik `gds-sso` work only on controllers.

alecgibson:
@tijmenb according to [this deprecation notice](https://github.com/alphagov/gds-sso/blob/master/lib/gds-sso/controller_methods.rb#L40) in the GDS-SSO gem, Signon now automatically requires the `signin` permission as part of our OAuth. Do I need to keep digging?

alecgibson:
Also see this RFC: https://github.com/alphagov/govuk-rfcs/blob/8cbb2a0de86de02f54ae37b245e79b46ad62cb6a/rfc-078-re-architect-signin-permissions-in-signon.md

tijmenb:
I'm 99% sure that if you make a request to https://content-performance-manager.integration.publishing.service.gov.uk/iframe-proxy, Signon won't be called anywhere in the stack because the proxy is a middleware thing. 

If it were part of the controller, [:authenticate_user!](https://github.com/alphagov/content-performance-manager/blob/master/app/controllers/application_controller.rb#L5) would be called, which protects everything inheriting from `ApplicationController`.

Or am I missing something?

alecgibson:
@tijmenb I think you're right at least regarding GDS-SSO; if I `raise` in the `authenticate_user!` on my dev environment, this only gets triggered in the main app, and not in the proxy.

However, I had a quick look at the Collections Publisher example you referred to in that earlier comment, and I _also_ get a `404` if I stay logged out and visit:
https://signon.integration.publishing.service.gov.uk/sidekiq

This is definitely making me think it's to do with Signon and/or our infrastructure. Do you know who would know about this in more detail?

alecgibson:
Oh no, I'm just being an idiot. Obviously I'm getting 404s on both of those endpoints because they've got this guard I'm trying to remove! Man, it's been a long day.

alecgibson:
@tijmenb @cbaines @pmanrubia I've updated this Pull Request with a slightly modified solution. What are your thoughts now?

alecgibson:
[Make content preview iframe work on Heroku](https://trello.com/c/qJ97UF51/709-make-content-preview-iframe-work-on-heroku)

alecgibson:
You can see the result on Heroku, where the iframe works, but the `X-FRAME-OPTIONS` header means that I cannot use it as a source for a frame on my own machine:
https://cpm-prototype-434.herokuapp.com/audits

alecgibson:
As discussed, may as well leave this as `SAMEORIGIN` because it's slightly more restrictive, but still works.

PR #435, opened 2017-12-11
Delete allocations to `anyone`

pmanrubia:
As a result of a bug that is already fixed, we have invalid
allocations in our database. These allocations have the uid
of `anyone`.

The issue was [fixed in this PR][1], and there is [a constraint
in place][2] so it won't happen again.

[1]: https://github.com/alphagov/content-performance-manager/pull/430
[2]: https://github.com/alphagov/content-performance-manager/pull/430/commits/26495acbaff05d2209f2b680dcba63a813df4377

### DIFF
--- a/db/migrate/20171207160000_delete_invalid_allocations.rb
+++ b/db/migrate/20171207160000_delete_invalid_allocations.rb
@@ -1,0 +1,5 @@
+class DeleteInvalidAllocations < ActiveRecord::Migration[5.1]
+  def up
+    Audits::Allocation.where(uid: 'anyone').delete_all
+  end
+end


pmanrubia:
Yes @alecgibson, thank you 👍

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171207164824) do
+ActiveRecord::Schema.define(version: 20171211164321) do
```
alecgibson: Do we need this version update now that this migration is being inserted back in time?

PR #436, opened 2017-12-12
Fix Heroku deploy

alecgibson:
This change fixes some problems with our `heroku:deploy` task:
- fixes the "Topics" filter, which requires the "parent" links to be included
- makes sure it selects an `organisation` document when fetching organisations whose titles might be duplicated in other documents

### DIFF
--- a/lib/tasks/heroku.rake
+++ b/lib/tasks/heroku.rake
@@ -8,7 +8,7 @@ namespace :heroku do
     )
 
     ## Find the organisation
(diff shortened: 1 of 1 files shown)


alecgibson:
[Fix Heroku deploy task](https://trello.com/c/JpLe9kQJ/710-fix-heroku-deploy-task)

alecgibson:
Review app: https://cpm-prototype-436.herokuapp.com/

PR #437, opened 2017-12-13
Import content item by version number

alecgibson:
We currently import content from the Publishing API by requesting a list
of content IDs of published content, and then requesting each item one
by one by its ID and its locale.

The problem with this approach is that if a published content item also
has a draft in progress, we will be returned the draft version instead
of the published version.

This change reads the `user_facing_version` also returned in the
original request for all published content, and then requests individual
items by ID, locale and version in order to retrieve the published
edition rather than the draft.

## Before

before

## After

after


alecgibson:
[Audit page represents Whitehall not GOV.UK](https://trello.com/c/EZP3OS59/641-audit-page-represents-whitehall-not-govuk)

PR #438, opened 2017-12-14
Disable JavaScript for multiple select tests

alecgibson:
A couple of our JavaScript-enabled tests appear to fail intermittently
on Jenkins. We have tried to fix these in the past, but it seems the fix
didn't work.

This change instead disables JavaScript in these tests, which should
remove all race conditions and make the tests more reliable. The
JavaScript for these elements is already tested elsewhere in this suite,
and in the JavaScript unit tests.


alecgibson:
[Fix inconsistent tests](https://trello.com/c/i6gY2ICm/711-fix-inconsistent-tests)

PR #453, opened 2017-12-18
User parameters converted from the Filter

pmanrubia:


andrewgarner:
[Wrong number of items reported in banner when allocating more than the max. number of items](https://trello.com/c/Z78QGp5t/707-wrong-number-of-items-reported-in-banner-when-allocating-more-than-the-max-number-of-items)

PR #455, opened 2017-12-21
Percentage of word files

theKHutDeveloper:
This story adds metrics for the percentage of word files. I created a number of files similar to that for the Pdf metrics and done some refactoring. I wasn't sure how much I should refactor so would really value some feedback in that area especially.

[Trello](https://trello.com/c/Zv4mvtrO/715-3-percentage-of-word-files)


theKHutDeveloper:
[(3) Percentage of Word files](https://trello.com/c/Zv4mvtrO/715-3-percentage-of-word-files)

PR #456, opened 2017-12-21
Add date dimension

andrewgarner:
Adds a dimension table for calendar dates to allow us to easily partition future fact tables based on aspects of date. Includes a factory method to allow us to populate the dimension table given a date object from the Ruby standard library.

Once we have an ETL framework in place we can put in place the logic needed to seed the dimension table daily.


andrewgarner:
[Build date dimension](https://trello.com/c/n1ndQ2F4/723-build-date-dimension)

PR #457, opened 2017-12-22
Tracking metrics through time: Organisation dimension

pmanrubia:
[Epic - Tracking metrics through time](https://trello.com/c/XAfsYd60/718-13-track-number-of-published-documents-on-a-daily-basis)

[Trello card](https://trello.com/c/wRuqvKOx/724-2-build-organisations-dimension)

The organisations will be used to filter metrics through time, so we
need to create a dimension that allows filtering by any of their
fields.

I have used some core attributes from the organisation, and in future
iterations other attributes could be removed / added.

All the organisations are retrieved from the [SearchAPI][rummager]; 
this mechanism has proved reliable and fast. This is the reason why 
I have chosen it. There are other options, like PublishingAPI, but it is 
not that fast and provides the same information.

[rummager]: https://github.com/alphagov/rummager

### DIFF
--- a/app/models/dimensions/organisation.rb
+++ b/app/models/dimensions/organisation.rb
@@ -1,0 +1,8 @@
+class Dimensions::Organisation < ApplicationRecord
+  validates :title, presence: true
+  validates :slug, presence: true
+  validates :link, presence: true
+  validates :content_id, presence: true
+  validates :organisation_state, presence: true
+  validates :content_id, presence: true
+end
--- a/app/models/etl/organisations.rb
+++ b/app/models/etl/organisations.rb
@@ -1,0 +1,45 @@
+require 'gds_api/rummager'
+
+class ETL::Organisations
+  def process
+    raw_data = extract
+    organisations = transform(raw_data)
+    load(organisations)
+  end
+
+private
+
+  def extract
+    rummager.search_enum(
+      {
+        filter_format: 'organisation',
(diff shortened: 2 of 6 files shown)


andrewgarner:
👀

app/models/dimensions/organisation.rb
```diff
@@ -0,0 +1,8 @@
+class Dimensions::Organisation < ApplicationRecord
+  validates :title, presence: true
+  validates :slug, presence: true
+  validates :link, presence: true
+  validates :content_id, presence: true
```
andrewgarner: I think you meant to write `organisation_id` here?

app/models/dimensions/organisation.rb
```diff
@@ -0,0 +1,8 @@
...
+  validates :organisation_state, presence: true
```
andrewgarner: I think we should drop the `organisation_` prefix here. It is duplication and should be handled during our ETL process where we translate between the external domain representation and our representation here.

app/models/dimensions/organisation.rb
andrewgarner: This may also be redundant or need a better name as we already have an implicit `id` attribute. If we were to drop the `organisation_` suffix how would this value differ?

pmanrubia:
@andrewgarner cc/ @KyleMacPherson Thank you for the review

PR #458, opened 2017-12-29
Tracking metrics through time: Items dimension

pmanrubia:
[Trello card](https://trello.com/c/vBXeO3de/725-2-build-items-dimension)

The grain for our first facts table is the `content item` as all requests/queries from the business are related to retrieving information regarding the performance of content items filtered/grouped by different criteria.

For example:

- Number of content items published per organisation
- Number of content items in PDF format for an organisation
- Number of visits in the last six months for content items tagged to a particular taxon

I have decided to validate only the `content_id` because, as of today, there are content_items with no organisation returned by `rummager`; this is going to change in incoming PRs when we move to other sources of data.

#### Performance

The process takes around 45 minutes to finish; this is a very first step towards a more robust and performant solution, but it seems good enough for our first MVP.

#### ETL process

We first retrieve all the content items from SearchAPI, and then, one by one we check if there is a new version of the content item, and if so, we save it.

The natural key is all the attributes that belong to the item
dimension. We can't use the `content_id` as the primary key because we are denormalising the entity as per the `star schema` approach of the datawhare house.

### DIFF
--- a/app/models/dimensions/item.rb
+++ b/app/models/dimensions/item.rb
@@ -1,0 +1,3 @@
+class Dimensions::Item < ApplicationRecord
+  validates :content_id, presence: true
+end
--- a/app/models/etl/items.rb
+++ b/app/models/etl/items.rb
@@ -36,7 +36,7 @@ private
 
   def load(items)
     new_records = items.select(&:new_record?)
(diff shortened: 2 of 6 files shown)


andrewgarner:
👀

PR #459, opened 2018-01-02
Tracking metrics through time: Metrics facts

pmanrubia:
[Trello card](https://trello.com/c/vBXeO3de/725-5-etl-items-dimension)

It stores all metrics fact for a given day. It references the following dimensions:

- Date
- Organisations
- Content Items

It does not duplicate Metrics so the same process can be run several times
on the same day.


andrewgarner:
👀

PR #463, opened 2018-01-02
Validate uniqueness of Dimensions::Organisation

KyleMacPherson:
Ensure that each instance of Dimensions::Organisation is unique
by scoping uniqueness of one attribute against all the others. If all
of the attributes are the same as another instance of a
Dimensions::Organisation then the instance will not be considered
unique. If the combination of attributes is unique then the instance
will be considered unique.


KyleMacPherson:
[Ensure natural key uniqueness for Dimensions:Organisation](https://trello.com/c/Z7M3hVCm/731-ensure-natural-key-uniqueness-for-dimensionsorganisation)

PR #466, opened 2018-01-04
Live pages report

andrewgarner:
Adds a basic initial report showing the number of live pages per day, filterable by organisation. Initially this will be limited until we can mine for historical data.

screen shot 2018-01-04 at 14 36 42

### DIFF
--- a/app/controllers/content/live_pages_reports_controller.rb
+++ b/app/controllers/content/live_pages_reports_controller.rb
@@ -2,6 +2,13 @@ class Content::LivePagesReportsController < ApplicationController
   def show
     @live_pages = Facts::Metric
                     .by_date_name
+                    .for_organisation(content_id: filter_params[:organisation])
                     .count
+  end
+
+private
+
+  def filter_params
+    params.permit(:organisation)
   end
 end
--- a/app/controllers/content/live_pages_reports_controller.rb
+++ b/app/controllers/content/live_pages_reports_controller.rb
@@ -1,0 +1,7 @@
+class Content::LivePagesReportsController < ApplicationController
+  def show
+    @live_pages = Facts::Metric
+                    .by_date_name
+                    .count
+  end
+end
--- a/app/views/layouts/audit_wrapper.html.erb
+++ b/app/views/layouts/audit_wrapper.html.erb
@@ -10,6 +10,10 @@
   <%= audits_path %>
 <% end %>
 
+<% content_for :navbar_items do %>
+
+<% end %>
+
 <% content_for :head do %>
   <%= csrf_meta_tags %>
   <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': 'reload' %>
--- a/spec/features/performance/live_pages_report_spec.rb
+++ b/spec/features/performance/live_pages_report_spec.rb
@@ -12,6 +12,13 @@ RSpec.feature 'Live pages report', type: :feature do
     then_i_can_see_the_total_live_pages_each_day
   end
 
+  scenario 'Filtering by organisation' do
+    given_i_am_logged_in
+    and_there_is_historical_page_data
+    when_i_navigate_to_the_live_pages_report
+    then_i_can_filter_the_report_by_organisation
+  end
+
   around(:example) do |example|
     Timecop.freeze(Time.local(2018, 1, 1)) do
       example.run
@@ -82,6 +89,27 @@ RSpec.feature 'Live pages report', type: :feature do
 
       expect(summary_table.live_pages.collect(&:text))
         .to contain_exactly('14', '12', '10', '8', '6', '4', '2')
+    end
+  end
+
+  def then_i_can_filter_the_report_by_organisation
+    @live_pages_report.filter do |filter|
+      filter.organisation.select 'Ministry of Peace'
+      filter.submit.click
+    end
+
+    @live_pages_report.summary_table do |summary_table|
+      expect(summary_table.dates.collect(&:text))
+        .to contain_exactly('31 December 2017',
+                            '30 December 2017',
+                            '29 December 2017',
+                            '28 December 2017',
+                            '27 December 2017',
+                            '26 December 2017',
+                            '25 December 2017')
+
+      expect(summary_table.live_pages.collect(&:text))
+        .to contain_exactly('7', '6', '5', '4', '3', '2', '1')
     end
   end
 end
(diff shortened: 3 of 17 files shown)


andrewgarner:
[Show the number of content items publishing by an organisation on a daily basis](https://trello.com/c/9KQuJdJr/729-show-the-number-of-content-items-publishing-by-an-organisation-on-a-daily-basis)

spec/features/performance/live_pages_report_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+    then_i_can_filter_the_report_by_organisation
+  end
+
+  def given_i_am_logged_in
+    create(:user)
+  end
+
+  def and_there_is_historical_page_data
+    # Let's assume it is New Year's Day
+    Timecop.freeze(Time.local(2018, 1, 1))
```
pmanrubia: How is the time unfrozen? I would be keen [to enable `safe_mode`](https://github.com/travisjeffery/timecop#timecopsafe_mode).

app/controllers/content/live_pages_reports_controller.rb
```diff
@@ -0,0 +1,33 @@
+class Content::LivePagesReportsController < ApplicationController
+  def show
+    relation = Facts::Metric
+                 .joins(:dimensions_date)
+                 .group(:date_name)
+
+    if organisation_filter.present?
```
pmanrubia: Could you please work towards extracting this logic out of the controller? Something simple like:

```ruby
@report = Reports::LivePages.run(organisation: params[organisation))
```

app/views/layouts/audit_wrapper.html.erb
```diff
@@ -10,6 +10,10 @@
   <%= audits_path %>
 <% end %>
 
+<% content_for :navbar_items do %>
```
pmanrubia: Do we need this here?

spec/features/performance/live_pages_report_spec.rb
```diff
@@ -0,0 +1,112 @@
+RSpec.feature 'Live pages report', type: :feature do
```
pmanrubia: 👍 This is really good

app/views/layouts/audit_wrapper.html.erb
andrewgarner: Yes. It overrides the nav bar items in the base layout so they don't appear in the audit tool. It's the same approach we have taken to overriding the app name etc.

spec/features/performance/live_pages_report_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+  def given_i_am_logged_in
+    create(:user)
+  end
+
+  def and_there_is_historical_page_data
+    # Let's assume it is New Year's Day
+    Timecop.freeze(Time.local(2018, 1, 1))
+
+    # Create dimensions for the last 7 days
+    dates = (7.days.ago.to_date..Date.yesterday)
```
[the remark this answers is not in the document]
andrewgarner: I think we should seed the database with dates which would remove the need for this setup here. I imagine it wouldn't be too hard to pick an opening date, perhaps when GOV.UK launched?

spec/features/performance/live_pages_report_spec.rb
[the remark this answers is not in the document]
andrewgarner: Fixed with b97c2982714d7711168d03be1ea42f1566b88ce6

app/controllers/content/live_pages_reports_controller.rb
andrewgarner: I've moved the query logic to scopes on `Facts::Metric` so now the controller just does:

```rb
Facts::Metric
  .by_date_name
  .for_organisation(content_id: filter_params[:organisation])
  .count
```

We could go further and extract the combination of the two scopes and the call to `count` into another object but I feel like that's just adding more layers at this time.

spec/features/performance/live_pages_report_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+              .map { |date| create(:dimensions_date, date: date) }
+
+    # We have two publishing organisations
+    ministry_of_love = create(:dimensions_organisation, title: 'Ministry of Love')
+    ministry_of_peace = create(:dimensions_organisation, title: 'Ministry of Peace')
+
+    # On Monday they publish once, twice on Tuesday, three times on Wednesday etc
```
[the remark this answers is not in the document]
andrewgarner: I didn't want to go any further at this stage until we have uncovered other test scenarios when we introduce additional reports.

pmanrubia:
@andrewgarner, thank you for your work on this PR.
I am closing this PR as the UI is still to be defined in the new quarter via user research. 

> I think we should seed the database with dates which would remove the need for this setup here. I imagine it wouldn't be too hard to pick an opening date, perhaps when GOV.UK launched?

I'd rather being explicit about how we setup each test, in particular, as we are using FactoryGirl and not Rails fixtures.

I think it is better to make everything very explicit, and once we have iterated the codebase through some sprints, start making decisions about implicit code in testing.

Something along the above lines would have done the trick for me:

```ruby
dates = create(dimensions_dates, from: 7.days.ago)
```

Regarding testing, and as a note for this PR:
The feature test has too much data in its fixture and is preventing simpler ways to set up the testing. I am in favor of features with simple fixtures that just express what the report does/show with minimum data, [and to the proper testing at a unit level](https://github.com/alphagov/content-performance-manager/pull/466#discussion_r160372942). 
From a feature perspective we don't really need 7 days of data. A couple of days, with a few items each should be enough to describe the feature, and to exercise the stack. If we simplify the fixture, with not much work, we could get something like the code below, [which does not require explanation](https://github.com/alphagov/content-performance-manager/pull/466/commits#r159841749) (we could get rid of the comment describing how to build the fixture):

```ruby
ministry_of_love = create(:organisation, title: 'Ministry of Love')
ministry_of_peace = create(:organisation, title: 'Ministry of Peace')

create_list :metric, 2, organisation: ministry_of_love, day: '2018-1-2'
create_list :metric, 3, organisation: ministry_of_love, day: '2018-1-3'

create_list :metric, 4, organisation: ministry_of_peace, day: '2018-1-2'
create_list :metric, 5, organisation: ministry_of_peace, day: '2018-1-3'
```
There is still a lot of improvement to set up our feature tests on a more expressive way, but this would be a start that is more aligned with the final picture of what we all want.

PR #468, opened 2018-01-18
Remove the Term Generation Tool from the CPM code base.

KyleMacPherson:
We may want to make significant changes to the code in the CPM. If there is legacy code which
needs to be maintained, this will add burden to the team and constrain our ability to iterate the code.
To counteract this we remove code related to the Term Generation Tool that is not necessary for the
CPM to function as expected.

The Taxonomy Team have indicated that they are no longer using the Term Generation
Tool.

We drop the tables in such a way that they can be recreated should the migrations need to be
rolled back.


KyleMacPherson:
[(2) Split Term Generation Tool](https://trello.com/c/l7fn1C1P/20-2-split-term-generation-tool)

PR #469, opened 2018-01-18
Migrate documentation about architectural decisions to GOV.UK ADR format

pmanrubia:
GOV.UK is following a specific format to represent [ADRs][1], hence I
am migrating to the new format the ADR to document architectural
decisions.

This PR does not change substantially the content, it only updates the format

[1]: http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions


doc/arch/adr-000-document-architectural-decisions.md
```diff
@@ -0,0 +1,21 @@
+# ADR 1: Document architectural decisions
```
chao-xian: The file name is `adr-000` but you titled this `ADR 1`. Also the other ADRs have leading `00`s as in `ADR 002:`

PR #473, opened 2018-01-22
Remove audit tool code from cpm

KyleMacPherson:
Remove the Audit Tool code from the Content Performance Manager so that the devs on the team are able to iterate the codebase for the Content Performance Manager without the burden of maintaining the legacy code for the Audit Tool.

PR #476, opened 2018-01-29
[DO NOT MERGE] Google analytics metrics

andrewgarner:
Adds pageview and unique pageview metrics to the metrics fact table.

**This is branched off `remove-audit-tool-code-from-cpm` . Will need to rebase and change the PR merge target once #473 is finalised.**

### DIFF
--- a/app/models/etl/metrics.rb
+++ b/app/models/etl/metrics.rb
@@ -14,21 +14,31 @@ private
   def create_metrics
     Facts::Metric.where(dimensions_date: date).delete_all
 
-    Dimensions::Item.where(latest: true).find_in_batches(batch_size: 50000) do |batch|
-      values = batch.pluck(:id, :organisation_id)
-      metrics = values.map do |value|
+    Dimensions::Item.select(:id, :organisation_id, :link).where(latest: true).find_in_batches(batch_size: 500) do |batch|
+      pageviews = google_analytics_service
+                    .pageviews_for_date(date.date, *batch.pluck(:link))
+                    .each_with_object({}) { |i, h| h[i[:base_path]] = i.slice(:page_views, :unique_page_views) }
+
+      metrics = batch.map do |item_dimension|
         {
           dimensions_date_id: date.date,
-          dimensions_item_id: value[0],
-          dimensions_organisation_id: dimension_organisation(value[1], organisations).try(:id)
+          dimensions_item_id: item_dimension.id,
+          dimensions_organisation_id: dimension_organisation(item_dimension.organisation_id, organisations).try(:id),
+          pageviews: pageviews.dig(item_dimension.link, :page_views),
+          unique_pageviews: pageviews.dig(item_dimension.link, :unique_page_views),
         }
       end
+
       Facts::Metric.import(metrics, validate: false)
     end
   end
 
   def date
     @date ||= ETL::Dates.process
+  end
+
+  def google_analytics_service
+    @google_analytics_service ||= GoogleAnalyticsService.new
   end
 
   def organisations
--- a/app/models/facts/metric.rb
+++ b/app/models/facts/metric.rb
@@ -5,6 +5,9 @@ class Facts::Metric < ApplicationRecord
 
   validates :dimensions_date, presence: true
   validates :dimensions_item, presence: true
-  validates :pageviews, numericality: { only_integer: true }
-  validates :unique_pageviews, numericality: { only_integer: true }
+
+  with_options numericality: { only_integer: true, allow_nil: true } do
+    validates :pageviews
+    validates :unique_pageviews
+  end
 end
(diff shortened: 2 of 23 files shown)


andrewgarner:
[[3] GA: store Google Analytics metrics by day](https://trello.com/c/KwRlmarN/29-3-ga-store-google-analytics-metrics-by-day)

app/models/facts/metric.rb
```diff
@@ -5,4 +5,6 @@ class Facts::Metric < ApplicationRecord
 
   validates :dimensions_date, presence: true
   validates :dimensions_item, presence: true
+  validates :pageviews, numericality: { only_integer: true }
```
pmanrubia: The validation for `only_integer: true` does not allow `nil` values. 
We should be able to store metrics facts without being explicit about the GA metrics. What about adding `allow_nil: true` ?

app/services/google_analytics/requests/page_views_request.rb
```diff
@@ -11,7 +11,7 @@ def build(args = {})
         GetReportsRequest.new.tap do |reports|
           reports.report_requests = Array.new.push(
             ReportRequest.new.tap do |request|
-              request.metrics = metrics(unique_page_views)
+              request.metrics = (metrics(pageviews) + metrics(unique_page_views))
```
pmanrubia: Minor: what about performing the calculation with?
```ruby
metrics(pageviews, unique_page_views)
```

spec/integration/publishing_spec.rb
```diff
@@ -12,9 +12,9 @@
 
     date = Dimensions::Date.build(Date.today)
 
-    Facts::Metric.create(dimensions_organisation: org1, dimensions_item: item1, dimensions_date: date)
-    Facts::Metric.create(dimensions_organisation: org1, dimensions_item: item2, dimensions_date: date)
-    Facts::Metric.create(dimensions_organisation: org2, dimensions_item: item3, dimensions_date: date)
+    Facts::Metric.create(dimensions_organisation: org1, dimensions_item: item1, dimensions_date: date, pageviews: 0, unique_pageviews: 0)
```
pmanrubia: If `pageviews` and `unique_pageviews` are not mandatory, then we just can ignore them here. Using `FactoryGirl`, and defaulting metrics, could also do the trick.

app/models/etl/metrics.rb
```diff
@@ -14,15 +14,21 @@ def process
   def create_metrics
     Facts::Metric.where(dimensions_date: date).delete_all
 
-    Dimensions::Item.where(latest: true).find_in_batches(batch_size: 50000) do |batch|
-      values = batch.pluck(:id, :organisation_id)
-      metrics = values.map do |value|
+    Dimensions::Item.select(:id, :organisation_id, :link).where(latest: true).find_in_batches(batch_size: 500) do |batch|
```
pmanrubia: I think we should have an independent `ETL::GAMetrics` that populates the Facts table with GA Metrics and has its own trigger (`rake` task). It should happen the same with Content Quality Metrics, as both of them should be able to run independently.

If GA is down, or having problems, or if we want to run the process for a particular day, we don't have to recreate the whole Facts Table for that day. In particular, when we are deleting the table first.

pmanrubia: To make it more robust, and regardless of having it split to its own process, I would create Jobs so we guarantee that a failure in one request does not prevent the whole process from working as expected.

app/services/google_analytics_service.rb
```diff
@@ -3,7 +3,19 @@ def client
...
-  def page_views(base_paths)
+  def pageviews_for_date(date, *base_paths)
+    formatted_date = date.strftime("%Y-%m-%d")
+
+    request = GoogleAnalytics::Requests::PageViewsRequest.new.build(
+      base_paths: base_paths,
+      start_dates: [formatted_date],
+      end_date: formatted_date
+    )
+    response = client.batch_get_reports(request)
```
pmanrubia: We are using a `batch_request` but we are not getting the most of it as we are only passing 1 Report Request. We are only retrieving data for 1 day, so we should be able to include until 5 reports with hundreds/thousands of paths in each request.

pmanrubia: We are also impliciting batching to 500 paths per request, which is not much considering that we only request data for 1 day.

spec/integration/publishing_spec.rb
andrewgarner: Fixed with eaa10d4de322c29e1eee762c867989a8558b1a07

app/models/facts/metric.rb
andrewgarner: Fixed with eaa10d4de322c29e1eee762c867989a8558b1a07

andrewgarner:
I've extracted the schema and model changes into [their own PR](https://github.com/alphagov/content-performance-manager/pull/478) whilst I address the ETL strategy.

andrewgarner:
Superseded by #481

PR #480, opened 2018-02-02
Preload content item dimension

KyleMacPherson:
Load content item dimension data from publishing api

Retrieve content items, with their content_id and base_path, from the publishing api in preparation for loading content item data from the content store using content_id.
Remove dimensions_organisations table and fields from the dimensions_items table that are unnecessary for this iteration.
Add publisher api test helper methods to items_spec.rb to elegantly stub requests made to the publishing api


andrewgarner:
We shouldn't commit the `.DS_Store` file. You can configure your local git to ignore them – https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

app/services/content/items_service.rb
```diff
@@ -15,6 +15,16 @@ def fetch_all_with_default_locale_only
       end
   end
 
+  def fetch_all_with_id_and_path
```
andrewgarner: The name of this method obfuscates behaviour that it implements. The `with_id_and_path` suffix suggests it will return a collection with `id` and `path` attributes where in fact it returns `content_id`, `base_path`, `title`,  and `description`. We only need `content_id` and `base_path`.

The previous method in the file has a `with_default_locale_only` suffix which reveals that we'll return the English version of the content item, or if that doesn't exist the first translation. We're not revealing that here. More importantly though, we don't want that behaviour, we only want to return content items that have an English locale.

db/schema.rb
```diff
@@ -87,48 +87,24 @@
...
     t.string "content_id"
-    t.string "title"
-    t.string "link"
-    t.string "description"
-    t.string "organisation_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.boolean "latest"
-    t.index ["content_id", "link", "organisation_id"], name: "dimensions_items_natural_key"
+    t.string "path"
```
andrewgarner: In the parlance of content items we should say "base path" not "path".

For example, the content item for https://www.gov.uk/income-tax has a base path of `/income-tax`. However https://www.gov.uk/income-tax/check-youre-paying-the-right-amount originates from the same content item and also therefore also haas a base path of `/income-tax.

andrewgarner:
I'm not sure why we are dropping characteristics of the content item dimension and the organisation dimension completely.

If my understanding of https://trello.com/c/APd2Zpqj is correct we've decided that we will mine the Publishing API for the content id and base path of all known content items. We'll then subsequently cross reference the content store for additional metadata to build a complete content item dimension table.

It would seem that all we need to do here is remove any restrictions that prevent us from creating a content item dimension with only the content id and base path.

db/migrate/20180131112642_change_dimensions_items_fields.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeDimensionsItemsFields < ActiveRecord::Migration[5.1]
```
andrewgarner: This migration suggests it is modifying the `dimensions_items` table however it changes `dimensions_items_temps` and is identical to `ChangeDimensionsItemsTempsFields`. It looks like some copy and pasting has gone wrong!

KyleMacPherson:
As discussed with @andrewgarner on slack, going to look at removing the dimensions_organisations in a separate PR 👌

app/models/etl/items.rb
```diff
@@ -1,95 +1,34 @@
...
   def extract
-    rummager.search_enum(
-      {
-        fields: 'content_id,title,link,description,organisations',
-      },
-      page_size: 1000,
-      additional_headers: {},
-    )
+    fields = %w[content_id base_path title description]
+    Content::ItemsService.new.fetch_all_with_default_locale_only(fields)
```
andrewgarner: This will also import non-english content items which is not the desired behaviour.

db/schema.rb
```diff
@@ -87,48 +87,32 @@
...
     t.string "content_id"
-    t.string "title"
-    t.string "link"
-    t.string "description"
-    t.string "organisation_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.boolean "latest"
-    t.index ["content_id", "link", "organisation_id"], name: "dimensions_items_natural_key"
+    t.string "organisation_id"
```
andrewgarner: It looks you need to regenerate the schema. There should be no `organisation_id` after the migrations have run. `title` should remain.

db/schema.rb
```diff
@@ -87,48 +87,32 @@
...
+    t.string "description"
+    t.string "link"
```
andrewgarner: Should there be a `link` as well as `base_path`?

spec/factories.rb
```diff
@@ -119,21 +119,7 @@
...
-    sequence(:description) { |i| "description - #{i}" }
-    sequence(:link) { |i| "link - #{i}" }
-    sequence(:organisation_id) { |i| "organisation_id - #{i}" }
-    sequence(:state) { |i| "state - #{i}" }
-    sequence(:content_id) { |i| "content_id - #{i}" }
-  end
-
   factory :dimensions_item, class: Dimensions::Item do
     sequence(:content_id) { |i| "content_id - #{i}" }
-    sequence(:title) { |i| "title - #{i}" }
```
andrewgarner: We should keep this now.

spec/factories.rb
```diff
@@ -119,21 +119,7 @@
...
-    sequence(:link) { |i| "link - #{i}" }
```
andrewgarner: We should keep this now and update it to `base_path`.

spec/factories.rb
```diff
@@ -119,21 +119,7 @@
...
-    sequence(:description) { |i| "description - #{i}" }
```
andrewgarner: We should keep this now.

app/models/etl/items.rb
pmanrubia: This is a good spot Andrew; we will address this in a separate PR.

db/migrate/20180131114516_change_dimensions_items_temps_fields.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeDimensionsItemsTempsFields < ActiveRecord::Migration[5.1]
+  def change
+    rename_column :dimensions_items_temps, :link, :base_path
```
andrewgarner: You're no longer using the `dimensions_items_temps` table so it should be dropped. (See the deletion of `TemporaryItemStore`.

PR #481, opened 2018-02-06
ETL for Google Analytics report data

andrewgarner:
This implements an ETL pipeline for Google Analytics report data. The pipeline will populate the metrics fact table with pageview and unique pageview metrics for a given day.

The Google Analytics Reporting API restricts us to 10,000 requests per day. Each request can return a maximum of 10,000 data points.

Only a subset of the GOV.UK content is actually viewed everyday by someone, so fetching all the metrics for a given day is more economical than making more computationally expensive queries to the API with constraints for know content item dimensions. At the time of writing we should be able to import a days worth of data by utilising ~15 requests.

A "source" class returns an enumerable collection that can be traversed, searched, filtered etc as part of a pipelines of steps.

A "transform" class filters out any rows where we don't have a latest content item dimension matching the page path and returns a set of attributes suitable for importing into the fact table.

A "destination" class updates the fact table.

Closes #476

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,6 +57,7 @@ group :development, :test do
   gem 'pry-rails'
   gem 'rails-controller-testing'
   gem 'rspec-rails'
+  gem 'rspec-sidekiq'
   gem 'shoulda-matchers'
   gem 'simplecov', require: false
   gem 'site_prism'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -373,6 +373,9 @@ GEM
       rspec-expectations (~> 3.7.0)
       rspec-mocks (~> 3.7.0)
       rspec-support (~> 3.7.0)
+    rspec-sidekiq (3.0.3)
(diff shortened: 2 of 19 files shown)


andrewgarner:
[(3) GA: store Google Analytics metrics by day](https://trello.com/c/KwRlmarN/29-3-ga-store-google-analytics-metrics-by-day)

pmanrubia:
Thank you for your work on this PR. 

We are superseding this PR in favor of #486 because it is processing the metrics individually so we have a very long transaction with more than 800K queries a BBDD. I have run it a few times and I have had some timeouts so I am not confident that this is the best approach. I have cherry-picked the following commits: 074e3f3, 2aa6d7c, a72bb73, c5219b1 for #486

In #486 I have tried to use a single update after have copied to a local table all the events. This will also allow to process all the base_paths that do not match the path for a content item directly. We will use the Content Store to resolve each individual path. 

Using a custom SQL we are able to load the fact table in 7-10 seconds.

PR #482, opened 2018-02-08
Store json in content

peteglondon:
Work complete for [Content Items: Store JSON in the Content Item](https://trello.com/c/oQhjeCza/40-3-content-items-store-json-in-the-content-item).

### DIFF
--- a/app/domain/importers/content_details.rb
+++ b/app/domain/importers/content_details.rb
@@ -1,0 +1,15 @@
+class Importers::ContentDetails
+  attr_reader :items_service, :content_id, :base_path
+
+  def initialize(content_id, base_path)
+    @content_id = content_id
+    @base_path = base_path
+    @items_service = ItemsService.new
+  end
+
+  def run
+    response = items_service.fetch_raw_json(base_path)
+    item = Dimensions::Item.find_by(content_id: content_id, latest: true)
+    item.update_attributes(raw_json: response)
+  end
+end
--- a/app/models/etl/items.rb
+++ b/app/models/etl/items.rb
@@ -30,5 +30,12 @@ private
 
   def load(items)
     Dimensions::Item.import(items, batch_size: 5000)
+    create_import_detail_job(items)
+  end
+
+  def create_import_detail_job(items)
+    items.each do |item|
+      ImportItemJob.perform_async(item[:content_id], item[:base_path])
+    end
   end
 end
--- a/app/services/items_service.rb
+++ b/app/services/items_service.rb
@@ -1,18 +1,24 @@
+require 'gds_api/content_store'
 class ItemsService
-  attr_accessor :client
+  attr_accessor :publishing_api_client, :content_store_client
 
   def initialize
-    self.client = Clients::PublishingAPI.new
+    self.publishing_api_client = Clients::PublishingAPI.new
+    self.content_store_client = GdsApi::ContentStore.new(Plek.new.find('content-store'))
   end
 
   def fetch_all_with_default_locale_only(fields)
-    client
+    publishing_api_client
       .fetch_all(fields)
       .group_by { |content_item| content_item[:content_id] }
       .values
       .map do |content_items_with_the_same_id|
         content_item_with_en_locale_or_first_other(content_items_with_the_same_id)
       end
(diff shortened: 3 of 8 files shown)


app/services/content/item_details_service.rb
```diff
@@ -0,0 +1,13 @@
+require 'gds_api/content_store'
+
+class Content::ItemDetailsService
```
pmanrubia: `ItemDetails` doesn't feel like what this object does. 
I would add another method to the `ItemsService` to retrieve the `raw_json` from the `ContentStore`

spec/domain/content/importers/item_details_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    context 'Import content details' do
+      let(:base_path) { '/base_path' }
+      let(:dimension_item) { create(:dimensions_item) }
+      let(:content_store_response) { content_item_for_base_path(dimension_item.base_path) }
+
+      it 'populates content json' do
+        allow(subject.item_details_service).to receive(:fetch).and_return(content_store_response)
+        subject.run
+        expect(dimension_item.reload.content_json).to eq content_store_response
+      end
```
pmanrubia: We should also test that we are filtering by the `latest` attribute.

spec/factories.rb
```diff
@@ -130,6 +130,7 @@
     sequence(:title) { |i| "title - #{i}" }
     sequence(:base_path) { |i| "link - #{i}" }
     sequence(:description) { |i| "description - #{i}" }
+    content_json nil
```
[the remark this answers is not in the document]
peteglondon: I'm not sure this field needs to be populated by the factory at all. We want it to be nil until is has been populated in the spec we are writing. I just put in in there so it would be explicit but now you've pointed it out, it would be better to be explicit in the spec itself.

app/domain/content/importers/content_details.rb
```diff
@@ -0,0 +1,17 @@
...
+    attr_reader :items_service, :content_id, :base_path
+
+    def initialize(content_id, base_path)
+      @content_id = content_id
+      @base_path = base_path
+      @items_service = ItemsService.new
+    end
+
+    def run
+      response = items_service.fetch_raw_json_from_content_store(base_path)
```
pmanrubia: The method name should not expose where the date is coming from:
What about renaming it to: ```items_service.get_raw_json(base_path)```

db/migrate/20180208170522_rename_dimensions_items_content_json_to_raw_json.rb
```diff
@@ -0,0 +1,5 @@
+class RenameDimensionsItemsContentJsonToRawJson < ActiveRecord::Migration[5.1]
+  def change
+    rename_column :dimensions_items, :content_json, :raw_json
```
pmanrubia: Instead of adding another migration in the same PR to rename a previous column name, it feels better to just rename the field in the first commit.
What about fixing up all commits into a single commit?

app/jobs/content/import_item_detail_job.rb
```diff
@@ -0,0 +1,9 @@
+module Content
```
pmanrubia: We don't need to scope any more, as we do not have the Audit Tool has been moved out of the CPM

app/domain/content/importers/content_details.rb
```diff
@@ -0,0 +1,17 @@
+module Content
```
pmanrubia: We don't need to scope any more, as we do not have the Audit Tool has been moved out of the CPM

spec/domain/content/importers/item_details_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'gds_api/test_helpers/content_store'
+
+module Content
+  RSpec.describe Importers::ContentDetails do
```
pmanrubia: The described class does not exist.

spec/factories.rb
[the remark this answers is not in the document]
pmanrubia: Yes, this is a good point Pete. Better to be explicit in the spec

PR #484, opened 2018-02-09
Update govuk_sidekiq

cbaines:
The main motivation for this is to get the Content Performance Manager
Sidekiq metrics in to the newer machine specific namespace in
Graphite.


pmanrubia:
👍 Thank you

PR #486, opened 2018-02-12
Import Google Analytics metrics in the Facts Table

pmanrubia:
Superseds #476 

[Trello card ](https://trello.com/c/KwRlmarN/29-3-ga-store-google-analytics-metrics-by-day)

#### User need

**As a** Developer 
**I want** track Google Analytics metrics on a daily basis over time
**So that** I can expose this information via an API endpoint

#### Process

1. Retrieve from Google Analytics the `pagePaths` with the number of `pageviews` and `unique_pageviews` for a given day
2. Store the list of metrics in the table `events_ga`. This will allow us to have a better control over the loading process
3. With a single update we update the `facts_metrics` table with all the events that have a matching `pagePath` equals to the `base_path` of the content item.

This is the SQL

```SQL
  UPDATE facts_metrics
  SET unique_pageviews = s.unique_pageviews,
      pageviews = s.pageviews
  FROM (
    SELECT pageviews,
           unique_pageviews,
           dimensions_items.id
    FROM events_gas, dimensions_items
    WHERE page_path = base_path AND latest = 'true'
  ) AS s
  WHERE dimensions_item_id = s.id AND dimensions_date_id = '#{date_to_s}'
```

4. We delete the events with a matching `pagePath` 

```SQL
  DELETE FROM events_gas
  WHERE date = '#{date_to_s}' AND
    page_path in (
       SELECT base_path
       FROM dimensions_items
       WHERE latest = 'true'
    )
```


db/migrate/20180210123458_create_events_gas.rb
```diff
@@ -0,0 +1,13 @@
+class CreateEventsGas < ActiveRecord::Migration[5.1]
+  def change
+    create_table :events_gas do |t|
```
KyleMacPherson: It is not immediately obvious that `events_gas` is related to google analytics imo. I think `ga_events` or maybe even 'google_analytics_events` would be more obvious and easier for new developers to the project to grok

pmanrubia: @KyleMacPherson it was a legitimate concern, thank you for raising it. 
I think @chao-xian explained really well the background for the naming 👍

pmanrubia:
@chao-xian (cc/ @KyleMacPherson) I have addressed your comment. 
Thank you for your review.

PR #490, opened 2018-02-13
Store content v2

theKHutDeveloper:
[Trello card: Storing the content in the data warehouse over time](https://trello.com/c/qfMM7eeb/73-3-storing-the-content-in-the-data-warehouse-over-time)

The are a number of different Content Schema Types and the page content for each of these types are stored in different ways. Around a third of these Content Schema Types store the page contents in the JSON details.body element. The is the focus of this pull request. 

Based on the work done on [Trello card:  Spike: Extract content from Content Item's JSON](https://trello.com/c/CkT8R7Mg/31-3-spike-extract-content-from-content-items-json) a list of the content schema types that store the page contents in details.body was generated. This is represented as a constant array (VALID_SCHEMA_TYPES) in app/models/dimensions/item.rb

The page contents are extracted from the Content Item’s JSON if the schema_name is in the VALID_SCHEMA_TYPES array.

### DIFF
--- a/app/models/dimensions/item.rb
+++ b/app/models/dimensions/item.rb
@@ -1,3 +1,55 @@
+require 'json'
+
 class Dimensions::Item < ApplicationRecord
   validates :content_id, presence: true
+
+  def get_content
+    json_object = JSON.parse(self.raw_json)
+    return if json_object.blank?
+    extract_by_schema_type(json_object)
+  end
+
+private
+
+  VALID_SCHEMA_TYPES = %w[
+    answer
+    case_study
+    consultation
+    corporate_information_page
+    detailed_guide
+    document_collection
+    fatality_notice
+    help
+    hmrc_manual_section
+    html_publication
+    manual
+    manual_section
+    news_article
+    publication
+    service_manual_guide
+    simple_smart_answer
+    specialist_document
+    speech
+    statistical_data_set
+    take_part
+    topical_event_about_page
+    working_group
+    world_location_news_article
+  ].freeze
+
+  def extract_by_schema_type(json)
+    schema = json.dig("schema_name")
+
+    if schema.nil? || !VALID_SCHEMA_TYPES.include?(schema)
+      raise InvalidSchemaError, "Schema does not exist"
+    end
+
+    html = json.dig("details", "body")
+    return if html.nil?
+    html.delete!("\n")
+    Nokogiri::HTML.parse(html).text
+  end
+end
+
+class InvalidSchemaError < StandardError;
 end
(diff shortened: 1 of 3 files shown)


theKHutDeveloper:
[(3) Storing the content in the data warehouse over time](https://trello.com/c/qfMM7eeb/73-3-storing-the-content-in-the-data-warehouse-over-time)

app/models/dimensions/item.rb
```diff
@@ -1,3 +1,57 @@
...
+    statistical_data_set
+    take_part
+    topical_event_about_page
+    working_group
+    world_location_news_article
+  ].freeze
+
+  def extract_by_schema_type(json)
+    schema = json.schema_name
+    return if schema.nil?
```
pmanrubia: What about following the same approach of raising an exception is Schema is nil? In theory, all content items should have an Schema.

spec/models/dimensions/item_spec.rb
```diff
@@ -2,4 +2,41 @@
...
+
+  describe "#get_content" do
+    it "returns nil if json is empty" do
+      item = create(:dimensions_item, raw_json: {}.to_json)
+      expect(item.get_content).to eq(nil)
+    end
+
+    it "returns content json if it is a valid schema and details.body exist" do
+      item = create(:dimensions_item, raw_json: test_json)
+      expect(item.get_content).to eq(expected_result)
```
pmanrubia: `expected_result` is only used in this test. It might be better to define it within the test.

What about adding using a fixture that has `html` and `\n`. This way you would increase the coverage of your test.

spec/models/dimensions/item_spec.rb
```diff
@@ -2,4 +2,41 @@
...
+    "This is a test"
+  end
+
+  describe "#get_content" do
+    it "returns nil if json is empty" do
+      item = create(:dimensions_item, raw_json: {}.to_json)
+      expect(item.get_content).to eq(nil)
+    end
+
+    it "returns content json if it is a valid schema and details.body exist" do
```
pmanrubia: I would create a `context` to consider both scenarios: when valid schema, and non valid schema

Something like:

```ruby
context 'when valid schema' do
end

context 'when invalid schema' do
end
```

This will remove from your test description the `and` clause.

spec/models/dimensions/item_spec.rb
```diff
@@ -2,4 +2,41 @@
...
+      item = create(:dimensions_item, raw_json: test_json)
+      expect(item.get_content).to eq(expected_result)
+    end
+
+    it "returns nil if json if it is a valid schema but details.body does NOT exist" do
```
pmanrubia: Could you please review the copy of `returns nil if json if it is...`

theKHutDeveloper: Thanks for the feedback. I will make these changes @pmanrubia

PR #503, opened 2018-02-16
Number of pdfs

KyleMacPherson:
Populate number_of_pdfs from the content_store api response at the same stage
at which we populate the raw json field for dimensions_items.

Add a `run_for_api_response` method as to metrics/number_of_pdfs.rb as we need
to keep the previous `run` method for legacy cpm.

### DIFF
--- a/app/domain/importers/content_details.rb
+++ b/app/domain/importers/content_details.rb
@@ -13,7 +13,8 @@ class Importers::ContentDetails
 
   def run
     response = items_service.fetch_raw_json(base_path)
+    number_of_pdfs = Performance::Metrics::NumberOfPdfs.parse(response.to_h['details'])
     item = Dimensions::Item.find_by(content_id: content_id, latest: true)
-    item.update_attributes(raw_json: response)
+    item.update_attributes(raw_json: response, number_of_pdfs: number_of_pdfs)
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180210231645) do
+ActiveRecord::Schema.define(version: 20180214145821) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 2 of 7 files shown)


db/schema.rb
```diff
@@ -93,9 +93,6 @@
     t.datetime "updated_at", null: false
     t.integer "pageviews", default: 0
     t.integer "unique_pageviews", default: 0
-    t.index ["dimensions_date_id", "dimensions_item_id"], name: "index_facts_metrics_unique", unique: true
```
pmanrubia: Not sure why this indexes are being deleted

app/domain/importers/content_details.rb
```diff
@@ -13,7 +13,8 @@ def initialize(content_id, base_path)
 
   def run
     response = items_service.fetch_raw_json(base_path)
+    number_of_pdfs = Performance::Metrics::NumberOfPdfs.run_for_api_response(response)
```
pmanrubia: I'd rather not to have a method specific for a response type.  What about passing only the details?

db/schema.rb
```diff
@@ -93,6 +94,9 @@
     t.datetime "updated_at", null: false
     t.integer "pageviews", default: 0
     t.integer "unique_pageviews", default: 0
+    t.index ["dimensions_date_id", "dimensions_item_id"], name: "index_facts_metrics_unique", unique: true
```
pmanrubia: This seems to be related to previous commit (see comments)

spec/domain/performance/metrics/number_of_pdfs_spec.rb
```diff
@@ -1,5 +1,7 @@
+require 'gds_api/test_helpers/content_store'
 module Performance
   RSpec.describe Metrics::NumberOfPdfs do
+    include GdsApi::TestHelpers::ContentStore
```
pmanrubia: I would not couple this to the Content Store.

pmanrubia:
This was paired with @KyleMacPherson, we think this is good enough to be merged, hence I am approving and merging.

KyleMacPherson:
[(2) Store the Number of PDFs into the data warehouse](https://trello.com/c/5a7AHGOb/106-2-store-the-number-of-pdfs-into-the-data-warehouse)

PR #514, opened 2018-02-20
Update the Content Items Dimension On Publishing-api Events.

peteglondon:
This is needed for the [Update the Content Items dimension on publishing-api events](https://trello.com/c/sxiYIvsf/72-3-update-the-content-items-dimension-on-publishing-api-events-depends-on-94) card.
Set dirty flag in Dimensions::Item on publishing api event when the item exists.
Create a new Dimensions::Item with dirty and latest flags set to true if it does not exist.

Duplicate content items dimension and fire Sidekiq job for items marked as dirty when
ETL::Master#process is run.

Does not handle unpublish events.

### DIFF
--- a/app/models/dimensions/item.rb
+++ b/app/models/dimensions/item.rb
@@ -24,6 +24,10 @@ class Dimensions::Item < ApplicationRecord
     update_attributes!(dirty: true)
   end
 
+  def gone!
+    update_attributes(status: 'gone')
+  end
+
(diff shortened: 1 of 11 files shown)


peteglondon:
[(3) Update the Content Items dimension on publishing-api events [Depends on 94]](https://trello.com/c/sxiYIvsf/72-3-update-the-content-items-dimension-on-publishing-api-events-depends-on-94)

PR #515, opened 2018-02-21
Number of word files

KyleMacPherson:
Populate number_of_word_files from the content_store api response at the same stage
at which we populate the raw json field for dimensions_items.

Very similar to this PR: https://github.com/alphagov/content-performance-manager/pull/503

### DIFF
--- a/spec/domain/performance/metrics/number_of_word_files_spec.rb
+++ b/spec/domain/performance/metrics/number_of_word_files_spec.rb
@@ -25,20 +25,48 @@ module Performance
     let(:content_without_document_keys) { build(:content_item, details: {}) }
     let(:content_without_details_key) { build(:content_item) }
 
-    it "returns the number of word files present" do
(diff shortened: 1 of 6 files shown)


theKHutDeveloper:
[(2) Store the Number of Word Documents into the data warehouse](https://trello.com/c/Gsse6kux/98-2-store-the-number-of-word-documents-into-the-data-warehouse)

spec/domain/performance/metrics/number_of_word_files_spec.rb
```diff
@@ -25,20 +25,48 @@ module Performance
     let(:content_without_document_keys) { build(:content_item, details: {}) }
     let(:content_without_details_key) { build(:content_item) }
 
-    it "returns the number of word files present" do
-      expect(subject.new(content_with_word_files).run).to eq(number_of_word_files: 3)
-    end
+    context "#run" do
```
pmanrubia: Instead of `context`, `describe` feels more accurate.

PR #520, opened 2018-02-27
Number of feedex issues

KyleMacPherson:
Initial work for set up importing of feedex results from the feedex endpoint which is currently a work in progress.

Following the pattern used to import Google Analytics data as seen [here](https://github.com/alphagov/content-performance-manager/pull/486)

### DIFF
--- a/app/models/etl/feedex.rb
+++ b/app/models/etl/feedex.rb
@@ -39,7 +39,8 @@ private
                number_of_issues,
                dimensions_items.id
         FROM events_feedexes, dimensions_items
-        WHERE page_path = base_path AND latest = 'true'
+        WHERE page_path = base_path
+          AND events_feedexes.date = '#{date_to_s}'
       ) AS s
       WHERE dimensions_item_id = s.id AND dimensions_date_id = '#{date_to_s}'
     SQL
@@ -52,8 +53,9 @@ private
       WHERE date = '#{date_to_s}' AND
         page_path in (
            SELECT base_path
-           FROM dimensions_items
-           WHERE latest = 'true'
+           FROM dimensions_items, facts_metrics
+           WHERE dimensions_items.id = facts_metrics.dimensions_item_id
+           AND facts_metrics.dimensions_date_id = '#{date_to_s}'
         )
     SQL
   end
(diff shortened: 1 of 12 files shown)


db/migrate/20180226111742_add_number_of_issues_to_facts_metrics.rb
```diff
@@ -0,0 +1,5 @@
+class AddNumberOfIssuesToFactsMetrics < ActiveRecord::Migration[5.1]
+  def change
+    add_column :facts_metrics, :number_of_issues, :integer
```
pmanrubia: Can you please add a default to `0`?

spec/services/feedex_service_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        .with(date, 2, 20).and_return(response_2)
+      allow(support_api).to receive(:feedback_by_day)
+        .with(date, 3, 20).and_return(response_3)
+    end
+
+    it 'makes the correct request and yields the first page' do
+      expect { |b| subject.find_in_batches(&b) }.to yield_successive_args(
+        [
+          {
+            date: date,
```
pmanrubia: It is a bit better to be explicit in the expectation and don't rely on a variable defined within the test.

spec/services/feedex_service_spec.rb
```diff
@@ -0,0 +1,93 @@
+require 'gds_api/support_api'
+RSpec.describe FeedexService do
+  let(:support_api) { double('support_api') }
+  let(:subject) { described_class.new(date, 20, support_api) }
+  describe '#find_in_batches' do
```
pmanrubia: Could you please add an empty line?

app/models/etl/feedex.rb
```diff
@@ -0,0 +1,66 @@
...
+      UPDATE facts_metrics
+      SET number_of_issues = s.number_of_issues
+      FROM (
+        SELECT base_path,
+               number_of_issues,
+               dimensions_items.id
+        FROM events_feedexes, dimensions_items
+        WHERE page_path = base_path AND latest = 'true'
+      ) AS s
+      WHERE dimensions_item_id = s.id AND dimensions_date_id = '#{date_to_s}'
```
pmanrubia: Could you please check https://github.com/alphagov/content-performance-manager/pull/524 and update accordingly? We should not rely on `latest` for the queries.

PR #521, opened 2018-02-27
Extract content from schemas: 'guide', 'travel_advice', 'place' & 'licence'

theKHutDeveloper:
[Trello: Store content from all formats](https://trello.com/c/k1WaJpum/115-3-store-content-from-all-formats-iteration-2)
This story continues the previous work started to store content for different content schema
types ->[Trello: Storing the content in the data warehouse over time](https://trello.com/c/qfMM7eeb/52-3-storing-the-content-in-the-data-warehouse-over-time). The schemas listed above do not store content the same way as in the previous story.

Storing content will enable us to apply content quality calculations.

### DIFF
--- a/app/models/dimensions/item.rb
+++ b/app/models/dimensions/item.rb
@@ -47,12 +47,15 @@ private
     detailed_guide
     document_collection
     fatality_notice
+    guide
     help
     hmrc_manual_section
     html_publication
+    licence
     manual
     manual_section
     news_article
+    place
     publication
     service_manual_guide
     simple_smart_answer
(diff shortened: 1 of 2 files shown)


spec/models/dimensions/item_spec.rb
```diff
@@ -39,6 +39,32 @@
         expect(item.get_content).to eq("Lorem ipsum dolor sit amet.")
       end
 
+      it "returns content json if schema is 'licence'" do
+        json = { schema_name: "licence",
```
MatMoore: This looks good but I think with these tests it's easy to test a JSON payload that isn't what the publishing API actually returns. 

There is a test helper called `GovukSchemas::RandomExample` which can help avoid this - it generates a random example that is validated against the schemas, so if you use that we have more confidence that all the field names match up.

It's a bit awkward to do because there are different schemas for notifications, publishing, and content store. Some existing examples:
- https://github.com/alphagov/email-alert-service/blob/0a3ed2701816a3aa11ed0639e31aea8ff2cfb196/spec/integration/random_input_spec.rb
-
 https://github.com/alphagov/rummager/blob/1b3d7d2fb4e5c13c1e17e6493150ef77ab12b349/spec/support/spec_helpers.rb#L40

app/models/dimensions/item.rb
```diff
@@ -61,18 +64,56 @@ def self.create_empty(content_id, base_path)
...
+  end
+
+  def extract_place(json)
+    html = []
+    html << json.dig("details", "introduction")
+    html << json.dig("details", "more_information")
+    html.join(" ")
+  end
+
+  def extract_parts(json)
```
MatMoore: this logic is really similar to what we use in search for creating the `indexable_content` field. Do you think we should share the code across the apps?

spec/models/dimensions/item_spec.rb
pmanrubia: Good point @MatMoore, happy to add integration tests in a similar way you have done in `email-alert-service`. I have collected your comment to address it in this quarter.

app/models/dimensions/item.rb
pmanrubia: Really worth exploring!

PR #528, opened 2018-02-28
Force the API to always return JSON

MatMoore:
At the moment the API errors if you go to it in the browser,
because it tries to render an HTML template that doesn't exist.

Not sure if there's a cleaner way to do this in rails, but this
forces it to always use the JSON template and content type.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,6 +5,9 @@ Rails.application.routes.draw do
 
   get '/audits', to: redirect(Plek.find('content-audit-tool', status: 302))
 
-  get '/api/v1/metrics/:content_id', to: "metrics#show"
+  scope '/api', defaults: { format: :json } do
+    get '/v1/metrics/:content_id', to: "metrics#show"
+  end
+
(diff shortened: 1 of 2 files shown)


peteglondon:
Probably easier just to put something like 
```
def show
    render json: Facts::Metric
      .joins(:dimensions_date)
      .joins(:dimensions_item)
      .where(dimensions_items: { content_id: content_id })
      .where('dimensions_dates.date between ? and ?', from, to)
      .order('dimensions_dates.date asc')
      .group('dimensions_dates.date')
      .sum("facts_metrics.#{metric}")
  end
```
to force it to return json

pmanrubia:
What about using `routes.rb`. I haven't tried it, but this might work.

```ruby
namespace :api, :defaults => { :format => 'json' } do
  ...
end
```

MatMoore:
Aha, thanks @pmanrubia, that seems to do the job. I've updated the PR. The `render :json` didn't work here because we're using a template to build the JSON rather than serialising objects.

PR #533, opened 2018-03-01
Add API documentation

MatMoore:
This is a skeleton for some API documentation. It's not the final thing, just a starting point. I'm putting it in this repo because it's easier to maintain if it lives with the code.

Note that the examples won't work right now, because we temporarily put the API behind signon in https://github.com/alphagov/content-performance-manager/pull/531

This uses the GDS tech docs template and an OpenAPI spec that is converted to a markdown template using widdershins. This code was forked from the Content API documentation: https://github.com/alphagov/govuk-content-api-docs

If we're happy with this pattern, we might want to think about making this
more reusable (perhaps integrating with https://github.com/alphagov/govuk-rails-app-template)

I've marked the API as alpha, because we're still working out what the API will look like, there is no support mailing list etc.

Because of this I haven't put much detail into the OpenAPI spec - we can iterate it as we evolve the API.

The next step is to publish to github pages and set up a publishing.service.gov.uk domain.  I'm not setting up continuous deployment yet although this is something we could look into.  The content API doc site used to be periodically deployed by jenkins although this is broken at the moment.

Trello: https://trello.com/c/qZppXq2l/134-setup-the-api-documentation-technical-stuff


MatMoore:
@pmanrubia that sounds like a good idea. That would make it simpler to automate deployment as well. I'll merge this for now and then raise another PR for it.

MatMoore:
@pmanrubia I think this is quite difficult actually, because the part we actually run is a Makefile. The bit that does the API reference is javascript, and that needs to run before the middleman build.

The middleman tasks just get pulled in from the gem through `require 'middleman-gh-pages'` and I think it might get more confusing if I tried to move the Makefile stuff into new rake tasks.

tijmenb:
What's the reason behind having the documentation site as part of the repo instead of only exposing an OpenAPI spec like content-store, and having a separate repo for the docs?

MatMoore:
We'd just rather keep them close together. It means when you update the API you can update the docs in the same PR. With them in separate repos it's easy to let them get out of sync and not pay as much attention to the docs.

tijmenb:
Ok, but I'm a bit worried about the proliferation of documentation techniques. We now have things documented in `/docs` (sometimes integrated in developer docs), OpenAPI based like content store, and this generated HTML.

Note that it turned out to be really messy and inconvenient to have the generated HTML in the docs in the gem (https://github.com/alphagov/govuk_schemas IIRC). You need to add the regeneration to the tests, otherwise people forget to regenerate every time they make a commit.

MatMoore:
Except for bundling the docs in the same repo, I'm aiming to stay as consistent as possible with the content API docs. That team already thought a lot about this, and imo the documentation and tools they've built are the best we have for GOV.UK APIs.

I considered using the developer docs but after talking with @kevindew we agreed that this is the wrong place for hosting public facing APIs.

The generated html lives on a separate `gh-pages` branch so it won't show up in PRs or be deployed with the API itself. There is already a solution for the publishing docs automatically, but it's not enabled right now. If I make any changes to how this works I'll implement it for both apis.

PR #538, opened 2018-03-05
Import quality metrics

KyleMacPherson:
Import and store content quality metrics from the quality metrics service


db/schema.rb
```diff
@@ -74,6 +74,8 @@
     t.integer "number_of_word_files"
     t.boolean "dirty", default: false
     t.string "status", default: "live"
+    t.integer "number_of_spelling_errors"
+    t.integer "number_of_readability_issues"
```
KyleMacPherson: I think I will change these to just `spelling_count` `readability_count` etc. What do you think @pmanrubia?

spec/domain/importers/content_details_spec.rb
```diff
@@ -33,6 +38,12 @@
       expect(latest_dimension_item.reload.number_of_word_files).to eq 94
     end
 
+    it 'stores the given quality metrics' do
+      subject.run
+      expect(latest_dimension_item.reload.number_of_spelling_errors).to eq 3
+      expect(latest_dimension_item.reload.number_of_readability_issues).to eq 10
+    end
+
```
pmanrubia: We also need to be sure that if an exception is thrown the exception is propagated, no data is persisted so the job is retried.

pmanrubia:
paired with @KyleMacPherson

PR #541, opened 2018-03-05
Handle GdsApi::HTTPNotFound Exception

peteglondon:
We are seeing this error in Sentry logs.
Do nothing when we get this error.


app/domain/importers/content_details.rb
```diff
@@ -18,10 +18,14 @@ def run
     item.update_attributes(attributes)
   rescue GdsApi::HTTPGone
     item.gone!
+  rescue GdsApi::HTTPNotFound
+    do_nothing
```
pmanrubia: 👍

PR #548, opened 2018-03-06
Update items in batches

peteglondon:
## As a
GDS devloper
## I want
The update of outdated content items to occur in batches
## So that
The performance of these updates is improved

At the moment each item marked as 'dirty' is individually duplicated and the old and new versions are saved, one at a time.

This is for the Trello card [Update outdated content in batches](https://trello.com/c/ubfK3Q2v/145-1-update-outdated-content-in-batches)

As part of this work we are limiting the processing of outdated items to those that were updated before a given date.


app/models/etl/dirty.rb
```diff
@@ -0,0 +1,30 @@
+class ETL::Dirty
```
pmanrubia: What about renaming this class to: `ETL::OutdatedItems`? or something more explicit about the fact that we are promoting new versions of the items?

app/models/etl/dirty.rb
```diff
@@ -0,0 +1,30 @@
...
+    create_new_version_for_dirty_items
+  end
+
+private
+
+  attr_reader :date
+
+  def create_new_version_for_dirty_items
+    Dimensions::Item.dirty_before(date).find_in_batches do |items|
+      new_items = items.map(&:new_version)
```
pmanrubia: What about breaking this block into two other methods that describe, at a higher level, what is happening?
Something along the following lines?
```ruby
Dimensions::Item.dirty_before(date).find_in_batches do |items|
  new_items = create_new_items(items)
  import_content_details(new_items)
end

File: spec/models/dimensions/item_spec.rb [hunk 3]
```diff
@@ -90,34 +90,53 @@ def build_raw_json(body:)
...
     end
+  end
 
-    it 'sets latest? and dirty? to false on the old item' do
-      dirty_item.new_version!
-
-      expect(dirty_item.reload).to have_attributes(
-        latest: false,
-        dirty: false
+  describe '#new_version' do
```
pmanrubia: This test seems to be duplicated.

pmanrubia:
👍

PR #553, opened 2018-03-07
Add additional validation and return error responses

MatMoore:
I've tried to follow the "Problem details" semantics (https://tools.ietf.org/html/rfc7807) because I don't think we have any other convention we use within GDS. If I'm wrong this can be changed fairly easily.

All the validation logic is extracted to a new class, to make use of `ActiveModel::Validation` without cluttering up the controller. This was inspired by this blog post https://coderwall.com/p/ea5vtw/validating-rest-queries-with-rails

In theory we should be able to validate requests against the same OpenAPI schema we use for the docs, but I don't think there is a good ruby library for this yet (though I did notice @kevindew is building https://github.com/kevindew/openapi3_parser !)

Trello: https://trello.com/c/K80isIdI/100-3-add-error-handling-to-api-endpoint

### DIFF
--- a/app/controllers/api/metrics_controller.rb
+++ b/app/controllers/api/metrics_controller.rb
@@ -1,5 +1,5 @@
 class Api::MetricsController < ApplicationController
-  before_action :validate_metric!
+  before_action :validate_params!
 
   def show
     query = Facts::Metric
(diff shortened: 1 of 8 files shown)


app/controllers/api/metrics_controller.rb
```diff
@@ -10,29 +10,26 @@ def show
...
+  def metric_params
+    @metric_params ||= Api::Metric.new(params)
   end
 
-  def validate_metric!
-    head :bad_request unless Facts::Metric.valid_metric? metric
+  def validate_params!
+    unless metric_params.valid?
+      response = {
+        type: "https://content-performance-api.publishing.service.gov.uk/errors/#validation-error",
```
pmanrubia: Shouldn't we be using `Plek`? 
Can you explain a bit more in the commit notes the details around the name: `content-performance-api`?

app/models/api/metric.rb
```diff
@@ -6,14 +6,14 @@ class Api::Metric
 
   attr_reader :metric, :from, :to, :content_id
 
-  validates :metric, presence: true, inclusion: {in: Facts::Metric::METRIC_WHITELIST}
-  validates :from, presence: true, format: {with: DATE_REGEX, message: "Dates should use the format YYYY-MM-DD"}
-  validates :to, presence: true, format: {with: DATE_REGEX, message: "Dates should use the format YYYY-MM-DD"}
-  validates :content_id, presence: true, format: {with: CONTENT_ID_REGEX, message: "Content ID must be a UUID."}
+  validates :metric, presence: true, inclusion: { in: Facts::Metric::METRIC_WHITELIST }
+  validates :from, presence: true, format: { with: DATE_REGEX, message: "Dates should use the format YYYY-MM-DD" }
```
pmanrubia: Would you mind squashing these format changes into: 90e017657619a8e8f92082dad4dbf583a2823606?

app/models/api/metric.rb
```diff
@@ -0,0 +1,34 @@
+class Api::Metric
+  DATE_REGEX = /\A\d\d\d\d-\d\d-\d\d\Z/
+  CONTENT_ID_REGEX = /\A[0-9A-F]{8}-[0-9A-F]{4}-[1-5][0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}\Z/i
```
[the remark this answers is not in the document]
MatMoore: I've made these private using `private_constant` but I couldn't move them to the private area without breaking the `validates` calls.

PR #558, opened 2018-03-08
Store content from 5 more formats

theKHutDeveloper:
This story [Trello: Store content from 13 formats (batch 3)](https://trello.com/c/7F654bYW/149-5-store-content-from-add-number-formats-batch-3) is a continuation of work to store content from content types, enabling us to apply content quality calculations. I have broken the story into two parts, with five schema types in this commit.

The schemas that form part of this commit are:
'email_alert_signup', 'transaction', 'finder_email_signup', 'location_transaction' and service_manual_topic.

Previous Work:
[PR 521 - Extract content from schemas: guide, travel_advice, 'place' & 'licence'](https://github.com/alphagov/content-performance-manager/pull/521)
[PR 490 - Store content v2](https://github.com/alphagov/content-performance-manager/pull/490)

### DIFF
--- a/app/models/dimensions/item.rb
+++ b/app/models/dimensions/item.rb
@@ -45,24 +45,29 @@ private
     corporate_information_page
     detailed_guide
     document_collection
+    email_alert_signup
     fatality_notice
+    finder_email_signup
     guide
     help
     hmrc_manual_section
     html_publication
     licence
+    location_transaction
     manual
     manual_section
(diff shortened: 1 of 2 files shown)


pmanrubia:
Good work!. Once we have all the formats in place, we need to extract this logic into its own class, and probably create presenters for each one of them, along with an integration test.

PR #572, opened 2018-03-13
Refactor content extraction

peteglondon:
Refactoring the content extraction into a number of classes.

Each content format now is implemented in it's own class in
app/models/content_extraction/parsers

The trello card is [Refactor: extract classes to get the content for a Content Item](https://trello.com/c/YFl3QjMR/159-2-refactor-extract-classes-to-get-the-content-for-a-content-item)

### DIFF
--- a/app/models/content_extraction/content_parser.rb
+++ b/app/models/content_extraction/content_parser.rb
@@ -1,5 +1,5 @@
-module Parsers
-  class ContentExtractors
+module ContentExtraction
+  class ContentParser
     include Singleton
 
     def self.register(*args)
@@ -38,5 +38,5 @@ module Parsers
     end
   end
 
-  Dir[File.dirname(__FILE__) + '/*.rb'].each { |file| require file }
+  Dir[File.dirname(__FILE__) + '/parsers/*.rb'].each { |file| require file }
 end
--- a/app/models/content_extraction/parsers/body_content_parser.rb
+++ b/app/models/content_extraction/parsers/body_content_parser.rb
@@ -1,4 +1,4 @@
-class Parsers::BodyContentParser
+class ContentExtraction::Parsers::BodyContentParser
   def parse(json)
     json.dig("details", "body")
   end
@@ -30,5 +30,5 @@ schemas_with_body = %w[
   ].freeze
 
 schemas_with_body.each do |schema|
-  Parsers::ContentExtractors.register(schema, Parsers::BodyContentParser.new)
+  ContentExtraction::ContentParser.register(schema, ContentExtraction::Parsers::BodyContentParser.new)
 end
--- a/app/models/content_extraction/parsers/email_alert_signup_parser.rb
+++ b/app/models/content_extraction/parsers/email_alert_signup_parser.rb
@@ -1,4 +1,4 @@
-class Parsers::EmailAlertSignupParser
+class ContentExtraction::Parsers::EmailAlertSignupParser
   def parse(json)
     html = []
     json.dig("details", "breadcrumbs").each do |crumb|
@@ -8,4 +8,4 @@ class Parsers::EmailAlertSignupParser
     html.join(" ")
   end
 end
-Parsers::ContentExtractors.register('email_alert_signup', Parsers::EmailAlertSignupParser.new)
+ContentExtraction::ContentParser.register('email_alert_signup', ContentExtraction::Parsers::EmailAlertSignupParser.new)
--- a/app/models/dimensions/item.rb
+++ b/app/models/dimensions/item.rb
@@ -11,7 +11,7 @@ class Dimensions::Item < ApplicationRecord
 
   def get_content
     return if raw_json.blank?
-    Parsers::ContentExtractors.extract_content(raw_json)
+    ContentExtraction::ContentParser.extract_content(raw_json)
   end
 
   def new_version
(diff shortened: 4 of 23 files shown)


app/models/content_extraction/content_parser.rb
```diff
@@ -0,0 +1,42 @@
+module ContentExtraction
+  class ContentParser
```
pmanrubia: What about renaming:

```ruby
ContentExtraction::ContentParser
```

to 

```ruby
Content::Parser
```

We would have a folder structure like:

```
- models
- models/content
- models/content/parsers/*
- models/content/parser.rb
```

app/models/content_extraction/content_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def self.register(*args)
+      instance.register(*args)
+    end
+
+    def self.extract_content(*args)
+      instance.extract_content(*args)
+    end
+
+    def extract_content(json)
```
pmanrubia: What about moving this method to the `private` area?

app/models/content_extraction/content_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+      schema = json.dig("schema_name")
+      parser = for_schema(schema)
+      raise InvalidSchemaError, "Schema does not exist: #{schema}" unless parser
+      parse_html parser.parse(json)
+    end
+
+    def register(schema_name, parser)
```
pmanrubia: What about moving this method to the `private` area?

app/models/content_extraction/parsers/email_alert_signup_parser.rb
```diff
@@ -0,0 +1,11 @@
+class ContentExtraction::Parsers::EmailAlertSignupParser
```
pmanrubia: I would remove the suffix `Parser` as we are already in a `Parsers` module. This apply to all parsers.

app/models/dimensions/item.rb
```diff
@@ -1,5 +1,6 @@
 require 'json'
 
+
```
pmanrubia: This empty line could be removed

spec/models/dimensions/item_spec.rb
```diff
@@ -10,11 +10,37 @@
     end
 
     context "when valid schema" do
-      it "returns content json if details.body exists" do
-        json = build_raw_json(body: 'the-body')
-
-        item = create(:dimensions_item, raw_json: json)
-        expect(item.get_content).to eq('the-body')
+      it 'returns content json from :body for all valid formats' do
```
pmanrubia: This seems like a unit test for the parser. If we unit test the method, it would be enough with stubbing the response. I would also add an integration [test as described in Mat's comment](https://github.com/alphagov/content-performance-manager/pull/521#discussion_r171204318).

app/models/content_extraction/parsers/body_content_parser.rb
```diff
@@ -0,0 +1,34 @@
+class ContentExtraction::Parsers::BodyContentParser
+  def parse(json)
+    json.dig("details", "body")
+  end
+end
+schemas_with_body = %w[
```
pmanrubia: This is cool, but I would rather not having code that initializes the parsers in plain ruby code. I know this is just a personal preference, and hopefully you are ok with it. 

I would be explicit about what parsers are registered in the main Parser class. Adding new parsers is quite an exceptional situation, and having more explicit code could be benefitial. Sorry, if I had not been clear about it before.

I would move this into something like:

```ruby
class ContentExtraction::Parsers::BodyContentParser
  def parse(json)
    json.dig("details", "body")
  end

  def schemas
   # List of schemas I can parse
  end
end
```

Then, the main class just need to register the same parser for each schema. It is more rudimentary, but it feels clearer to me.

pmanrubia:
As discussed with @peteglondon, we are going to merge these changes without addressing the comments as of now, as we have other PRs that depend on this refactoring. These comments will be tackled immediately after.

app/models/content_extraction/content_parser.rb
pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/576

app/models/content_extraction/content_parser.rb
pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/576

app/models/content_extraction/content_parser.rb
pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/576

app/models/content_extraction/parsers/email_alert_signup_parser.rb
pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/576

app/models/dimensions/item.rb
pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/576

spec/models/dimensions/item_spec.rb
pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/576

app/models/content_extraction/parsers/body_content_parser.rb
pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/576

PR #575, opened 2018-03-14
Remove code that calculate and exposes the metrics for the last day

KyleMacPherson:
[Trello card](https://trello.com/c/CjKzVOHf/161-5-retire-the-cpm-delete-non-used-code)

We have built a system that stores metrics through time
so we don't need now the code to store the metrics for
the last day.

We are deleting on the same go:

1. Support for Heroku deploys
2. Google Tag Manager


pmanrubia:
@KyleMacPherson could you please review my changes. Yours look GTM

KyleMacPherson:
Reviewed and approved Pablo!

PR #577, opened 2018-03-14
Rename number of issues

KyleMacPherson:
Rename `number_of_issues` to `feedex_comments` for clarity.

### DIFF
--- a/spec/requests/api/v1/metrics_spec.rb
+++ b/spec/requests/api/v1/metrics_spec.rb
@@ -78,10 +78,10 @@ RSpec.describe '/api/v1/metrics/:content_id', type: :request do
 
   describe 'Daily metrics' do
     before do
-      create :metric, dimensions_item: item, dimensions_date: day1, pageviews: 10, number_of_issues: 10
-      create :metric, dimensions_item: item, dimensions_date: day2, pageviews: 20, number_of_issues: 20
-      create :metric, dimensions_item: item, dimensions_date: day3, pageviews: 30, number_of_issues: 30
-      create :metric, dimensions_item: item, dimensions_date: day4, pageviews: 40, number_of_issues: 40
+      create :metric, dimensions_item: item, dimensions_date: day1, pageviews: 10, feedex_comments: 10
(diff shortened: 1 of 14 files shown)


db/migrate/20180309144814_rename_number_of_issues_to_feedex_comments.rb
```diff
@@ -0,0 +1,6 @@
+class RenameNumberOfIssuesToFeedexComments < ActiveRecord::Migration[5.1]
+  def change
+    rename_column :facts_metrics, :feedex_comments, :feedex_comments
```
pmanrubia: Not sure the rename is valid?

KyleMacPherson:
[(1) Rename `number_of_issues` to `feedex_comments` (Part of EPIC #216)](https://trello.com/c/txIU2saj/180-rename-numberofissues-to-feedexcomments)

db/migrate/20180309144814_rename_number_of_issues_to_feedex_comments.rb
KyleMacPherson: Sorry @pmanrubia, well spotted.  I should not have used `find_and_replace`! 😂 😔

spec/requests/api/v1/metrics_spec.rb
```diff
@@ -92,20 +92,20 @@
     end
 
     it 'returns `feedex issues` between two dates' do
-      get "/api/v1/metrics/number_of_issues/#{content_id}/time-series", params: { from: '2018-01-13', to: '2018-01-15' }
+      get "/api/v1/metrics/feedex_comments/#{content_id}/time-series", params: { from: '2018-01-13', to: '2018-01-15' }
```
pmanrubia: In this case we are changing an API endpoint that we shouldn't change (we are changing the URL of an API of another system)

pmanrubia: Sorry, @KyleMacPherson my bad. This is ok, as it is our internal API.
Discard this message, please.

PR #582, opened 2018-03-15
Monitoring and tracing

peteglondon:
Log the start time, end time and duration of ETL:: processes.
Log the number of items being processed.


app/models/concerns/traceable.rb
```diff
@@ -0,0 +1,15 @@
+require 'active_support/concern'
+
+module Concerns::Traceable
+  extend ActiveSupport::Concern
+
+  included do
+    def time(process:)
+      started = Time.now
```
pmanrubia: What about raising an exception if no block is passed?

app/models/concerns/traceable.rb
```diff
@@ -0,0 +1,15 @@
...
+      logger.info "Process: '#{process}' started at #{started}"
+      yield
+      ended = Time.now
+      logger.info "Process: '#{process}' ended at #{ended}, duration: #{ended - started}"
```
pmanrubia: It would be good to log when an exception is raised.
Just capture the exception, trace it, and re-raise it

app/models/etl/dates.rb
```diff
@@ -1,3 +1,5 @@
+#TODO do we need this class?
```
pmanrubia: Do you need this #TODO?

peteglondon: This was just a reminder to check if this class is needed, It is only used by tests so probably should be deleted.

pmanrubia: Ok, cool. Let's just remove it then 👍

PR #586, opened 2018-03-16
Add integration test to ETL::Master process

pmanrubia:
We need to ensure that our Master process runs is working as expected,
as it orchestrates all other ETL processes.

I am only covering the happy path of this process, as all the edge cases
and other interactions are tested at a unit level.

### DIFF
--- a/spec/integration/master_process_spec.rb
+++ b/spec/integration/master_process_spec.rb
@@ -1,0 +1,158 @@
+require 'sidekiq/testing'
+require 'gds_api/test_helpers/content_store'
+
+RSpec.describe 'Master process spec' do
+  include GdsApi::TestHelpers::ContentStore
+
+  let(:today) { Date.new(2018, 2, 21) }
+
+  around do |example|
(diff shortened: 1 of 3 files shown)


spec/integration/master_process_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+
+  let(:content_id) { 'id1' }
+  let(:base_path) { '/the-base-path' }
+  let(:item_content) { 'This is the content.' }
+
+  let!(:an_item) { create :dimensions_item, content_id: 'a-content-id' }
+  let!(:outdated_item) { create :dimensions_item, content_id: content_id, base_path: base_path, latest: false }
+  let!(:item) { create :dimensions_item, content_id: content_id, base_path: base_path, outdated: true, updated_at: 2.days.ago }
+
+  it 'something' do
```
peteglondon: What is it doing?

pmanrubia: I have renamed it, thank you 😊

PR #588, opened 2018-03-19
Use PAAS url for content quality metrics

KyleMacPherson:
Use PAAS rather than heroku to host the service for increased performance. 
I have checked that all is running as expected in Integration.

### DIFF
--- a/spec/integration/master_process_spec.rb
+++ b/spec/integration/master_process_spec.rb
@@ -104,7 +104,7 @@ RSpec.describe 'Master process spec' do
       passive: { 'count' => 5 },
       repeated_words: { 'count' => 8 },
     }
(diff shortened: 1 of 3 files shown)


spec/integration/master_process_spec.rb
```diff
@@ -104,7 +104,7 @@ def stub_quality_metrics_response
       passive: { 'count' => 5 },
       repeated_words: { 'count' => 8 },
     }
-    stub_request(:post, 'https://gov-quality-metrics.herokuapp.com/metrics').
+    stub_request(:post, 'https://govuk-content-quality-metrics.cloudapps.digital/metrics').
```
pmanrubia: Could you please fix up this commit with the previous one? In theory all commits should have all tests passing.

PR #591, opened 2018-03-20
Ensure ETL::Master only runs once per date

KyleMacPherson:
In order to ensure that we do not have duplicate
Dimensions::Date objects in our database.

### DIFF
--- a/app/data/etl/master.rb
+++ b/app/data/etl/master.rb
@@ -10,6 +10,7 @@ class ETL::Master
   end
 
   def process
+    raise DuplicateDateError if Dimensions::Date.exists?(date)
     time(process: :master) do
       ETL::OutdatedItems.process(date: dimensions_date.date)
       ETL::Metrics.process(date: dimensions_date.date)
@@ -25,4 +26,7 @@ private
   end
 
   attr_reader :date
+
+  class DuplicateDateError < StandardError;
+  end
 end
--- a/app/models/dimensions/date.rb
+++ b/app/models/dimensions/date.rb
@@ -30,6 +30,10 @@ class Dimensions::Date < ApplicationRecord
     end
   end
 
+  def self.exists?(date)
+    Dimensions::Date.where(date: date).exists?
+  end
+
   validates :date, presence: true
 
   validates :date_name, presence: true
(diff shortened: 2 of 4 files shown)


app/data/etl/master.rb
```diff
@@ -24,5 +25,12 @@ def dimensions_date
...
   end
 
+  def already_processed_for(date)
+    Dimensions::Date.where(date: date).exists?
+  end
+
   attr_reader :date
 end
+
+class DuplicateDateError < StandardError;
```
pmanrubia: Could you please define this error within the `ETL::Master` class?

app/data/etl/master.rb
```diff
@@ -10,6 +10,7 @@ def initialize(date: Date.yesterday)
   end
 
   def process
+    raise DuplicateDateError if already_processed_for(date)
```
pmanrubia: If we move the logic into `Dimensions::Date` we could write something like:

```ruby
Dimensions::Date.exits?(date)
```

spec/etl/master_spec.rb
```diff
@@ -17,6 +17,14 @@
     allow(ETL::Metrics).to receive(:process)
   end
 
+  def create_dimensions_date_for_previous_day
+    create(:dimensions_date, date: Date.new(2018, 2, 19))
+  end
+
+  it 'does not process if already processed for date' do
+    create_dimensions_date_for_previous_day
```
pmanrubia: What about inlining the code and write:

```ruby
create(:dimensions_date, date: Date.yesterday)

expect { subject.process }.to raise_error(DuplicateDateError)
```

KyleMacPherson:
@pmanrubia Thanks for the review Pablo, I have addressed the issues!

app/models/dimensions/date.rb
```diff
@@ -30,6 +30,10 @@ def self.for(date)
     end
   end
 
+  def self.exists?(date)
```
pmanrubia: Could you please add a unit test for this method?

spec/models/dimensions/date_spec.rb
```diff
@@ -1,6 +1,7 @@
 require 'rails_helper'
 
 RSpec.describe Dimensions::Date, type: :model do
+  let(:date) { ::Date.new(2017, 12, 21) }
```
pmanrubia: The date is only used in the `.exists?` `describe`. Would you mind moving it there? It is better to have the fixtures as close as possible to the test.

PR #596, opened 2018-03-20
Do not calculate content quality metrics again if content has not changed

peteglondon:
There is no need to re-calculate the quality metrics unless the content has changed.

This work breaks down as follows.

* Store a SHA1 digest of the content parsed from the raw JSON in the Dimensions::Item.
* Check that digest against the new content retrieved from the content store.
* Only fire the ImportQualityMetricsJob if the content digest is different.

The Trello card is [Do not calculate content quality metrics again if content has not changed](https://trello.com/c/GliSXCSl/170-3-do-not-calculate-content-quality-metrics-again-if-content-has-not-changed).

As a dev/ops of the data warehouse
I want to calculate quality metrics only when the content has been updated
So that I don't waste GOV.UK resources on infrastructure

### DIFF
--- a/app/data/metadata/parser.rb
+++ b/app/data/metadata/parser.rb
@@ -5,7 +5,8 @@ class Metadata::Parser
       Metadata::Parsers::NumberOfPdfs,
       Metadata::Parsers::NumberOfWordFiles,
       Metadata::Parsers::Metadata,
-      Metadata::Parsers::PrimaryOrganisation
+      Metadata::Parsers::PrimaryOrganisation,
+      Metadata::Parsers::ContentHash
     ]
   end
 
--- a/app/data/metadata/parser.rb
+++ b/app/data/metadata/parser.rb
@@ -6,7 +6,7 @@ class Metadata::Parser
       Metadata::Parsers::NumberOfWordFiles,
       Metadata::Parsers::Metadata,
       Metadata::Parsers::PrimaryOrganisation,
-      Metadata::Parsers::ContentHash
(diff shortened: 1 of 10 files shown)


app/data/metadata/parser.rb
```diff
@@ -5,7 +5,8 @@ def initialize
       Metadata::Parsers::NumberOfPdfs,
       Metadata::Parsers::NumberOfWordFiles,
       Metadata::Parsers::Metadata,
-      Metadata::Parsers::PrimaryOrganisation
+      Metadata::Parsers::PrimaryOrganisation,
+      Metadata::Parsers::ContentHash
```
pmanrubia: Could you please add a training `,`

spec/domain/importers/content_details_spec.rb
```diff
@@ -24,7 +24,8 @@
         public_updated_at: Time.new(2015, 6, 3, 11, 13, 44),
         primary_organisation_title: 'Home Office',
         primary_organisation_content_id: 'cont-id-1',
-        primary_organisation_withdrawn: false
+        primary_organisation_withdrawn: false,
+        content_hash: 'ContentHashContentHash'
```
pmanrubia: Could you please add ','?

spec/domain/importers/content_details_spec.rb
```diff
@@ -49,7 +50,8 @@
...
```
pmanrubia: Could you please add ','?

PR #603, opened 2018-03-22
Store all locales in the data warehouse

peteglondon:
**As a** Managing editor
**I want** have data for all content items
**So that** I can make informed decisions about any content item that I am responsible for

# What
Store all locales in the data warehouse. The sandbox and metrics endpoints filter out only the 'en' locale

# Why
We are attempting to only store content that has the 'en' locale in the data warehouse. The code that filters out the default locale is not working as intended. We have realised it is better to store everything.

The Trello card is [Store all locales in the data warehouse](https://trello.com/c/Md4ac5Xs/208-store-all-locales-in-the-data-warehouse)

### DIFF
--- a/app/data/etl/outdated_items.rb
+++ b/app/data/etl/outdated_items.rb
@@ -39,7 +39,7 @@ private
   def import_content_details(items)
     log process: :import_content_details, message: "creating #{items.length} ImportContentDetailsJobs"
     items.each do |item|
-      ImportContentDetailsJob.perform_async(item.content_id, item.base_path, item.locale)
+      ImportContentDetailsJob.perform_async(item.id)
     end
   end
 end
--- a/app/data/etl/outdated_items.rb
+++ b/app/data/etl/outdated_items.rb
@@ -39,7 +39,7 @@ private
   def import_content_details(items)
     log process: :import_content_details, message: "creating #{items.length} ImportContentDetailsJobs"
     items.each do |item|
-      ImportContentDetailsJob.perform_async(item.content_id, item.base_path)
+      ImportContentDetailsJob.perform_async(item.content_id, item.base_path, item.locale)
     end
   end
 end
--- a/app/domain/importers/content_details.rb
+++ b/app/domain/importers/content_details.rb
@@ -2,15 +2,16 @@ require 'odyssey'
 
 class Importers::ContentDetails
   include Concerns::Traceable
-  attr_reader :items_service, :content_id, :base_path
+  attr_reader :items_service, :content_id, :base_path, :locale
 
   def self.run(*args)
     new(*args).run
   end
 
-  def initialize(content_id, base_path, _ = {})
+  def initialize(content_id, base_path, locale, _ = {})
     @content_id = content_id
     @base_path = base_path
+    @locale = locale
     @items_service = ItemsService.new
   end
 
@@ -18,14 +19,13 @@ class Importers::ContentDetails
     item = nil
     begin
       time(process: :content_details) do
-        item = Dimensions::Item.find_by(content_id: content_id, latest: true)
-
+        item = Dimensions::Item.by_natural_key(content_id: content_id, locale: locale).first
         item_raw_json = items_service.fetch_raw_json(base_path)
         attributes = Metadata::Parser.parse(item_raw_json)
         content_changed = item.content_hash != attributes[:content_hash]
         item.update_attributes(attributes)
 
-        ImportQualityMetricsJob.perform_async(item.id) if content_changed
+        import_quality_metrics_for_en(item) if content_changed
       end
     rescue GdsApi::HTTPGone
       item.gone!
@@ -36,6 +36,10 @@ class Importers::ContentDetails
   end
 
 private
+
+  def import_quality_metrics_for_en(item)
+    ImportQualityMetricsJob.perform_async(item.id) if locale == 'en'
+  end
 
   def handle_not_found
     log process: :content_details, message: "NotFound when retrieving '#{base_path}' from content store"
--- a/app/domain/importers/content_details.rb
+++ b/app/domain/importers/content_details.rb
@@ -1,6 +1,8 @@
 require 'odyssey'
 
 class Importers::ContentDetails
+  include Concerns::Traceable
+
   attr_reader :items_service, :id
 
   def self.run(*args)
@@ -30,7 +32,7 @@ class Importers::ContentDetails
     end
   end
 
-  private
+private
 
   def handle_not_found(base_path:)
     log process: :content_details, message: "NotFound when retrieving '#{base_path}' from content store"
--- a/app/domain/importers/content_details.rb
+++ b/app/domain/importers/content_details.rb
@@ -19,7 +19,7 @@ class Importers::ContentDetails
     item = nil
     begin
       time(process: :content_details) do
-        item = Dimensions::Item.by_natural_key(content_id: content_id, locale: locale).first
+        item = Dimensions::Item.by_natural_key(content_id: content_id, locale: locale)
         item_raw_json = items_service.fetch_raw_json(base_path)
         attributes = Metadata::Parser.parse(item_raw_json)
         content_changed = item.content_hash != attributes[:content_hash]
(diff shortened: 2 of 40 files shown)


app/models/dimensions/item.rb
```diff
@@ -8,6 +8,10 @@ class Dimensions::Item < ApplicationRecord
     where('updated_at < ?', date).where(outdated: true)
   end
 
+  scope :by_natural_key, ->(content_id:, locale:) do
+    where(content_id: content_id, locale: locale, latest: true)
```
pmanrubia: Any reason for not using `find_by`? It seems better to return the Item, and not the an `Array`

db/migrate/20180322104809_add_index_dimensions_items_locale.rb
```diff
@@ -0,0 +1,6 @@
+class AddIndexDimensionsItemsLocale < ActiveRecord::Migration[5.1]
```
pmanrubia: I am not sure we need this index, as we never search only by locale.

app/data/etl/outdated_items.rb
```diff
@@ -39,7 +39,7 @@ def create_new_items(items)
   def import_content_details(items)
     log process: :import_content_details, message: "creating #{items.length} ImportContentDetailsJobs"
     items.each do |item|
-      ImportContentDetailsJob.perform_async(item.content_id, item.base_path)
+      ImportContentDetailsJob.perform_async(item.content_id, item.base_path, item.locale)
```
pmanrubia: What about passing only the `item.id`? We are passing all these parameters, just to be able to find the right item.

db/migrate/20180322104809_add_index_dimensions_items_locale.rb
peteglondon: Do we need to add it to the index that includes content_id and latest?

pmanrubia: Not sure Pete. We could postpone the decision?

app/domain/importers/content_details.rb
```diff
@@ -2,46 +2,48 @@
...
-
-        item_raw_json = items_service.fetch_raw_json(base_path)
+        item = Dimensions::Item.find(id)
+        item_raw_json = items_service.fetch_raw_json(item.base_path)
         attributes = Metadata::Parser.parse(item_raw_json)
         content_changed = item.content_hash != attributes[:content_hash]
         item.update_attributes(attributes)
 
-        ImportQualityMetricsJob.perform_async(item.id) if content_changed
+        import_quality_metrics_for_en(item) if content_changed
```
pmanrubia: I would extract the logic to decide wether applying quality metrics or not to its own class. Something like:

```ruby
class Dimensions::Item
  def apply_quality_metrics?(item, new_attributes)
  end
end
```

or

```ruby
class QualityMetrics 
  def self.apply?(content_item, new_attributes)
  end
end
```

app/models/concerns/outdateable.rb
```diff
@@ -7,8 +7,8 @@ module Concerns::Outdateable
     scope :outdated, -> { where(outdated: true, latest: true) }
     scope :outdated_before, ->(date) { outdated.where('outdated_at < ?', date) }
 
-    def outdate!
-      update_attributes!(outdated: true, outdated_at: Time.zone.now)
+    def outdate!(base_path:)
```
pmanrubia: I would not pass the `base_path` into the concern as it seems out of its main responsibility. This is going to change shortly once we start using `base_paths` as the natural key for Items, but for the time being, making the update explicit in https://github.com/alphagov/content-performance-manager/blob/87e84fb826f7b41d4b7d18d918622dcbb5e5f18d/lib/publishing_api_consumer.rb#L20-L21, along with an explanation in the commit notes would be beneficial.

app/models/dimensions/item.rb
```diff
@@ -5,6 +5,10 @@ class Dimensions::Item < ApplicationRecord
 
   validates :content_id, presence: true
 
+  def self.by_natural_key(content_id:, locale:)
```
pmanrubia: I know it is already been covered in the integration tests, but could you please add a unit test here?

db/migrate/20180321152146_add_locale_to_dimensions_items.rb
```diff
@@ -0,0 +1,5 @@
+class AddLocaleToDimensionsItems < ActiveRecord::Migration[5.1]
+  def change
+    add_column :dimensions_items, :locale, :string
```
pmanrubia: Locale should be `not null`. In this PR, I would default it to `en`, although we would need to move the initialiser into the model.

spec/domain/importers/content_details_spec.rb
```diff
@@ -2,14 +2,21 @@
...
+    let!(:parsed_content_hash) { 'NewContentHash' }
+    let!(:latest_dimension_item_fr) do
+      create(:dimensions_item,
+        content_id: content_id, base_path: base_path, locale: 'fr',
+        latest: true, raw_json: nil, content_hash: existing_content_hash)
+    end
     let!(:latest_dimension_item) do
       create(:dimensions_item,
-        content_id: content_id, base_path: base_path,
+        content_id: content_id, base_path: base_path, locale: 'en',
```
pmanrubia: Shouldn't `en` be set automatically for all DimensionItems in the `FactoryBot` factory?

spec/integration/initial_load_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'gds_api/test_helpers/content_store'
+
+RSpec.describe 'Initial load from publishing api' do
```
pmanrubia: 👍

spec/integration/master_process_spec.rb
```diff
@@ -16,11 +16,20 @@
...
   let(:base_path) { '/the-base-path' }
+  let(:locale) { 'en' }
   let(:item_content) { 'This is the content.' }
 
-  let!(:an_item) { create :dimensions_item, content_id: 'a-content-id' }
-  let!(:item) { create :dimensions_item, content_id: content_id, base_path: base_path, latest: false }
-  let!(:outdated_item) { create :outdated_item, content_id: content_id, base_path: base_path }
+
+  let!(:an_item) { create :dimensions_item, content_id: 'a-content-id', locale: locale }
+  let!(:outdated_item) do
```
pmanrubia: It seems that during the `git rebase` we have lost the changes [of this commit](https://github.com/alphagov/content-performance-manager/pull/605/commits/9f6e2332501344b4060c676ad3bf2990e3681403)

spec/lib/publishing_api_consumer_spec.rb
```diff
@@ -8,24 +8,26 @@
   it_behaves_like 'a message queue processor'
 
   context 'when the Dimensions::Item already exists - all events but unpublish' do
-    let(:latest_item) { create(:dimensions_item, latest: true, outdated: false) }
+    let!(:latest_item_en) { create(:dimensions_item, latest: true, outdated: false, locale: 'en') }
+    let!(:latest_item_de) { create(:dimensions_item, latest: true, outdated: false, locale: 'de') }
```
pmanrubia: I would make it implicit that all items are `latest`, `outdated` and `en` in the Factory. And only set values if this is not the case. This would improve readability.

spec/models/dimensions/item_spec.rb
```diff
@@ -1,8 +1,21 @@
 require 'rails_helper'
 
 RSpec.describe Dimensions::Item, type: :model do
+  let(:now) { Time.new(2018, 2, 21, 12, 31, 2) }
+
   it { is_expected.to validate_presence_of(:content_id) }
 
+  describe '##by_natural_key' do
```
pmanrubia: Only one `#` is needed

spec/models/concerns/outdateable_spec.rb
```diff
@@ -48,20 +48,26 @@
   end
 
   describe '#outdate!' do
+    let(:new_base_path) { '/new/base/path' }
```
pmanrubia: Please, read my comment above. I'd rather not mixing `base_path` in the `outdateable` concern.

app/models/dimensions/item.rb
peteglondon: It already has one ```describe '##by_natural_key' do``` mentioned below

pmanrubia:
LGTM 👍 
Could you please test in integration before merging?

PR #609, opened 2018-03-26
Cache all API responses for a day

MatMoore:
Data is updated daily so we can set a long cache time.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -1,4 +1,6 @@
 class Api::BaseController < ApplicationController
+  before_action :set_cache_headers
+
   rescue_from(ActionController::UnpermittedParameters) do |pme|
     error_response(
       "unknown-parameter",
@@ -8,6 +10,15 @@ class Api::BaseController < ApplicationController
   end
 
 private
+
+  def set_cache_headers
+    # Set cache headers to expire the page at 1am when we fetch new data.
+    expiry_time = Time.zone.tomorrow.at_beginning_of_day.change(hour: 1)
+    current_time = Time.zone.now
+    cache_time = (expiry_time - current_time) % 24.hour
+
+    expires_in cache_time, public: true
+  end
 
   def error_response(type, error_hash)
     # Type is an arbitrary URI identifying the error type
--- a/app/controllers/api/healthcheck_controller.rb
+++ b/app/controllers/api/healthcheck_controller.rb
@@ -1,5 +1,6 @@
 class Api::HealthcheckController < Api::BaseController
   skip_before_action :authenticate_user!
+  skip_before_action :set_cache_headers
 
   def index
     database = ActiveRecord::Base.connected? ? :ok : :critical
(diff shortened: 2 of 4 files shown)


app/controllers/api/base_controller.rb
```diff
@@ -9,6 +11,10 @@ class Api::BaseController < ApplicationController
 
 private
 
+  def set_cache_headers
+    expires_in 24.hours, public: true
```
pmanrubia: The data is updated daily, but It might be more accurate if we expire it at the end of the day, as we will have fresh data every day at 1am.

app/controllers/api/base_controller.rb
```diff
@@ -1,4 +1,6 @@
 class Api::BaseController < ApplicationController
+  before_action :set_cache_headers
```
pmanrubia: It would be good to add a test that ensures that we have the headers in place the first time. We can trust then that the framework will support caching in subsequent requests.

Something simple like:

```ruby  
it "set an ETag header" do
    response.headers['ETag'].should be_present
  end
it "sets 'Last-Modified'" do
    response.headers['Last-Modified'].should be_present
end 
```

app/controllers/api/base_controller.rb
```diff
@@ -9,6 +11,10 @@ class Api::BaseController < ApplicationController
 
 private
 
+  def set_cache_headers
```
pmanrubia: Should we exclude it from the `healthcheck` controller?

app/controllers/api/base_controller.rb
MatMoore: How do you want this to work?

Do you mean something like this?

```
    hour = Time.now.utc.hour
    expiry_hour = 1
    cache_duration = (expiry_hour - hour) % 24
    expires_in cache_duration.hours, public: true
```

app/controllers/api/base_controller.rb
MatMoore: I don't think this test is valuable. I would expect it to pass even without the code I just put in?

app/controllers/api/base_controller.rb
pmanrubia: yes, something like:

```ruby
expiry_time = Time.zone.tomorrow.at_beginning_of_day.change(hour: 1)
current_time = Time.zone.now 

hours = ((expiry_time - current_time) / 1.hour).round
```

app/controllers/api/base_controller.rb
pmanrubia: Sorry I didn’t try the code myself. I just thought that it would work as the headers shouldn’t exist unless we set them explicitly.  I find value in being explicit about having a test that give us confidence that our cache is in place. Can you think of a better way?

pmanrubia: Sorry I didn’t try the code myself. I just thought that it would work as the headers shouldn’t exist unless we set them explicitly.  I find value in being explicit about having a test that give us confidence that our cache is in place. Can you think of a better way?

pmanrubia:
Could you please add unit tests for the controllers so:

1. We are sure that we don't cache the healthcheck api call
2. We set up cache headers in the controller ?

MatMoore:
@pmanrubia I've added tests and changed the logic to cache until 1am.

PR #610, opened 2018-03-26
Add docm, and docx extensions when parsing closed formats

theKHutDeveloper:
[Trello: Include extensions .docx, docm to count Word documents](https://trello.com/c/SEm5OomS/169-2-include-extensions-docx-docm-to-count-word-documents)

Previously our solution included these extensions, but it was not very 
robust as it would also accept an invalid word file extension if it ended
'.docj' for example.

Our solution now means that only the specified extensions will be accepted.

We have move the logic to extract the content to the `NumberOfFiles` so
when counting the number of closed formats, we only need to specify the
extensions.

These changes have impacted `NumberOfPdfs` which has also been 
adapted to adhere to the changes made.

### DIFF
--- a/app/data/metadata/parsers/number_of_files.rb
+++ b/app/data/metadata/parsers/number_of_files.rb
@@ -1,23 +1,30 @@
 module Metadata::Parsers::NumberOfFiles
+  def self.number_of_files(body, extensions_regex)
+    documents = extract_documents(body)
+    all_links = all_links(documents)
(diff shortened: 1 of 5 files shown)


theKHutDeveloper:
[(2) Include extensions .docx, docm to count Word documents](https://trello.com/c/SEm5OomS/169-2-include-extensions-docx-docm-to-count-word-documents)

PR #615, opened 2018-03-27
Only test PublishingAPI events in integration test

pmanrubia:
@peteglondon:

We can have confidence in our integration tests (end_to_end tests) by:

1. Testing the master process and the data acquisition
2. Testing how the Items dimension grows when handling events from `PublishingAPI`

This PR removes all assertions and set up from #603 related to data acquisition and 
limits the scope to the Items dimension and its relationship with the events that come
from the `PublishingAPI`

These are the benefits that I find:

1. We still have end to end tests, as we simulate events from Publishing API, and we 
make our code assertions after running the master process
2. Improves readability and long-term maintainability as it only needs to grow when we
change how we handle events

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -272,7 +272,7 @@ GEM
       omniauth (~> 1.2)
     os (0.9.6)
     parallel (1.12.1)
-    parser (2.5.0.4)
+    parser (2.5.0.3)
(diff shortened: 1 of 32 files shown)


pmanrubia:
Closed as commits have been used in https://github.com/alphagov/content-performance-manager/pull/603

PR #616, opened 2018-03-27
Remove 'gov.uk' prefix from ga base path

KyleMacPherson:
Remove prefix `/https://www.gov.uk/` from Google Analytics if needed so we
can match the GA events with the dimension item.

### DIFF
--- a/app/data/etl/ga.rb
+++ b/app/data/etl/ga.rb
@@ -12,6 +12,7 @@ class ETL::GA
   def process
     time(process: :ga) do
       extract_events
+      transform_events
       load_metrics
     end
   end
@@ -25,6 +26,10 @@ private
       Events::GA.import(events, batch_size: 10_000)
       batch += 1
     end
+  end
+
+  def transform_events
+    fix_invalid_prefix_in_page_paths
   end
 
   def load_metrics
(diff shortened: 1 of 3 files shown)


app/data/etl/ga.rb
```diff
@@ -27,6 +28,15 @@ def extract_events
     end
   end
 
+  def transform_events
+    events_with_prefix = Events::GA.where("page_path ~ '^\/https:\/\/www.gov.uk'")
+    log(process: :ga, message: "Transforming #{events_with_prefix.count} events with page_path starting https://gov.uk")
+    events_with_prefix.each do |ga|
```
pmanrubia: It is better to use `find_each` so not all data is loaded into memory.

pmanrubia: Instead of `ga`, I would rename it to `event`

app/data/etl/ga.rb
```diff
@@ -27,6 +28,15 @@ def extract_events
     end
   end
 
+  def transform_events
+    events_with_prefix = Events::GA.where("page_path ~ '^\/https:\/\/www.gov.uk'")
```
pmanrubia: Could you please extract a method for the prefix transformation. This way we don't care about the implementation details when reading the code. Something similar to: `fix_invalid_prefix_in_page_paths`?

KyleMacPherson:
[(3) Remove prefix `/https://www.gov.uk/` from Google Analytics and backfill GA for 1 month (part of reliability EPIC)](https://trello.com/c/p6NaaUzR/197-3-remove-prefix-https-wwwgovuk-from-google-analytics-and-backfill-ga-for-1-month-part-of-reliability-epic)

PR #650, opened 2018-04-13
Add a schema diagram

MatMoore:
This adds a digram of the schema to the README to explain how the data warehouse is structured.

[Preview](https://github.com/alphagov/content-performance-manager/blob/3618764b1a5c17bce037daed8316ecbc21ba2465/README.md)


MatMoore:
I'm not concerned about it growing big right now - if it becomes a problem we can always move it later. I see it as the equivalent to showing a screenshot in the README - it shows what the thing is and you can tell if the repository is relevant to you without reading a wall of text.

At the moment you just have to run a rake task to regenerate the diagram itself, but if we have a free text description of every column it will take a lot of work to keep it up to date and we'll miss things, so that's the part I would want to streamline - maybe by moving it to the dev docs so there's a process in place for maintaining it.

What level of detail would you expect in `doc/facts_and_dimensions.md`?

If it's just the metrics themselves we can generate docs for those based on [this metadata](https://github.com/alphagov/content-performance-manager/blob/master/config/metrics.yml).

MatMoore:
@pmanrubia I'm unsure what the state of this PR is. Do you want me to close it?

MatMoore:
Going to close for now as I'm not actively working on it. Feel free to reopen.

PR #654, opened 2018-04-16
Change directory structure

KyleMacPherson:
Change the file structure in order to better reflect our codebase. 

The application is no longer a traditional rails app and as such the directory structure of our app no longer made a lot of sense. This PR moves the ETL directory into a more prominent position in our file structure and encapsulates all of our ETL logic, with different ETL processes having their own subdirectories and processors in order to make the whole process easier to understand.

The publishing api consumer is brought out of the lib folder and into a streams folder in the app directory to help developers to be aware of this element of the application and more easily understand it's use.

The services for each ETL process are added to the ETL subdirectories to keep logic for each process grouped together.

There are no changes to the actual codebase other than the renaming of classes to reflect the new file structure.


KyleMacPherson:
[Speed up loading process](https://trello.com/c/sz0loMBR/732-speed-up-loading-process)

spec/domain/importers/quality_metrics_spec.rb
```diff
@@ -1,55 +0,0 @@
-RSpec.describe Importers::QualityMetrics do
```
pmanrubia: This test seems to have been removed.

PR #655, opened 2018-04-18
Refactor ETL and publishing API stream so we store every edition of a content item

MatMoore:
# Summary
This changes both the nightly ETL process and the publishing API consumer so that multiple editions of a content item can be tracked per day. At the moment, we update `Dimensions::Item` records whenever we see multiple messages for the same content item in the same day, instead of inserting a new record.

This will have implications if we want to show historical quality metrics within publishing apps, because publishing apps show users a complete version history, rather than daily snapshots.

We're also storing fact-like data on the dimension itself, which is not following proper conventions and may make it harder to adopt 3rd party tools.

The solution we agreed upon was to create a new facts table with edition-level granularity, in addition to the current facts table that is a daily snapshot of every content item. This PR doesn't add that facts table - I'll do this in a future PR.

Trello: https://trello.com/c/SWlCL8xI/240-5-track-content-editions-metrics-through-time

# How it works now
- The publishing API consumer creates a new `Dimensions::Item` if there is no existing item for the same content_id with `latest: true`
- Otherwise, it marks the existing item as `outdated: true` and updates its base path in place
- Overnight, the `OutdatedItemsProcessor` processes every item that has `outdated: true` and `latest: true`
  - It creates a new version of the item with `latest: true, outdated: false`
  - It updates the previous version to be `latest: false, updated: false`
  - It triggers an async job to set the content details and quality metrics on the new version
- Then, `MetricsProcessor` creates a `Facts::Metric` record for every item with `latest: true`

# How it should work after this PR
- The publishing API consumer creates `Dimensions::Item` records for every edition, and marks them as `outdated`.
- The publishing API consumer marks the latest item per content id per day using `latest=true`
- Overnight, the `OutdatedItemsProcessor` processes every item that has `outdated: true`
  - It triggers an async job to set the content details and quality metrics on the new version
  - That job sets `outdated: false` on the item
- Then, `MetricsProcessor` creates a `Facts::Metric` record for every item with `latest: true`

# How it should work when we're done with this story
We can remove the `outdated` flag and `OutdatedItemsProcessor` from the codebase.

- The publishing API consumer creates `Dimensions::Item` records for every edition
- The consumer creates a `Facts::Version` record for each edition
  - It populates it with content details from the message
  - It fetches the quality metrics and records those too
- It marks the latest item per content id per day using `latest=true`
- Overnight, `MetricsProcessor` creates a `Facts::Metric` record for every item with `latest: true`
  - (Optional) It takes a snapshot of the version facts to make it easier to query daily metrics

# Things to fix
- [x] At the moment the "latest" version is whatever version we happened to process last. However, RabbitMQ can send us messages out of order, or duplicate messages. We should read the publishing API version and/or message creation time so that we can reconcile our concept of latest with the publishing API's concept of latest, and avoid creating multiple records for the same edition. 

  See [Rummager ADR 002](https://github.com/alphagov/rummager/blob/master/doc/arch/adr-002-handle-race-conditions-in-govuk-index.md) for a discussion of this problem in the context of search.

- [x] I need to add in an integration test that makes sure `Facts::Metric` has the right granularity when there are multiple editions per day

### DIFF
--- a/app/models/dimensions/item.rb
+++ b/app/models/dimensions/item.rb
@@ -14,9 +14,28 @@ class Dimensions::Item < ApplicationRecord
     Item::Content::Parser.extract_content(raw_json)
   end
 
-  def new_version
-    new_version = self.dup
-    new_version.assign_attributes(latest: true, outdated: false)
+  def copy_to_new_outdated_version!(current_base_path)
+    # Create a new version of this content item, assuming that a document's
+    # content_id and locale are fixed, but the base_path may change.
+    #
+    # Outdated means that quality metrics and content details we store on the dimension
+    # are missing and need to be updated.
+    #
+    # At the moment we populate these from a snapshot of the content item overnight.
+    # The concept of "outdated" will go away when we populate content item data directly
+    # from the publishing api consumer.
+    #
+    # This method is only expected to be called on the latest version of an item per day.
+    raise "Tried to create a new outdated version but this version is not the latest" unless latest
+
+    new_version = Dimensions::Item.create_empty(
+      content_id: content_id,
+      base_path: current_base_path,
+      locale: locale
+    )
+
+    update_attributes(latest: false)
+
     new_version
   end
 
--- a/app/models/dimensions/item.rb
+++ b/app/models/dimensions/item.rb
@@ -14,7 +14,7 @@ class Dimensions::Item < ApplicationRecord
     Item::Content::Parser.extract_content(raw_json)
   end
 
-  def copy_to_new_outdated_version!(current_base_path)
+  def copy_to_new_outdated_version!(base_path:, payload_version:)
     # Create a new version of this content item, assuming that a document's
     # content_id and locale are fixed, but the base_path may change.
     #
@@ -30,8 +30,9 @@ class Dimensions::Item < ApplicationRecord
 
     new_version = Dimensions::Item.create_empty(
       content_id: content_id,
-      base_path: current_base_path,
-      locale: locale
+      base_path: base_path,
+      locale: locale,
+      payload_version: payload_version
     )
 
     update_attributes(latest: false)
@@ -47,14 +48,15 @@ class Dimensions::Item < ApplicationRecord
     attributes[:locale] == 'en' && attributes[:content_hash] != content_hash
   end
 
-  def self.create_empty(content_id:, base_path:, locale:)
+  def self.create_empty(content_id:, base_path:, locale:, payload_version:)
     create(
       content_id: content_id,
       base_path: base_path,
       locale: locale,
       latest: true,
       outdated: true,
-      outdated_at: Time.zone.now
+      outdated_at: Time.zone.now,
+      publishing_api_payload_version: payload_version
     )
   end
 end
--- a/app/streams/publishing_api_consumer.rb
+++ b/app/streams/publishing_api_consumer.rb
@@ -11,24 +11,51 @@ class PublishingApiConsumer
 private
 
   def process_message(message)
-    content_id = message.payload['content_id']
-    base_path = message.payload['base_path']
+    content_id = message.payload.fetch('content_id')
+    base_path = message.payload.fetch('base_path')
+    payload_version = message.payload.fetch('payload_version')
     locale = message.payload['locale'] || 'en'
 
     item = Dimensions::Item.by_natural_key(content_id: content_id, locale: locale)
+
     if item
-      handle_existing(item, base_path, message.delivery_info.routing_key)
+      existing_payload_version = item.publishing_api_payload_version || 0
+
+      if payload_version > existing_payload_version
+        handle_existing(
+          item: item,
+          base_path: base_path,
+          routing_key: message.delivery_info.routing_key,
+          payload_version: payload_version
+        )
+      else
+        # Skip out of date and repeated messages from the publishing API.
+        # This is to ensure that our "latest" record is actually the latest.
+        #
+        # RabbitMQ guarantees at-least-once delivery when using acknowledgements,
+        # so messages can be duplicated, plus we can't guarantee the order of
+        # processing because we can have multiple consumers working in parallel.
+        #
+        # This logic could later be extended to also process out of date messages,
+        # as long as we preserve the latest flag on the existing item.
+        logger.info "Skipping message from publishing API: published with #{payload_version} but we've already received a message with #{item.publishing_api_payload_version}"
+      end
     else
-      Dimensions::Item.create_empty(content_id: content_id, base_path: base_path, locale: locale)
+      Dimensions::Item.create_empty(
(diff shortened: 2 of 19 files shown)


MatMoore:
I'd prefer to get the test in just because this is quite a big change, but I'll do this now. The other part can wait though.

MatMoore:
@KyleMacPherson I think this can be merged now

MatMoore:
I've just pushed another commit which changes the consumer to respect the order of the messages based on payload_version. 

That should be the last thing for this PR.

app/streams/publishing_api_consumer.rb
```diff
@@ -11,24 +11,56 @@ def process(message)
...
+    content_id = message.payload.fetch('content_id')
+    base_path = message.payload.fetch('base_path')
+    payload_version = message.payload.fetch('payload_version')
     locale = message.payload['locale'] || 'en'
 
     item = Dimensions::Item.by_natural_key(content_id: content_id, locale: locale)
+
     if item
-      handle_existing(item, base_path, message.delivery_info.routing_key)
+      existing_payload_version = item.publishing_api_payload_version || 0
```
pmanrubia: Not sure about this line. What is the intention here? to prevent running a migration to update all the payload versions to 0? Can we have versions with not `api_payload_version`? 

If the event always has a version, I would rather making it explicit by:
1. Running a migration to ensure that existing payloads a 0 version number.
2. Preventing nil version at the database/model level

If the event could have a nil version, this would entail that we would never be updating those events.

app/streams/publishing_api_consumer.rb
```diff
@@ -11,24 +11,56 @@ def process(message)
...
+
+      if payload_version > existing_payload_version
```
pmanrubia: It feels better to move this logic into the item and be more descriptive int he method name:

```ruby
  item.old_version?(new_payload)
```

app/streams/publishing_api_consumer.rb
```diff
@@ -11,24 +11,56 @@ def process(message)
...
+        handle_existing(
+          item: item,
+          base_path: base_path,
```
pmanrubia: Would it make sense to wrap the payload with an object? Something like:

```ruby
class NewVersionEvent
end
```

This would simplify the signature of our methods (there is no need to pass the `content_id`, `base_path`, `locale`...) so many times, and would allow being more explicit in our codebase.

app/streams/publishing_api_consumer.rb
```diff
@@ -11,24 +11,56 @@ def process(message)
...
+          routing_key: message.delivery_info.routing_key,
+          payload_version: payload_version
+        )
+      else
+        # Skip out of date and repeated messages from the publishing API.
```
pmanrubia: What about wrapping the comment in a string, and append it as additional information to the event? I think it is quite descriptive and would help when looking the logs. I am not a fan of big comments in code.

app/models/dimensions/item.rb
```diff
@@ -28,14 +48,15 @@ def quality_metrics_required?(attributes)
...
-  def self.create_empty(content_id:, base_path:, locale:)
+  def self.create_empty(content_id:, base_path:, locale:, payload_version:)
     create(
       content_id: content_id,
       base_path: base_path,
       locale: locale,
       latest: true,
       outdated: true,
-      outdated_at: Time.zone.now
+      outdated_at: Time.zone.now,
```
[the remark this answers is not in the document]
MatMoore: That would make more sense yeah. Do you think this is useful information to store? Because I was planning to get rid of the `outdated` flag once it's no longer needed by the ETL process.

app/streams/publishing_api_consumer.rb
MatMoore: Yeah this was to avoid a migration. I can do that.

app/streams/publishing_api_consumer.rb
MatMoore: I disagree - I think the model shouldn't have any knowledge of publishing API message formats. And this becomes a single comparison operation if I write a migration.

app/streams/publishing_api_consumer.rb
MatMoore: I'll think about this. Maybe a new object would help but `NewVersionEvent` is not intuitive to me.

app/streams/publishing_api_consumer.rb
MatMoore: To the log event you mean?

app/models/dimensions/item.rb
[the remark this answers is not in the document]
pmanrubia: Yes, true, we can get rid of it after we finish this work. 

In any case, it would useful to track when was the Content Item updated (the effective time of the new version), as we shouldn't rely on the fields (`updated_at` and `created_at` of AR). We could rename the attribute in the future.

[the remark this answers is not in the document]
MatMoore: I've looked into this and it seems we don't have a timestamp for the event itself.

The code that generates the message is here:
https://github.com/alphagov/publishing-api/blob/master/app/presenters/edition_presenter.rb#L37

For major/minor/links messages, we will have the `updated_at` of the edition itself, but that is a different event to whatever triggered the message we process. Unpublishing messages won't have a date at all.

app/streams/publishing_api_consumer.rb
MatMoore: do you mean something like this?

```ruby
        logger.info %Q{Skipping message from publishing API: published with #{payload_version} but we've already received a message with #{item.publishing_api_payload_version}.
This is to ensure that our "latest" record is actually the latest.
RabbitMQ guarantees at-least-once delivery when using acknowledgements, so messages can be duplicated, plus we can't guarantee the order of processing because we can have multiple consumers working in parallel.}
```

I'd prefer the original or not including the comment at all, I think this clutters up the logs for not much benefit. The reason I added a the comment is because it's non-obvious behaviour that may not make sense when looking at the code

app/streams/publishing_api_consumer.rb
MatMoore: I've extracted a new class `PublishingApiMessageProcessor` which the consumer now delegates to.

app/models/dimensions/item.rb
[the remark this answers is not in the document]
pmanrubia: mmm, this is unexpected. I would expect the events to be self-contained (including the time where they were generated). This is not in any case something to be sorted out in this PR, as this is external to the CPM.

app/streams/publishing_api_consumer.rb
pmanrubia: I was thinking in crafting a message using `heredocs`, and probably extracting a method with `what` we are actually doing; let's have a second iteration, as this is a big PR, ad a good step forward.

PR #657, opened 2018-04-19
Remove services folder

chao-xian:
This PR follows on from the restructuring of the code in #654. The code in `services` needed a little bit of refactoring before we could move the files, so was deemed out of scope for #654 as that was a lift and shift but gave us the basis for this PR.

The `ItemsService` file has been removed as it added an additional level of abstraction that appeared to be unnecessary.


chao-xian:
[(2) Refactor and move code in Services folder into the new app structure](https://trello.com/c/kddke6Jo/247-2-refactor-and-move-code-in-services-folder-into-the-new-app-structure)

app/etl/item/clients/google_analytics.rb
```diff
@@ -0,0 +1,13 @@
+require 'google/apis/analyticsreporting_v4'
+require 'googleauth'
+
+class Item::Clients::GoogleAnalytics
```
pmanrubia: What about moving this client to ETL/GA folder?

app/etl/item/clients/publishing_api.rb
```diff
@@ -0,0 +1,53 @@
+require 'gds_api/publishing_api_v2'
+
+class Item::Clients::PublishingAPI
```
pmanrubia: `PublishingAPI` is used in the `items` process. What are your thoughts about moving it there?

app/etl/item/clients/publishing_api.rb
```diff
@@ -0,0 +1,53 @@
...
+  PER_PAGE = 50
```
pmanrubia: If we have this constant `PER_PAGE`, do we also need the instance variable `@per_page`?

app/etl/item/clients/publishing_api.rb
chao-xian: Fixed in 6f512e2ec47733f83831527af491fcaac224590e

app/etl/item/clients/publishing_api.rb
chao-xian: Fixed in d625cb032655d13beb3378a33dca9b12cd702d81

chao-xian:
Thanks @pmanrubia - have addressed your comments.

PR #661, opened 2018-04-22
Fix etl rake task calling wrong classname

chao-xian:
The Master process was refactored previously and renamed but the
calling rake task was not updated.


chao-xian:
[BUG Fix rake task for importing data](https://trello.com/c/GGHDMzSm/250-bug-fix-rake-task-for-importing-data)

PR #238, opened 2017-07-24
Remove redundant `FactoryGirl` prefixes

alecgibson:

PR #664, opened 2018-04-23
Import data from the `Is this useful` metric via GA

KyleMacPherson:
The team has found, through user research, that adding in data to record user responses to 'is this page useful?' would be useful to help content publishers and managers make informed decisions about the content they publish.

The metrics come from GA events which are triggered when the user clicks one of the responses. We query GA for these events and follow the same pattern as we use when importing other GA metrics. We cannot however query for these metrics in the existing GA::Service query because, in order to identify which response the user has given, we need to query by event action as well as page path.

We follow the pattern of storing the GA data in a staging table called events_gas before inserting the data into our main facts_metrics table.

Change the name of a couple of files which were missed in the directory restructure PR.

This work is based on the findings of this spike


app/etl/ga/service.rb
```diff
@@ -1,6 +1,6 @@
 class GA::Service
   def client
-    @client ||= GoogleAnalytics::Client.new.build
+    @client ||= Clients::GoogleAnalytics.new.build
```
pmanrubia: This code need to be changed if you rebase with master.

KyleMacPherson:
[(3) Store yes/no satisfaction over time (depends #238)](https://trello.com/c/U84U4isJ/236-3-store-yes-no-satisfaction-over-time-depends-238)

db/migrate/20180418135558_add_is_this_useful_columns_to_events_gas.rb
```diff
@@ -0,0 +1,6 @@
+class AddIsThisUsefulColumnsToEventsGas < ActiveRecord::Migration[5.1]
+  def change
+    add_column :events_gas, :is_this_useful_yes, :integer
```
pmanrubia: I would default this columns to 0

db/migrate/20180418135632_add_is_this_useful_columns_to_facts_metrics.rb
```diff
@@ -0,0 +1,6 @@
+class AddIsThisUsefulColumnsToFactsMetrics < ActiveRecord::Migration[5.1]
+  def change
+    add_column :facts_metrics, :is_this_useful_yes, :integer
```
pmanrubia: I would default this column to 0 and not null

app/etl/ga/service.rb
```diff
@@ -14,8 +14,29 @@ def find_in_batches(date:, batch_size: 10_000)
       .each_slice(batch_size) { |slice| yield slice }
   end
 
+  def find_user_feedback_in_batches(date:, batch_size: 10_000)
```
pmanrubia: This class is doing too much. What about extracting a class to deal with user feedback events?

pmanrubia: Please, notice the `services` folder has changed. Please, split the specs as well.

app/etl/ga/user_feedback_processor.rb
```diff
@@ -0,0 +1,87 @@
...
+               dimensions_items.id
+        FROM events_gas, dimensions_items
+        WHERE page_path = base_path
+              AND events_gas.date = '#{date_to_s}'
+      ) AS s
+      WHERE dimensions_item_id = s.id AND dimensions_date_id = '#{date_to_s}'
+    SQL
+  end
+
+  def clean_up_query
```
pmanrubia: I would only clean up events that has been created on my process (having `is_this_useful_no` or `is_this_useful_yes` with a value != to 0) It would need a test as well.

app/etl/ga/user_feedback_processor.rb
```diff
@@ -0,0 +1,87 @@
...
+        page_path in (
+           SELECT base_path
+           FROM dimensions_items, facts_metrics
+           WHERE dimensions_items.id = facts_metrics.dimensions_item_id
+           AND facts_metrics.dimensions_date_id = '#{date_to_s}'
+        )
+    SQL
+  end
+
+  def fix_invalid_prefix_in_page_paths
```
pmanrubia: This seems duplicated with the views processor. Could you please extract into a class/concern?
Also the test could be extracted into a `Shared Example`

PR #665, opened 2018-04-23
Fix database healthcheck

chao-xian:
We were using ActiveRecord::Base.connected? which appears to be
deprecated. We have no tests at this point as it's not clear
how we would take a good approach to testing this endpoint.


chao-xian:
[(1) BUG api healthcheck shows database as critical when app is ok](https://trello.com/c/0R44n1NS/251-1-bug-api-healthcheck-shows-database-as-critical-when-app-is-ok)

PR #667, opened 2018-04-25
Content Schema Parser Integration Test

theKHutDeveloper:
[Trello: Ensure we extract content from all formats to apply Quality Metrics](https://trello.com/c/GXnKmtEP/218-2-ensure-we-extract-content-from-all-formats-to-apply-quality-metrics)

Add integration test to test the extraction process against the content schema JSON. To ensure that the extraction methods extract data that is a true reflection of what the Publishing API returns.

The integration tests revealed a number of potential errors with some of the schema extraction methods, which did not handle cases where keys specific to the schemas were not found. These have been fixed in this PR.

### DIFF
--- a/app/etl/item/content/parsers/contact.rb
+++ b/app/etl/item/content/parsers/contact.rb
@@ -2,9 +2,9 @@ class Item::Content::Parsers::Contact
   def parse(json)
     html = []
     html << json.dig('details', 'description')
-    html << get_email(json)
-    html << get_postal(json)
-    html << get_phone(json)
+    html << get_email(json) unless get_email(json).nil?
+    html << get_postal(json) unless get_postal(json).nil?
+    html << get_phone(json) unless get_phone(json).nil?
     html.join(" ")
   end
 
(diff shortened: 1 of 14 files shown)


chao-xian:
Paired with me. Note that unit tests for the handling of schemas extraction errors aren't in this PR. We've already discussed this and suggested it as something that will be added later.

app/etl/item/content/parsers/contact.rb
```diff
@@ -2,9 +2,12 @@ class Item::Content::Parsers::Contact
...
     html << json.dig('details', 'description')
-    html << get_email(json)
-    html << get_postal(json)
-    html << get_phone(json)
+    emails = get_email(json)
+    html << emails unless emails.nil?
+    postals = get_postal(json)
+    html << postals unless postals.nil?
+    phones = get_phone(json)
+    html << phones unless phones.nil?
```
KyleMacPherson: [minor] I think you could use `html << get_email(json) unless get_email(json).nil?` to keep the method a bit shorter and easier to read.

app/etl/item/content/parsers/parts.rb
```diff
@@ -1,11 +1,14 @@
 class Item::Content::Parsers::Parts
   def parse(json)
     html = []
-    json.dig("details", "parts").each do |part|
-      html << part["title"]
-      html << part["body"]
+    parts = json.dig("details", "parts")
+    unless parts.nil?
```
KyleMacPherson: [Minor] guard clause could be good here too

spec/integration/parser_process_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    schemas = GovukSchemas::Schema.all(schema_type: "frontend")
+
+    schema_list = []
+    schemas.each_value do |schema|
+      schema_list << schema
+    end
+    schema_list
+  end
+
+  def unused_schemas
```
KyleMacPherson: [Minor] I would expect `unused_schemas` to be called `invalid schemas` because of the invalid schemas variable and the valid_schema method. Any reason we have gone for `unused` rather than `invalid`?

spec/integration/parser_process_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      schemas.each do |schema|
+        expect(Item::Content::Parser.extract_content(schema)).to be_nil
+      end
+    end
+  end
+
+  def get_schema_collection
+    schemas = GovukSchemas::Schema.all(schema_type: "frontend")
+
+    schema_list = []
```
KyleMacPherson: [Minor] I think here we could just return schemas.values

app/etl/item/content/parsers/travel_advice_index.rb
```diff
@@ -2,10 +2,13 @@ class Item::Content::Parsers::TravelAdviceIndex
   def parse(json)
     html = []
     children = json.dig("links", "children")
-    children.each do |child|
-      html << child["country"]["name"]
+    unless children.nil?
```
KyleMacPherson: [Minor] Could potentially use a guard clause

PR #670, opened 2018-04-27
User feedback in sandbox

KyleMacPherson:
Add `is_this_useful` metrics to the sandbox UI ensuring they show up on all necessary pages as well as being available in charts and in the CSV export.


KyleMacPherson:
[(2) Expose yes/no satisfaction metric via the Sandbox (depends on 236)](https://trello.com/c/jq7SYXvR/255-2-expose-yes-no-satisfaction-metric-via-the-sandbox-depends-on-236)

PR #673, opened 2018-04-30
Add number of internal searches to API endpoint

KyleMacPherson:


chao-xian:
[(2) Expose Number of Internal Searches metric via the API (Depends #260)](https://trello.com/c/qkC41V9R/264-2-expose-number-of-internal-searches-metric-via-the-api-depends-260)

PR #675, opened 2018-04-30
Add base_path to InvalidSchemaError

theKHutDeveloper:
[Trello: Store the base_path when logging InvalidSchemaErrors in Sentry](https://trello.com/c/6ARiy1up/273-store-the-basepath-when-logging-invalidschemaerrors-in-sentry)

This commit is part of a story to 'Support all schemas when parsing
content items' - [Trello](https://trello.com/c/Pjp3RUGu/241-3-support-all-schemas-when-parsing-content-items-to-extract-content). Currently content schemas wer are not extracting
are being recorded as InvalidSchemaErrors.

We need to make a decision as to whether we want to store the
content from these 'Invalid' content schemas or not. Storing the
base_path will help us make decisions as we can access the specific
content relating to the error message.

### DIFF
--- a/app/etl/item/content/parser.rb
+++ b/app/etl/item/content/parser.rb
@@ -11,11 +11,12 @@ class Item::Content::Parser
 
   def extract_content(json)
     schema = json.dig("schema_name")
+    base_path = json.dig("base_path")
     parser = for_schema(schema)
     if parser.present?
       parse_html parser.parse(json)
     else
-      GovukError.notify(InvalidSchemaError.new("Schema does not exist: #{schema}"))
+      GovukError.notify(InvalidSchemaError.new("Schema does not exist: #{schema}"), extra: { base_path: base_path.to_s })
 
       nil
     end
--- a/spec/etl/item/content/parser_spec.rb
+++ b/spec/etl/item/content/parser_spec.rb
@@ -465,21 +465,34 @@ RSpec.describe Item::Content::Parser do
       end
     end
 
-    context 'when invalid schema' do
-      before do
-        expect(GovukError).to receive(:notify).with(an_instance_of(InvalidSchemaError))
+    context "when invalid schema" do
+      describe "has no schema_name and no base_path" do
+        it "raises an InvalidSchemaError and returns nil" do
+          subject.extract_content document_type: 'answer'
+          expect(GovukError).to receive(:notify).with(InvalidSchemaError.new("Schema does not exist: "), extra: { base_path: "" })
+          expect(subject.extract_content(document_type: 'answer')).to be_nil
+        end
       end
 
-      it 'Logs an error InvalidSchemaError if json schema_name is not known' do
-        subject.extract_content schema_name: 'blah'
+      describe "has an unknown schema_name but no base_path" do
+        it "logs InvalidSchemaError with the schema_name" do
+          json = { schema_name: "blah", links: {} }
+
+          expect(GovukError).to receive(:notify).with(InvalidSchemaError.new("Schema does not exist: blah"), extra: { base_path: "" })
+
+          result = subject.extract_content(json.deep_stringify_keys)
+          expect(result).to be_nil
+        end
       end
 
-      it 'raises InvalidSchemaError if non-empty json does not have a schema_name' do
-        subject.extract_content document_type: 'answer'
-      end
-
-      it 'returns nil' do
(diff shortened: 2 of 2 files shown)


spec/etl/item/content/parser_spec.rb
```diff
@@ -465,21 +465,38 @@ def build_raw_json(body:, schema_name:)
...
     end
 
-    context 'when invalid schema' do
-      before do
-        expect(GovukError).to receive(:notify).with(an_instance_of(InvalidSchemaError))
-      end
+    context "when invalid schema" do
+      describe "has no schema_name and no base_path" do
+        before do
+          expect(GovukError).to receive(:notify).with(InvalidSchemaError.new("Schema does not exist: "), extra: { base_path: "" })
```
MatMoore: These tests are fine as is, but I would structure these two a bit differently.

The reason being that tests are easier for me to understand if they follow a very simple structure

- setup
- do a thing
- assertions

The second two are really important for understanding what a test is doing, so I'd rather have them in the body of the test, even if it means duplicating some code.

With `expect(foo).to receive(:bar)` the order is reversed but I still think of it as an assertion rather than setup, so I'd rather inline this like in the tests below. Otherwise it's easy to miss the `before` and think the test isn't asserting anything.

The other thing that seems strange to me is the test names, because "it raised InvalidSchemaError" and "it returns nil" are mutually exclusive if you're talking about the same thing. The problem is there are two 'it's!

There is an error that gets raised internally but nothing outside the class depends on that behaviour - the thing you care about is really the side effect of calling `GovukError` rather than the exception itself, so the name of the spec should reflect that.

Maybe something like this?

```ruby
context "when the content has no schema_name and no base path" do
  describe '#extract_content' do
    it "returns nil and notifies us about the invalid schema" do
       ...
    end
  end
end
```

spec/etl/item/content/parser_spec.rb
```diff
@@ -465,21 +465,38 @@ def build_raw_json(body:, schema_name:)
...
       end
 
-      it 'raises InvalidSchemaError if non-empty json does not have a schema_name' do
-        subject.extract_content document_type: 'answer'
+      describe "has an unknown schema_name but no base_path" do
+        it "logs InvalidSchemaError with the schema_name" do
+          json = { schema_name: "blah", links: {} }
+          subject.extract_content(json.deep_stringify_keys)
+          expect(GovukError).to receive(:notify).with(InvalidSchemaError.new("Schema does not exist: blah"), extra: { base_path: "" })
+          expect(subject.extract_content(json.deep_stringify_keys)).to be_nil
```
MatMoore: Here `subject.extract_content` is called twice. Would be clearer like this:

```ruby
        it "logs InvalidSchemaError with the schema_name" do
          json = { schema_name: "blah", links: {} }

          expect(GovukError).to receive(:notify).with(InvalidSchemaError.new("Schema does not exist: blah"), extra: { base_path: "" })
          result = subject.extract_content(json.deep_stringify_keys)

          expect(result).to be_nil
        end
```

spec/integration/parser_process_spec.rb
```diff
@@ -31,6 +31,7 @@
     it 'returns nil for unused schemas' do
       schemas = invalid_schemas
       schemas.each do |schema|
+        Item::Content::Parser.extract_content(schema)
```
MatMoore: Is this needed? It's called in the following line

app/etl/item/content/parser.rb
```diff
@@ -11,11 +11,12 @@ def self.extract_content(*args)
 
   def extract_content(json)
     schema = json.dig("schema_name")
+    base_path = json.dig("base_path")
     parser = for_schema(schema)
     if parser.present?
       parse_html parser.parse(json)
     else
-      GovukError.notify(InvalidSchemaError.new("Schema does not exist: #{schema}"))
+      GovukError.notify(InvalidSchemaError.new("Schema does not exist: #{schema}"), extra: { base_path: base_path.to_s })
```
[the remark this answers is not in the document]
pmanrubia: @MatMoore in theory we would not expect to have these messages. 
We should be able to handle all of the schemas, so sending an error here is a way for us to know that we need to handle it. We aim to have a 0 errors in Sentry.

spec/integration/parser_process_spec.rb
theKHutDeveloper: Good spot I forget to remove that line

PR #676, opened 2018-04-30
Store internal searches metrics

chao-xian:
Track the number of times per page that the internal search within GOV.UK is being used.

This PR follows the approach of #664. We now reach a point where we have too much duplication, which I will address in a subsequent PR to keep the work separate.


chao-xian:
[(3) Store the number of internal searches from a page in data warehouse](https://trello.com/c/DsSyxc3O/260-3-store-the-number-of-internal-searches-from-a-page-in-data-warehouse)

PR #692, opened 2018-05-04
Extract content from 4 more schemas

theKHutDeveloper:
[Trello: Create parsers for schemas (batch 2)](https://trello.com/c/S70veFeg/287-2-create-parsers-for-schemas-batch-2)

We have made the decision that we DO want to extract content from the following schemas as they can be used to generate quality metrics:

Schemas
---------
mainstream_browse_page
organisation
service_manual_homepage
step_by_step_nav

This is part of EPIC [Trello: Support all schemas when parsing content items to extract content](https://trello.com/c/Pjp3RUGu/241-3-epic-support-all-schemas-when-parsing-content-items-to-extract-content)

### DIFF
--- a/app/etl/item/content/parsers/mainstream.rb
+++ b/app/etl/item/content/parsers/mainstream.rb
@@ -1,0 +1,27 @@
+class Item::Content::Parsers::Mainstream
+  def parse(json)
+    html = []
+    html << json.dig("title") unless json.dig("title").nil?
+    html << json.dig("description") unless json.dig("description").nil?
+
+    children = json.dig("links", "children")
+    if children.present?
+      children.each do |child|
+        html << child["title"]
+      end
+    end
+
+    related_topics = json.dig("links", "related_topics")
+    if related_topics.present?
+      related_topics.each do |topic|
+        html << topic["title"]
+      end
+    end
+
+    html.compact.join(" ")
+  end
+
+  def schemas
+    %w[mainstream_browse_page]
+  end
+end
--- a/app/etl/item/content/parsers/service_manual_homepage.rb
+++ b/app/etl/item/content/parsers/service_manual_homepage.rb
@@ -1,0 +1,21 @@
+class Item::Content::Parsers::ServiceManualHomepage
+  def parse(json)
+    html = []
+    html << json.dig("title")
+    html << json.dig("description")
+
+    children = json.dig("links", "children")
+    if children.present?
+      children.each do |child|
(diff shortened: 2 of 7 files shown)


app/etl/item/content/parsers/mainstream.rb
```diff
@@ -0,0 +1,27 @@
...
+        html << topic["title"] unless topic["title"].nil?
+      end
+    end
+
+    html.join(" ") unless html.empty?
+  end
+
+  def schemas
+    %w[mainstream_browse_page]
+  end
```
pmanrubia: What are your thoughts about?

- using `if` instead of `unless`
- removing `.nil?` and use `compact`
- remove `unless html.empty?`

```ruby
  def parse(json)
    html = []
    html << json.dig("title")
    html << json.dig("description")

    children = json.dig("links", "children")
    if children.present?
      children.each do |child|
        html << child["title"]
      end
    end

    related_topics = json.dig("links", "related_topics")
    if related_topics.present?
      related_topics.each do |topic|
        html << topic["title"]
      end
    end

    html.compact.join(" ")
  end
```

app/etl/item/content/parsers/service_manual_homepage.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    children = json.dig("links", "children")
+    unless children.nil?
+      children.each do |child|
+        html << child["title"] unless child["title"].nil?
+        html << child["description"] unless child["description"].nil?
+      end
+    end
+
+    html.join(" ") unless html.empty?
```
pmanrubia: See comment above about DRY-ing it up

app/etl/item/content/parsers/step_by_step.rb
```diff
@@ -0,0 +1,37 @@
...
+              lists = content["contents"]
+              unless lists.nil?
+                lists.each do |list|
+                  html << list["text"] unless list["text"].nil?
+                end
+              end
+            end
+          end
+        end
+      end
```
pmanrubia: See comment above about DRY-ing it up

app/etl/item/content/parsers/mainstream.rb
theKHutDeveloper: @pmanrubia I agree that `parser_spec` is getting too large. I will work on this in a new PR
@MatMoore I took the view `if we can extract the data we should`, if it has no value we don't have to use it or we can remove it.

PR #693, opened 2018-05-04
Add all quality metrics to the Sandbox

pmanrubia:
We have also removed the summary section as it was 
introducing big delays in the queries and it wasn't a real
need for users.

In the same go we have refactored the tests, and simplified
the layout of the features.

paired with @chao-xian


chao-xian:

PR #695, opened 2018-05-09
Remove metadata from Facts Editions

pmanrubia:
We should't have metadata in the facts table.


chao-xian:

PR #701, opened 2018-05-14
Add additional ga metrics

KyleMacPherson:
The team decided we would add additional GA metrics, namely `entrances`, `exits`, `bounce_rate` and `avg_time_on_page` to help understand user behaviour on a page to page basis.

For efficiency reasons we have added the query for this data to our existing `views` query within the `ViewsService` and `ViewsProcessor`. Due to the change in the data that these files are working with I have renamed them to `ViewsAndNavigation` processor and service respectively.


KyleMacPherson:
[(3) Add additional GA metrics to the DW](https://trello.com/c/hJiJdlZG/300-3-add-additional-ga-metrics-to-the-dw)

PR #704, opened 2018-05-14
Fix Bug: master process no longer running

pmanrubia:
[Trello card](https://trello.com/c/NY0893lD/315-2-fix-etl-master-process-not-running-on-the-nights)

### What 

[We no longer check for the presence of a date](https://deploy.publishing.service.gov.uk/job/Deploy_App/5824/console) to validate that the
process didn't run already. We now rely on the existence of metrics
which seems more consistent with the nature of the process.

### Description

To prevent metrics duplication the master process checks that we 
have not Date entry in the `Dimensions::Date` for the given day. But, 
since we have started tracking editions on a daily basis, the date is 
being created already before the `master` process runs which prevent 
the master process from running.


spec/etl/master/master_processor_spec.rb
```diff
@@ -18,10 +18,19 @@
     allow(Master::MetricsProcessor).to receive(:process)
   end
 
-  it 'does not process if already processed for date' do
-    create(:dimensions_date, date: Date.yesterday)
+  describe 'runs only once per day' do
+    it 'raises `DuplicateDateError` if there are already metrics for the day' do
+      ddate = create(:dimensions_date, date: Date.yesterday)
```
chao-xian: Not sure but this looks like a typo with `ddate`? Otherwise GTG.

pmanrubia: It was a shorter way for `dimensions_date`, but I have updated it as it is obviously not clear enough :)

PR #706, opened 2018-05-14
Add schemas to ignore

theKHutDeveloper:
[Trello: Make a decision on the remaining schemas](https://trello.com/c/2enowzan/317-make-a-decision-on-the-remaining-schemas)

There are a number of schemas that we do not want to extract data from as they will not give us any useful content for quality metrics.

This PR adds the following schemas to the list that we do not want to extract content from: 

organisations_homepage
topic
world_location
placeholder_organisation
placeholder_world_location
placeholder_working_group
placeholder_topical_event
placeholder_policy_area 
placeholder_ministerial_role

Now that we have added these schemas we should not be seeing any InvalidSchemaErrors for any of these schemas.

This PR is part of the [EPIC - Trello: Support all schemas when parsing content items to extract content](https://trello.com/c/Pjp3RUGu/241-3-epic-support-all-schemas-when-parsing-content-items-to-extract-content)

### DIFF
--- a/app/etl/item/content/parsers/no_content.rb
+++ b/app/etl/item/content/parsers/no_content.rb
@@ -10,14 +10,23 @@ class Item::Content::Parsers::NoContent
       external_content
       generic
       homepage
+      organisations_homepage
       person
       placeholder_corporate_information_page
+      placeholder_ministerial_role
+      placeholder_organisation
+      placeholder_policy_area
+      placeholder_topical_event
+      placeholer_working_group
+      placeholder_world_location
       placehold_worldwide_organisation
       placeholder_person
       placeholder
       policy
-      special_route
       redirect
+      special_route
+      topic
+      world_location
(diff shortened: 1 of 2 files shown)


app/etl/item/content/parsers/no_content.rb
```diff
@@ -10,14 +10,23 @@ def schemas
...
       homepage
+      organisations_homepage
       person
       placeholder_corporate_information_page
+      placeholder_ministerial_role
+      placeholder_organisation
+      placeholder_policy_area
+      placeholder_topical_event
+      placeholer_working_group
+      placeholder_world_location
```
cbaines: This is probably a bigger question, but these look to me like document types, rather than the names of schemas (e.g. I can't see a `placeholder_world_location` schema, I imagine it just uses the `placeholder` schema)?

PR #709, opened 2018-05-15
Refactor editions in sandbox controller

chao-xian:
This PR is mostly refactoring to start using the altered schema from [previous work by @MatMoore](https://trello.com/c/SWlCL8xI) in the sandbox interface.

It is a follow on from #686 and part of this [Epic](https://trello.com/c/igfOJois)

The scope of the story was only for updating the sandbox, but we could also refactor the `metrics_controller.rb` to use the extracted code in the `Reports::Series` class. Even more work would be required and this PR is already encompassing a lot of changes, so I've left that out for a future PR.


chao-xian:

spec/domain/reports/series_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+      metric2 = create(:metric, dimensions_item: item1, dimensions_date: day1)
+      metric3 = create(:metric, dimensions_item: item1, dimensions_date: day2)
+      create(:metric, dimensions_item: item2, dimensions_date: day1)
+      create(:metric, dimensions_item: item2, dimensions_date: day2)
+
+      results = described_class.new.by_base_path('/path1').run
+
+      expect(results).to match_array([metric1, metric2, metric3])
+    end
+  end
```
pmanrubia: What about testing the actual `chaining` so we are sure that it is not broken?

PR #710, opened 2018-05-16
Add more GA metrics to the sandbox

KyleMacPherson:
Add `entrances`, `exits`, `bounce_rate` and `avg_time_on_page` to the sandbox ui and add necessary tests.


KyleMacPherson:
[(2) Show new GA metrics in the Sandbox DEPENDS ON #300](https://trello.com/c/0CISsQS1/301-2-show-new-ga-metrics-in-the-sandbox-depends-on-300)

PR #714, opened 2018-05-22
Update API to filter by base_path

theKHutDeveloper:
We are now able to obtain metrics by querying the base_path instead of having to know the content_id of the base_path by which to filter.

We have added the `format_base_path_param` method to the MetricsController to ensure that the base_path is in the format that is expected by the database, so that we can query DimensionsItems by base_path.


theKHutDeveloper:
[(3) Update API to filter by base-path](https://trello.com/c/gU0lwEBx/357-3-update-api-to-filter-by-base-path)

PR #715, opened 2018-05-24
Add stubbed consumer for bulk import RabbitMQ

chao-xian:
We added a new RabbitMQ queue but didn't set up a consumer. We aren't
ready yet for the consumer to do anything, but we are stubbing it
so that we don't get Icinga alerts that check that message queues
aren't being consumed.


chao-xian:

PR #717, opened 2018-05-24
Add missing rake task

chao-xian:
Would help if the rake task to start a message queue consumer was
included.


chao-xian:

PR #728, opened 2018-06-04
Preload items multi page

pmanrubia:
[Trello card](https://trello.com/c/m3vzLnR5/372-3-tech-debt-use-only-publishing-api-to-preload-items)

Removes dependency with PubilshingAPI to iterate through all
the Content Items that are in `live` state.

It also perform the following refactorings:

- Remove `requires` in test that are not needed
- Split tests in consumer / processor (for publishing-api events)


spec/tasks/publishing_api_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
+
+RSpec.describe 'rake publishing_api:*', type: task do
+  let(:consumer) { double('message_consumer', run: nil) }
+
+  it 'start listener for PublishingAPI events' do
```
MatMoore: I would prefer to test the processor more extensively than test these rake tasks. The rake tasks' only responsibility is to make a single function call, and this code will rarely change, whereas the processor itself will change a lot.

I think if the consumer is not running in production, the most likely scenario is it crashed because of an unhandled exception.

MatMoore: also I think there's a typo here: "start listener" should be "starts listening"

app/streams/publishing_api/message_processor.rb
```diff
@@ -56,7 +56,7 @@ def handle_existing(item)
   def handle_new_version(item)
     new_item = item.copy_to_new_version!(base_path: base_path, payload_version: payload_version)
 
-    Items::Jobs::ImportContentDetailsJob.perform_async(new_item.id, current_date.day, current_date.month, current_date.year)
+    Items::Importers::ContentDetails.run(new_item.id, current_date.day, current_date.month, current_date.year)
```
MatMoore: I'm happy with this change but just wondering if we should do the same with the quality metrics job.

This would mean the unit of work is always one content item, and if everything fails we retry that whole content item from scratch.

I think this makes the system simpler to reason about overall.

spec/tasks/publishing_api_spec.rb
pmanrubia: I find value in unit testing the rake task, because as you say, it main responsibility is to make a function call. I like it because we are being proactive towards preventing the Jenkins job from failing and having to be reactive in fixing it.

When it comes to the testing, agree that we need an integration tests for the processor, but not sure if this should live here, as this is just an entry point for the process. I would be really happy to create another PR to add an integration test that cover the basic scenarios.

app/streams/publishing_api/message_processor.rb
pmanrubia: What are your thoughts about dealing with errors on the PAAS service that we are invoking to retrieve the quality metrics. The system could be down, and we would need rollback all changes to the dataware dimensions...

MatMoore: I'm not sure yet. I think this applies anywhere we do something that has the potential to leave the data in an incomplete state.

One way might be to have an explicit [staging area](https://en.wikipedia.org/wiki/Staging_(data)), then at the end of processing an item you can store the complete data in one transaction.

pmanrubia: As discussed offline, we will have a more in-depth discussion about it in the next developer's session.

PR #729, opened 2018-06-04
Document discussions we've had about stream processing

MatMoore:
This documents some stuff we discussed as a team a couple of weeks ago. I want to capture some of the decisions as the stories we're working on will be difficult to understand without some context.

I've attempted to draw what I think the data flow of this process will look like once we've made these changes. If part of this doesn't make sense, let me know, as I'm about to try implementing it...

This doesn't include any decisions about error handling and asynchronous processing. Some of Pablo's work in https://github.com/alphagov/content-performance-manager/pull/728 has touched on this but we're going to follow up on it later.

[Rendered version](https://github.com/alphagov/content-performance-manager/blob/c1ce0ec416a7cc28b5bbb64d695b150b8f1bae48/doc/arch/adr-009-treat-parts-separately.md)

Trello: https://trello.com/c/m3vzLnR5/372-3-tech-debt-use-only-publishing-api-to-preload-items and https://trello.com/c/ZaHM1qZD/381-5-update-etl-to-handle-sub-pages-for-guides


doc/arch/adr-009-treat-parts-separately.md
```diff
@@ -0,0 +1,53 @@
...
+We should extract all the information we need from the publishing API message that triggers the process.
+
+### Data flow
+This diagram shows the ideal data flow for ingesting content into the data warehouse.
+
+The outputs are `Dimension::Item` records (identified by URL and version) and a `Facts::Edition` record for each one, which stores the edition metrics.
+
+The content item is split into parts at the beginning of the pipeline, based on the `details.parts` object in the content item JSON.
+
+For each part, we need to work out if we're updating an item or creating a new one. Previously we made this decision for the combined item. New items are created if the publishing API message represents a `major`, `minor` or `unpublish` update. If the publishing API message is a `link` update we just update the metadata on the existing item.
```
pmanrubia: If I remember correctly, on a new event we will mark all the existing parts of the Content Item as `latest=false` and then treat it as a new edition (recreating all the part)? Basically the same as it is happening now, but considering that the content Item has multiple parts. 
Not sure if this is clear enough, or If I am misunderstanding anything...

MatMoore: Ah ok. I'll add something in describing this. Does this mean we effectively ignore the content hash for multi part documents?

pmanrubia: I would say so. The hash only apply to the content of each part, I think.

MatMoore:
@pmanrubia OK sounds good - I'll update this one to cover just the consuming from publishing api part.

MatMoore:
@pmanrubia  I think I might have interpreted your comment as the opposite of what you meant 🤔 

But I've come up with this:
 https://github.com/alphagov/content-performance-manager/blob/723c6bdf34635b804dc9685f9900bc34aef64073/doc/arch/adr-009-consume-from-publishing-api.md

It now focuses on the refactor of the message handler, but I've included a short section about how it affects the multi-part stuff.

If you want we could move that section to a followup ADR?

pmanrubia:
It is fine, we can document all on this PR.

PR #730, opened 2018-06-04
Make bulk load consumer preload data

MatMoore:
Paired with @chao-xian

This is the first step towards removing the content store integration, and processing multi part documents.

We're going to implement multi parts in the bulk load process first, but we will refactor as we go so that bulk loading doesn't diverge too much from the regular processing.

We've introduced a new class so we can easily extract data from the message queue representation of a content item, instead of the content store one. The next step is to pull out some logic from the existing parsers so that we can tell if an item is multi part or not.

### DIFF
--- a/app/streams/publishing_api/content_item.rb
+++ b/app/streams/publishing_api/content_item.rb
@@ -1,0 +1,45 @@
+class PublishingAPI::ContentItem
+  attr_reader :content_id
+  attr_reader :base_path
+  attr_reader :payload_version
+  attr_reader :locale
+  attr_reader :metadata
+  attr_reader :details
+  attr_reader :links
+
+  def self.parse_message(message_payload)
+    content_id = message_payload.fetch('content_id')
+    base_path = message_payload.fetch('base_path')
+    payload_version = message_payload.fetch('payload_version')
+    locale = message_payload['locale'] || 'en'
+    metadata = Item::Metadata::Parsers::Metadata.parse(message_payload)
+    details = message_payload.fetch('details')
+    links = message_payload.fetch('expanded_links')
+
+    new(
+      content_id: content_id,
+      base_path: base_path,
+      payload_version: payload_version,
+      locale: locale,
+      links: links,
+      details: details,
+      metadata: metadata,
+    )
+  end
+
+  def initialize(content_id:, base_path:, payload_version:, locale:, links:, details:, metadata:)
+    @content_id = content_id
+    @base_path = base_path
+    @payload_version = payload_version
+    @locale = locale
+    @links = links
+    @details = details
+    @metadata = metadata
+  end
+
+private
+
+  def current_date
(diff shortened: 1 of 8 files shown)


app/streams/publishing_api/message_processor.rb
```diff
@@ -22,24 +19,21 @@ def process
         # RabbitMQ guarantees at-least-once delivery when using acknowledgements,
         # so messages can be duplicated, plus we can't guarantee the order of
         # processing because we can have multiple consumers working in parallel.
-        logger.info "Skipping message from publishing API: published with #{payload_version} but we've already received a message with #{item.publishing_api_payload_version}"
+        logger.info "Skipping message from publishing API: published with #{incoming_content_item.payload_version} but we've already received a message with #{item.publishing_api_payload_version}"
       end
     else
       Dimensions::Item.create_empty(
```
MatMoore: This case seems wrong as it doesn't run the importer

chao-xian:

app/streams/publishing_api/content_item.rb
```diff
@@ -0,0 +1,45 @@
...
+  attr_reader :metadata
+  attr_reader :details
+  attr_reader :links
+
+  def self.parse_message(message_payload)
+    content_id = message_payload.fetch('content_id')
+    base_path = message_payload.fetch('base_path')
+    payload_version = message_payload.fetch('payload_version')
+    locale = message_payload['locale'] || 'en'
+    metadata = Item::Metadata::Parsers::Metadata.parse(message_payload)
```
pmanrubia: If we are getting the metadata at this point I would move that code inside the `streams` folder, so we don't couple the stream to `etl/item`

app/streams/publishing_api/content_item.rb
```diff
@@ -0,0 +1,45 @@
+class PublishingAPI::ContentItem
```
pmanrubia: This is great 👍

app/streams/publishing_api/message_processor.rb
chao-xian: Have this tracked in https://trello.com/c/Yqb5YxIq/449-story-template-take-copy now.

app/streams/publishing_api/content_item.rb
MatMoore: Move the parser you mean? That sounds like a good idea, I think everything under `etl/item` and `etl/items` is actually only used by the `streams` code.

pmanrubia: I mean move only the `Metadata` parser. The other parsers like PDF counter... I would leave them in the `etl/item`...

MatMoore: I'm going to do this in a separate PR because I need to refactor `MessageProcessor` more before this can be removed from the existing `Parser` class.

PR #733, opened 2018-06-06
Fix bug causing no method errors when running content_quality_service

KyleMacPherson:
If item.get_content returns nil then it causes an error to be thrown during our quality_metrics ETL process. This change
inserts a guard clause which tops the content_quality_service from running if item.get_content returns nil.

### DIFF
--- a/app/etl/items/importers/quality_metrics.rb
+++ b/app/etl/items/importers/quality_metrics.rb
@@ -14,9 +14,10 @@ class Items::Importers::QualityMetrics
 
   def run
     item = Dimensions::Item.find(@item_id)
-
-    quality_metrics = content_quality_service.run(item.get_content)
-    item.facts_edition.update(**quality_metrics)
+    if item.get_content
+      quality_metrics = content_quality_service.run(item.get_content)
+      item.facts_edition.update(**quality_metrics)
+    end
   rescue InvalidSchemaError
     do_nothing
   end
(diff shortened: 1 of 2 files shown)


spec/etl/items/importers/quality_metrics_spec.rb
```diff
@@ -56,6 +56,11 @@
       )
     end
 
+    it 'returns without running content_quality_service if #get_content returns nil' do
+      allow_any_instance_of(Dimensions::Item).to receive(:get_content).and_return(nil)
```
pmanrubia: Instead of stubbing this method, I would rather to create an Item that has no content. This will help in refactorings as we don't mock the implementation.

We could use this:

```ruby
    payload = GovukSchemas::RandomExample.for_schema(notification_schema: "detailed_guide")
    ...
```

We can even modify FactoryGirl to create a dimension item with no content on a explicit way.
What do you think?

app/etl/items/importers/quality_metrics.rb
```diff
@@ -14,7 +14,7 @@ def initialize(item_id, _ = {})
 
   def run
     item = Dimensions::Item.find(@item_id)
-
+    return unless item.get_content
     quality_metrics = content_quality_service.run(item.get_content)
```
pmanrubia: In general it is a bit better to have only one exit point.
What about expressing this logic with:

```ruby
if item.get_content
  # do something
end
```

PR #734, opened 2018-06-07
Raise error if content quality service does not return 200 response code

KyleMacPherson:
Use GovukError.notify to to notify Sentry of the error.

### DIFF
--- a/app/etl/item/content_quality_service.rb
+++ b/app/etl/item/content_quality_service.rb
@@ -16,6 +16,7 @@ private
       body: { content: content }.to_json,
       headers: { 'Content-Type' => 'application/json' }
     )
+    raise QualityMetricsError("response body: #{response.body}") unless response.code == 200
     response.parsed_response
   end
 
@@ -42,4 +43,6 @@ private
   def count_metric(response, metric_name)
     response.dig(metric_name, 'count') || 0
   end
+
+  class QualityMetricsError < StandardError; end
 end
--- a/app/etl/items/jobs/import_quality_metrics_job.rb
+++ b/app/etl/items/jobs/import_quality_metrics_job.rb
@@ -3,5 +3,8 @@ class Items::Jobs::ImportQualityMetricsJob < Items::Jobs::ApplicationJob
 
   def run(*args)
     Items::Importers::QualityMetrics.run(*args)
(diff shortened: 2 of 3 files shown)


KyleMacPherson:
[(2) Fix: Undefined method `dig' for String](https://trello.com/c/u9AWrC3a/397-2-fix-undefined-method-dig-for-string)

app/etl/item/content_quality_service.rb
```diff
@@ -42,4 +43,6 @@ def convert_results(response, content)
   def count_metric(response, metric_name)
     response.dig(metric_name, 'count') || 0
   end
+
+  class NoContentError < StandardError; end
```
pmanrubia: Not sure...  If response code is not `200` we don't know what the error is. I would not named it `NoContentError`. I'd rather rename it to `QualityMetricsError` (or similar), and I would add the error to the exception so we can track it. What do you think?

app/etl/items/jobs/import_quality_metrics_job.rb
```diff
@@ -3,5 +3,13 @@ class Items::Jobs::ImportQualityMetricsJob < Items::Jobs::ApplicationJob
 
   def run(*args)
     Items::Importers::QualityMetrics.run(*args)
+  rescue StandardError => e
+    GovukError.notify(
```
pmanrubia: Could you please use one line for notifying the error?

spec/etl/items/jobs/import_quality_metrics_job_spec.rb
```diff
@@ -0,0 +1,7 @@
+RSpec.describe Items::Jobs::ImportQualityMetricsJob do
```
pmanrubia: 👍

app/etl/item/content_quality_service.rb
```diff
@@ -16,6 +16,7 @@ def fetch(content)
       body: { content: content }.to_json,
       headers: { 'Content-Type' => 'application/json' }
     )
+    raise QualityMetricsError("response body: #{response.body}") unless response.code == 200
```
KyleMacPherson: @pmanrubia wasn't sure exactly what you meant. Something like this?

PR #736, opened 2018-06-08
Remove redundant `Reports::Series.for_en` method

KyleMacPherson:
Reports::Series already defaults to english so there is no need for an explicit for_en method


KyleMacPherson:
[(2) English locale is the default in Sandbox and API](https://trello.com/c/ESHSVyfD/402-2-english-locale-is-the-default-in-sandbox-and-api)

PR #738, opened 2018-06-12
Move `with_editions_metrics` logic to series model

KyleMacPherson:
In order to give the controller less responsibility, and keep the business logic within the model,
this PR moves the logic around whether or not to provide edition metric to the Reports::Series model.

### DIFF
--- a/app/domain/reports/series.rb
+++ b/app/domain/reports/series.rb
@@ -6,12 +6,6 @@ class Reports::Series
     self
   end
 
-  def with_edition_metrics
-    @with_edition_metrics = true
-
-    self
-  end
-
   def by_organisation_id(org_id)
     @org_id = org_id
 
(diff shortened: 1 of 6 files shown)


KyleMacPherson:
[(5) Simplify how we query series via Report::Series](https://trello.com/c/ImEr60hE/401-5-simplify-how-we-query-series-via-reportseries)

app/domain/reports/series.rb
```diff
@@ -1,17 +1,22 @@
...
 
-  def with_edition_metrics
-    @with_edition_metrics = true
+  def with_metrics(metrics)
+    @metrics = metrics
 
     self
   end
 
+  def include_edition_metrics?
```
pmanrubia: Isn't this a private method? It shouldn't be public.

spec/domain/reports/series_spec.rb
```diff
@@ -129,4 +129,24 @@
       expect(described_class.new.content_items).to match_array([item1, item2])
     end
   end
+
+  describe '#include_edition_metrics' do
```
pmanrubia: The method is private so we shouldn't test it.
I would create a fixture of data so the code need to check if the metrics are editions or not, then run the code, and check each option.

PR #739, opened 2018-06-12
Preload items: remove content store dependency

pmanrubia:
Paired with @chao-xian 

[Trello card](https://trello.com/c/m3vzLnR5/372-8-use-only-publishing-api-to-preload-items-basepath)

* Removes dependency with Content Store by subscribing to bulk publish in PublishingAPI.
* Simplifies the process of handling messages that comes from the PublishingAPI
* Add constraints to the Database so we ensure that we have no data inconsistencies.
* Calculates metadata attributes when parsing the content because we have all the information available in the payload of the message.


chao-xian:
[(8) Use only publishing api to preload items (basepath)](https://trello.com/c/m3vzLnR5/372-8-use-only-publishing-api-to-preload-items-basepath)

PR #746, opened 2018-06-15
Log in Sentry 'NoMethodError' occurring in Content Parser

theKHutDeveloper:
[Trello: Fix: NoMethodError app/etl/item/content/parser.rb in parse_html ](https://trello.com/c/zaA0hdSv/424-2-fix-nomethoderror-app-etl-item-content-parserrb-in-parsehtml) 

To be able to fix the NoMethodError we need to understand why this error is occurring, by logging these events we will be in a position to generate a solution.


app/etl/item/content/parser.rb
```diff
@@ -14,7 +14,11 @@ def extract_content(json)
     base_path = json.dig("base_path")
     parser = for_schema(schema)
     if parser.present?
-      parse_html parser.parse(json)
+      begin
+        parse_html parser.parse(json)
+      rescue NoMethodError => e
```
pmanrubia: Would you mind rescuing `StandardError`?

PR #752, opened 2018-06-18
Fix parsing body array

chao-xian:
Allow for parsing of body returned as array

The content schemas allow for body content to return as a string or
as an array for multiple content types [1] but we had only assumed the
case for strings. This caused issues when we tried to call string
methods on an array.

This was sending a lot of error messages in our logs.

[1]
https://github.com/alphagov/govuk-content-schemas/blob/master/formats/shared/definitions/publishing_api_base.jsonnet#L125
and
https://github.com/alphagov/govuk-content-schemas/blob/master/formats/shared/definitions/publishing_api_base.jsonnet#L129

### DIFF
--- a/app/etl/item/content/parsers/body_content.rb
+++ b/app/etl/item/content/parsers/body_content.rb
@@ -1,6 +1,11 @@
 class Item::Content::Parsers::BodyContent
   def parse(json)
-    json.dig("details", "body")
+    parsed = json.dig("details", "body")
+    if parsed.present? && parsed.is_a?(Array)
+      parsed = Hash[*parsed.map(&:values).flatten].fetch("text/html", nil)
+    end
+
+    parsed
   end
 
   def schemas
--- a/app/etl/item/content/parser.rb
+++ b/app/etl/item/content/parser.rb
@@ -83,6 +83,6 @@ private
     @parsers ||= {}
   end
 
-  class InvalidSchemaError < StandardError;
+  class InvalidSchemaError < StandardError
   end
 end
--- a/spec/etl/item/content/body_content_spec.rb
+++ b/spec/etl/item/content/body_content_spec.rb
@@ -1,8 +1,8 @@
 RSpec.describe Item::Content::Parser do
   subject { described_class.instance }
 
-  it 'returns content json from :body for all valid formats' do
-    valid_types = %w[
+  let(:valid_types) do
+    %w[
       answer
       calendar
       case_study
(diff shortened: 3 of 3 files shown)


chao-xian:
[BUG: fix `undefined method `delete!' for Array`](https://trello.com/c/H4ZDWQG9/434-bug-fix-undefined-method-delete-for-array)

app/etl/item/content/parsers/body_content.rb
```diff
@@ -1,6 +1,9 @@
 class Item::Content::Parsers::BodyContent
   def parse(json)
-    json.dig("details", "body")
+    parsed = json.dig("details", "body")
+    parsed = parsed[0].dig('content') if parsed.present? && parsed.is_a?(Array)
```
MatMoore: It looks like when the body is an array there can be multiple items with different mime types

I ran this query on dev:

```
publishing_api_development=> select foo.size, count(*) from (select json_array_length(json_extract_path(details, 'body')) as size from editions where json_extract_path(details, 'body', '0') is not null) foo group by foo.size;
 size | count
------+-------
    1 | 33809
    2 | 97092
```

An example of the second case is `/guidance/manual-of-patent-practice-mopp/section-14-the-application`

which has a body like this:

```json
[
{"content_type":"text/govspeak","content":"##Section 14: The application   \r\n\r\n###14.01\t\r\n\r\nThis section prescribes the manner of making an application, its form and content, the requirements which must be fulfilled by the contents, and provides for withdrawal of the application. ...."},
{"content_type":"text/html","content":"<h2 id=\"section-14-the-application\">Section 14: The application</h2>\n\n<h3 id=\"section\">14.01</h3>\n..."}
]
```

I think in that case we should be preferring one of them (text/html or text/govspeak) rather than just taking the first one.

Otherwise I don't think we can guarantee what we end up with and we might calculate metrics in an inconsistent way.

app/etl/item/content/parsers/body_content.rb
```diff
@@ -1,6 +1,9 @@
 class Item::Content::Parsers::BodyContent
   def parse(json)
-    json.dig("details", "body")
```
pmanrubia: Before merging, I would rather finding why our integration tests are not covering this problem. I am currently working on this.

app/etl/item/content/parsers/body_content.rb
chao-xian: Thanks @MatMoore - fixed in 1e74471dc0c6c9aafc1f5384fea42cbecf14d7f2

chao-xian:
I agree with @pmanrubia. Have addressed @MatMoore's comment meanwhile.

pmanrubia:
Thank you @chao-xian
Superseded by https://github.com/alphagov/content-performance-manager/pull/756, which is a continuation of this work.

cc://  @MatMoore

PR #756, opened 2018-06-20
[WIP] Integration tests parsing schemas

pmanrubia:


app/etl/item/content/parsers/body_content.rb
```diff
@@ -1,6 +1,11 @@
 class Item::Content::Parsers::BodyContent
   def parse(json)
-    json.dig("details", "body")
+    parsed = json.dig("details", "body")
+    if parsed.present? && parsed.is_a?(Array)
+      parsed = Hash[*parsed.map(&:values).flatten].fetch("text/html", nil)
```
MatMoore: This line was quite hard for me to understand just from reading it.

Can you split it into multiple statements? I think you can also lose the splat operator if you do `parsed.map(&:values).to_h`.

Something like

```ruby
content_by_content_type = parsed.map(&:values).to_h
content = content_by_content_type.fetch("text/html", nil)
```

spec/support/schemas_iterator.rb
```diff
@@ -0,0 +1,9 @@
+module SchemasIterator
```
MatMoore: can we move this to the schemas gem?

pmanrubia: Good idea, I will give it a try

app/etl/item/content/parsers/body_content.rb
pmanrubia: Addressed in 1f365079d5efd3af103d8c8dc07eeda52ab49b63

pmanrubia:
Final review/pairing with @theKHutDeveloper

PR #762, opened 2018-06-22
Rescue publishing api consumer rake tasks to prevent sigterm errors b…

KyleMacPherson:
…eing raised

As discussed with @MatMoore, this PR includes no tests due to the fact that there is no good way to test this functionality, and any test written to do so would just be testing the ruby `rescue` method.


theKHutDeveloper:
[(1) Fix: SignalException: SIGTERM](https://trello.com/c/3UYmbd1k/398-1-fix-signalexception-sigterm)

PR #763, opened 2018-06-22
Support multi-part pages

MatMoore:
This adds support for storing individual parts of multi-part pages (guides and travel advice).

More details in the individual commit messages. There are specs in `spec/integration/streams/process_sub_pages_spec.rb` and I recommend reviewing those first, as there may be edge cases I've missed.

Trello: https://trello.com/c/oDKoKiim/415-5-update-etl-to-handle-sub-pages-for-guides-basepath

### DIFF
--- a/app/streams/publishing_api/item_handler.rb
+++ b/app/streams/publishing_api/item_handler.rb
@@ -1,8 +1,7 @@
 class PublishingAPI::ItemHandler
-  def initialize(old_item:, new_item:, subpage:)
+  def initialize(old_item:, new_item:)
     @old_item = old_item
     @new_item = new_item
-    @subpage = subpage
   end
 
   def process!
@@ -25,7 +24,6 @@ private
 
   attr_reader :old_item
   attr_reader :new_item
-  attr_reader :subpage
 
   def new_version?
     new_item && new_item.newer_than?(old_item)
@@ -42,6 +40,6 @@ private
 
   def grow_dimension!
     new_item.promote!(old_item)
-    Item::Processor.run(new_item, Date.today, subpage: subpage)
+    Item::Processor.run(new_item, Date.today)
   end
 end
--- a/app/streams/publishing_api/item_handler.rb
+++ b/app/streams/publishing_api/item_handler.rb
@@ -1,0 +1,55 @@
+class PublishingAPI::ItemHandler
+  def initialize(old_item:, new_item:, subpage:)
+    @old_item = old_item
+    @new_item = new_item
+    @subpage = subpage
+
+    # old_item may be nil if the item belongs to a brand new document
+    # new_item may be nil if a base_path is reused for a different content id
+    raise ArgumentError if old_item.nil? && new_item.nil?
+  end
+
+  def process!
+    if new_item
+      grow_dimension! if new_version?
+    else
+      deprecate_only!
+    end
+  end
+
+  def process_links!
+    if new_item
+      grow_dimension! if new_version? && links_have_changed?
+    else
+      deprecate_only!
+    end
+  end
+
+private
+
+  attr_reader :old_item
+  attr_reader :new_item
+  attr_reader :subpage
+
+  def new_version?
+    new_item && new_item.newer_than?(old_item)
+  end
+
+  def links_have_changed?
+    return true if old_item.nil?
+
+    HashDiff::Comparison.new(
+      old_item.expanded_links.deep_sort,
+      new_item.expanded_links.deep_sort
+    ).diff.present?
+  end
+
+  def grow_dimension!
+    new_item.promote!(old_item)
+    Item::Processor.run(new_item, Date.today, subpage: subpage)
+  end
+
+  def deprecate_only!
+    old_item.deprecate!
+  end
+end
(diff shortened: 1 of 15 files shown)


app/streams/publishing_api/item_handler.rb
```diff
@@ -0,0 +1,53 @@
+class PublishingAPI::ItemHandler
+  def initialize(old_item:, new_item:, subpage:)
+    @old_item = old_item
+    @new_item = new_item
+    @subpage = subpage
+
+    # old_item may be nil if the item belongs to a brand new document
+    # new_item may be nil if a base_path is reused for a different content id
```
pmanrubia: It is not clear to me how the `base_path` can be nil. If reusing a `base_path` implies that we need to guard agains a `nil` issue at this point, I would try to detect earlier, and make it explicit in our code.

app/streams/publishing_api/message_adapter.rb
```diff
@@ -1,35 +1,78 @@
...
         public_updated_at: parse_time('public_updated_at'),
-        schema_name: message.payload.fetch('schema_name'),
+        schema_name: item_message.payload.fetch('schema_name'),
         latest: true,
-        raw_json: message.payload,
+        raw_json: item_message.payload,
       )
     end
 
+    def to_dimension_items
```
pmanrubia: Could you please add a test for this? You might be able to dry it up afterwards as the only different between `to_dimension_item` and `to_dimension_items` is the `base_path`

app/streams/publishing_api/message_handler.rb
```diff
@@ -7,39 +7,61 @@ def self.process(*args)
...
   end
 
   def process
-    return unless new_item.older_than?(old_item)
-
-    if is_links_update?
-      grow_dimension! if links_have_changed?
-    else
-      grow_dimension!
+    items.each do |item|
```
pmanrubia: I would rename `items` to `items_handlers` as the Item does not have this behaviour.

app/streams/publishing_api/message_handler.rb
```diff
@@ -7,39 +7,61 @@ def self.process(*args)
...
-  attr_reader :message, :new_item, :old_item
+  attr_reader :message
 
-  def is_links_update?
-    routing_key = message.delivery_info.routing_key
-    routing_key.ends_with?('.links')
+  def get_old_items(new_base_paths)
+    content_id = message.payload["content_id"]
+
+    # Get everything with the same content id, so we can deprecate parts
```
pmanrubia: I would extract this query into the `Item` and I would use the tests to describe the how it behaves. This way we can get rid of the comment as well.

app/streams/publishing_api/message_handler.rb
```diff
@@ -7,39 +7,61 @@ def self.process(*args)
...
+      Dimensions::Item.where(base_path: new_base_paths, latest: true)
+    )
   end
 
-  def links_have_changed?
-    current_links = old_item.raw_json['expanded_links'].deep_sort
-    new_links = new_item.raw_json['expanded_links'].deep_sort
+  def items
+    adapter = PublishingAPI::MessageAdapter.new(message)
+    subpages = adapter.subpages_by_base_path
```
pmanrubia: It is not clear to me why we have this logic. 
Wouldn't it be enough with:

1. Get all the items for the message (using the adapter)
2. For each new item:

```ruby
PublishingAPI::ItemHandler.new(
  old_item: old_item #Find the old_item for the `new_item.base_path`,
  new_item: new_item,
#  subpage: subpage Not needed, as it would be the `new_item.base_path`
)
```

I know there have to be reasons for the way you have written this code... I would love to find better ways to make it more obvious for the readers.

spec/integration/streams/process_sub_pages_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+
+    expect(parts).to eq Set[
+      ["/base-path/part1", "Part 1"],
+      ["/base-path/part2", "Part 2"],
+      ["/base-path/part3", "Part 3"],
+      ["/base-path/part4", "Part 4"]
+    ]
+  end
+
+  it "can incrementally reprocess messages containing multiple parts, if some of them fail" do
```
pmanrubia: Not sure if we need to do this, and if the pros/cons to handle this situations is worth from a complexity perspective.

MatMoore:
This ready for another code review.

app/streams/publishing_api/message_adapter.rb
```diff
@@ -1,21 +1,48 @@
...
-
     def initialize(message)
       @message = message
     end
 
-    def to_dimension_item
-      Dimensions::Item.new(
-        content_id: message.payload.fetch('content_id'),
-        base_path: message.payload.fetch('base_path'),
+    def existing_dimension_items
```
pmanrubia: This method doesn't feel like part of the adapter. I would move it to the Handler, although I might be missing something.

MatMoore:
@pmanrubia is this what you had in mind?

MatMoore:
I've found a bug with the logic for deprecating old items. It is not taking into account locale, so if you have two items like `/government/world-location-news/ambassadors-speech-at-international-anticorruption-day` and `/government/world-location-news/ambassadors-speech-at-international-anticorruption-day.fr`, then it is marking one of them as outdated because they share the same content ID.

I'm going to add a test for this and fix this now.

PR #764, opened 2018-06-25
Fix bug in links_have_changed logic in message_handler

KyleMacPherson:
We are getting a lot of errors when we try to compared links for items in which 'expanded_links' field
is not included in their raw_json field. This PR checks for the presence of `expanded_links` before it
compares the values.

### DIFF
--- a/app/models/dimensions/item.rb
+++ b/app/models/dimensions/item.rb
@@ -30,6 +30,14 @@ class Dimensions::Item < ApplicationRecord
     update(latest: true)
   end
 
+  def expanded_links
+    if raw_json && raw_json['expanded_links']
+      raw_json['expanded_links']
+    else
+      {}
+    end
+  end
+
 protected
 
   def deprecate!
--- a/app/streams/publishing_api/message_handler.rb
+++ b/app/streams/publishing_api/message_handler.rb
@@ -32,10 +32,11 @@ private
   end
 
   def links_have_changed?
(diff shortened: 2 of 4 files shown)


KyleMacPherson:
@pmanrubia the `fetch` method did not seem to work, even when I used `with_indifferent_access` to allow querying of the hash by symbols so I went with this solution.

app/streams/publishing_api/message_handler.rb
```diff
@@ -32,8 +32,17 @@ def is_links_update?
   end
 
   def links_have_changed?
-    current_links = old_item.raw_json['expanded_links'].deep_sort
-    new_links = new_item.raw_json['expanded_links'].deep_sort
+    current_links = if old_item && old_item.raw_json['expanded_links']
```
pmanrubia: Could you please add a method in the Item to return the expanded links. You could unit test it and this method can be simplified to:

```ruby
HashDiff::Comparison.new(
  current_links.expanded_links.deep_sort,  
  new_links.expanded_links.deep_sort
).diff.present?
```

app/models/dimensions/item.rb
```diff
@@ -30,6 +30,10 @@ def promote!(old_item)
     update(latest: true)
   end
 
+  def expanded_links
```
pmanrubia: Could you please add a unit test for this method? Not sure what happens when `raw_json` is nil...

PR #773, opened 2018-07-03
Update sprockets

chao-xian:
The sprockets version used by the scripts to generate the docs/api has a security vulnerability. Updated by updating `govuk_tech_docs` and regenerating the site.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -539,4 +539,4 @@ RUBY VERSION
    ruby 2.5.1p57
 
 BUNDLED WITH
-   1.16.1
+   1.16.2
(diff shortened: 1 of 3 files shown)


cbaines:
Looks like you've updated a few other gems, not just Sprockets?

Also, how are the `package-lock.json` file changes involved? (I don't even know what that file does)

cbaines:
Interesting, looking at the CVE report [1], it seems to relate to the Sprockets server, so static sites shouldn't be affected.

1: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-3760

Anyway, thanks for adding some more information. I think it would be clearer to say "The sprockets version used by the scripts to generate the docs/api has a security vulnerability", rather than saying that "The sprockets version ... uses a vulnerable version of sprockets".

chao-xian:
@cbaines re static site: yeah but Github kept complaining anyway. Have updated the PR and commit messages now as well from your latest suggestion. Thanks Chris!

PR #776, opened 2018-07-06
Introduce helper to remove noise from tests

peteglondon:
Introduced MetricsHelpers to allow a simple way to
create metrics in tests and refactored tests to use them.

You can now create a metric with code like:

create_metric base_path: base_path, latest: false, date: '2018-01-13',
      edition: {
        number_of_pdfs: 30, number_of_word_files: 30, readability_score: 123
      },
      daily: {
        pageviews: 20, unique_pageviews: 10
      }

### DIFF
--- a/spec/etl/feedex/processor_spec.rb
+++ b/spec/etl/feedex/processor_spec.rb
@@ -1,20 +1,16 @@
 require 'gds-api-adapters'
 
 RSpec.describe Feedex::Processor do
-  subject { described_class }
-
-  let!(:item1) { create :dimensions_item, base_path: '/path1' }
-  let!(:item2) { create :dimensions_item, base_path: '/path2' }
+  include MetricsHelpers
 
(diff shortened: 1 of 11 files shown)


spec/etl/feedex/processor_spec.rb
```diff
@@ -23,25 +19,23 @@
...
-      fact = create :metric, dimensions_item: item, dimensions_date: dimensions_date, feedex_comments: 9
+      fact = create_metric base_path: '/non-matching-path', date: '2018-02-20', daily: { feedex_comments: 9 }
 
       described_class.process(date: date)
 
       expect(fact.reload).to have_attributes(feedex_comments: 9)
     end
 
     it 'deletes the events that matches the base_path of an item' do
-      item2.destroy
```
pmanrubia: Are we ok with removing the `item2.destroy` ? In theory it was there for a reason

PR #780, opened 2018-07-10
Store events received in the PublishingAPI

pmanrubia:
[Trello card](https://trello.com/c/WCKDsjZH/419-5-extract-events-into-their-own-table)

We currently receive events via RabbitMQ with each update in Publishing API.
To be able to recreate the changes through time (should any bug appears), 
we need to track the events so they could be replied if needed

### What

Stores the events in its own table and link with the Dimension::Item

### Why

It will increase the robustness of the CPM, and will also make it more efficient as the queries about metrics don't need to carry the raw_json.

### DIFF
--- a/app/models/publishing_api_event.rb
+++ b/app/models/publishing_api_event.rb
@@ -3,4 +3,8 @@ class PublishingApiEvent < ApplicationRecord
 
   validates :payload, presence: true
   validates :routing_key, presence: true
+
+  def is_links_update?
+    routing_key.ends_with?('.links')
+  end
 end
--- a/app/models/publishing_api_event.rb
+++ b/app/models/publishing_api_event.rb
@@ -1,0 +1,6 @@
+class PublishingApiEvent < ApplicationRecord
+  has_many :dimensions_items, class_name: 'Dimensions::Item'
+
+  validates :payload, presence: true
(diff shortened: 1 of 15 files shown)


app/models/publishing_api_event.rb
```diff
@@ -0,0 +1,6 @@
+class PublishingApiEvent < ApplicationRecord
+  has_many :dimensions_items, class_name: 'Dimensions::Item'
```
peteglondon: I'm a bit confused about this relationship, At first glance I would expect the one-to-many to be the other way around.

pmanrubia: Yes, but if for example the message is for a guide creation, we will have many items (one per path), and all the items are linked to the same message. 
I hope this makes sense. Let me know if you need further clarification.

pmanrubia: Yes, but if for example the message is for a guide creation, we will have many items (one per path), and all the items are linked to the same message. 
I hope this makes sense. Let me know if you need further clarification.

peteglondon: That makes sense

pmanrubia:
Thank you for your review. 
I have rebased and forced pushed after resolving 1 conflict

PR #782, opened 2018-07-10
Add dimensions item for each sub page of travel advice content

KyleMacPherson:
We have 2 issues with our code which creates a dimensions item for each subpath of a multipart content item.

1) The base page should not have the slug appended to it’s base path.
2) The base page of travel advice documents do not have a dimensions item created for their base page due to the difference in schema between travel advice documents and guide documents.

This PR fixes these issues.

For issue (1) I have not appended the slug to the base path for the first part. This does however rely on the assumption that parts are all rendered in order.

For issue (2) I have used the content found in the `summary` field nested within `details` as the base page and have transformed this data into a schema which our current parts parser can parse.


chao-xian:
[(5) Process Travel Advice Subpaths](https://trello.com/c/So2lRiT2/456-5-process-travel-advice-subpaths)

PR #790, opened 2018-07-16
Don't get quality metrics unless  the content has changed

peteglondon:
The quality metric service is quite slow. In order
to reduce the load, this commit prevents the quality
metrics being re-calculated unless the text content has
changed since the previous version.

Trello:[Do not get quality metrics if content has not changed](https://trello.com/c/cuXkXicW/457-5-do-not-get-quality-metrics-if-content-has-not-changed)


pmanrubia:
@peteglondon I would like to have a look at it

KyleMacPherson:
No worries boss @pmanrubia

app/domain/etl/item/content/content_populator.rb
```diff
@@ -0,0 +1,11 @@
+class Etl::Item::Content::ContentPopulator
+  def self.process
+    new.process
+  end
+
+  def process
+    Dimensions::Item.where(document_text: nil).each do |item|
```
pmanrubia: It might be better to rely on [find_each](http://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each) so we don't overload the process memory?

app/domain/etl/item/content/content_populator.rb
```diff
@@ -0,0 +1,11 @@
+class Etl::Item::Content::ContentPopulator
```
pmanrubia: This code is a `one off` thing, that is migrating content as per a content schema. I am not sure it should live within our ETL processes folder.

What are your thoughts about:

1. Inline this code within the rake task, and unit test it at the rake task level
2. Move this rake task into the following folder: `/lib/tasks/one_off/001-migrate-dimension-item.rake`

app/domain/etl/item/processor.rb
```diff
@@ -1,25 +1,46 @@
...
-      edition.update(Etl::Item::Quality::Service.new.run(item.document_text))
+    if content_changed?
+      create_new_edition
+    else
+      clone_existing_edition
     end
   end
+
+  def content_changed?
+    return true unless old_item
```
pmanrubia: Could you please check what happens with both content items have no text (return `nil`)? This method will return `true` but in reality should be `false`?

app/domain/etl/item/processor.rb
```diff
@@ -1,25 +1,46 @@
...
+    Facts::Edition.create!({
+      number_of_pdfs: Etl::Item::Metadata::NumberOfPdfs.parse(new_item.raw_json),
+      number_of_word_files: Etl::Item::Metadata::NumberOfWordFiles.parse(new_item.raw_json),
+      dimensions_date: dimensions_date,
+      dimensions_item: new_item
+}.merge(quality_metrics))
+  end
+
+  def quality_metrics
+    return {} if new_item.document_text.blank?
```
pmanrubia: It seems to me that the decision around returning `{}` when the `document_text` is blank would be better placed within the service, so can isolate this behaviour in the service and not in the client.

I would change the signature of the method to:

```ruby
def run(new_item)
end
```

Instead of passing the text itself.

app/domain/etl/item/processor.rb
peteglondon: It does return the correct thing.I just added an extra test to cover this scenario.

app/domain/etl/item/processor.rb
peteglondon: Done!

app/domain/etl/item/content/content_populator.rb
peteglondon: Done!

app/domain/etl/item/processor.rb
peteglondon: This logic is moved into Etl::Item::Quality::Service#run

PR #816, opened 2018-07-24
Store list of attributes from events into the dimensions_items table

theKHutDeveloper:
[Trello: Store list of event attributes](https://trello.com/c/6L9Rejah/530-3-store-list-of-attributes-from-the-events-into-the-dimensionsitem-depends-on-474)

We are storing a number of event attributes so that we do not have to re-populate data backwards over time. Because we can't store all the events, we need to store in advance as much information as possible so we can satisfy the user needs for the long run.

### DIFF
--- a/app/streams/publishing_api/message_adapter.rb
+++ b/app/streams/publishing_api/message_adapter.rb
@@ -63,6 +63,12 @@ module PublishingAPI
         public_updated_at: parse_time('public_updated_at'),
         schema_name: message.payload.fetch('schema_name'),
         raw_json: message.payload,
+        phase: message.payload.fetch('phase', nil),
+        publishing_app: message.payload.fetch('publishing_app', nil),
+        rendering_app: message.payload.fetch('rendering_app', nil),
+        analytics_identifier: message.payload.fetch('analytics_identifier', nil),
+        update_type: message.payload.fetch('update_type', nil),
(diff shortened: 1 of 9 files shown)


app/streams/publishing_api/message_adapter.rb
```diff
@@ -63,6 +63,12 @@ def attributes
         public_updated_at: parse_time('public_updated_at'),
         schema_name: message.payload.fetch('schema_name'),
         raw_json: message.payload,
+        phase: message.payload.fetch('phase', nil),
+        publishing_app: message.payload.fetch('publishing_app', nil),
+        rendering_app: message.payload.fetch('rendering_app', nil),
+        analytics_identifier: message.payload.fetch('analytics_identifier', nil),
+        update_type: message.payload.fetch('update_type', nil),
+        links: message.payload.fetch('links', nil),
```
pmanrubia: I would rather storing the `expanded_links` instead of the `links`. This way we also store the exact representation of those entities when the item is updated.

If you rename the attribute to `expanded_links` you would also be able [to remove the method ](https://github.com/alphagov/content-performance-manager/blob/13642c1e0dc18d633c84ba9aae1f706cfc449842/app/models/dimensions/item.rb#L30) that is currently wrapping the `raw_json`. 

On the same go you would also be able to remove `raw_json`?

PR #821, opened 2018-07-25
Revert "Remove unique index on base_path and latest"

chao-xian:
Reverts alphagov/content-performance-manager#820


chao-xian:
Schema timestamp needs to be bumped. Closing this PR to raise another.

PR #835, opened 2018-08-02
Assign default fetch value to message.payload.locale

KyleMacPherson:
Key error is currently raised if we attempt `message.payload.fetch(locale)` on a
message from the publishing api rabbitmq queue without a locale.

Issue was caused in this pr: https://github.com/alphagov/content-performance-manager/pull/825/files

Add a default value to the fetch calls to avoid raising key errors to sentry.


KyleMacPherson:
[(2) Fix missing `locale` bug](https://trello.com/c/FEjADuVB/573-2-fix-missing-locale-bug)

PR #837, opened 2018-08-03
Fix missing `expanded_links` bugs

KyleMacPherson:
NoMethodError is currently being raised when items do not
have expanded_links and we attempt to call .deep_sort on nil.

Fix bug by assigning expanded_links to a variable and ensuring it
can never be nil.

### DIFF
--- a/app/streams/publishing_api/item_handler.rb
+++ b/app/streams/publishing_api/item_handler.rb
@@ -19,10 +19,14 @@ private
 
   def links_have_changed?
     return true if old_item.nil?
+    return false if new_item.nil?
+
+    old_links = old_item.expanded_links ? old_item.expanded_links : {}
+    new_links = new_item.expanded_links ? new_item.expanded_links : {}
 
     HashDiff::Comparison.new(
-      old_item.expanded_links.deep_sort,
-      new_item.expanded_links.deep_sort
+      old_links.deep_sort,
+      new_links.deep_sort
     ).diff.present?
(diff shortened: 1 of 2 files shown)


app/streams/publishing_api/item_handler.rb
```diff
@@ -19,10 +19,14 @@ def process_links!
 
   def links_have_changed?
     return true if old_item.nil?
+    return false if new_item.nil?
+
+    old_links = old_item.expanded_links ? old_item.expanded_links : {}
```
chao-xian: This will work, but I'm wondering whether we should set `{}` as the default for `expanded_links`. 

Would mean migrations to update the column, add that default for previous data that currently have `nil`, and adding an `attributes` default on the model (maybe). Maybe for a separate PR though. But I'm not really sure that's a good approach, so want to know what are your thoughts?

KyleMacPherson: Weirdly the default value is already set to `{}` which could suggest that our data is dodgy, however I think it is worth pushing this to fix the errors in sentry for now.

chao-xian: Yep thanks for looking. Feel free to merge for now. As I said, we can look at that later.

KyleMacPherson:
[(2) Fix expanded links bug](https://trello.com/c/cXOACCNy/575-2-fix-expanded-links-bug)

PR #854, opened 2018-08-16
Remove unused quality metrics code

KyleMacPherson:
Follow up work from https://github.com/alphagov/content-performance-manager/pull/846.

As explained in that PR we are no longer need to make calls to the
govuk-content-quality-service. This PR removes the code we used for those calls
as it is no longer needed.


spec/support/quality_metrics_helpers.rb
```diff
@@ -1,16 +1,7 @@
 module QualityMetricsHelpers
```
pmanrubia: Not sure if we still need this class?

spec/integration/item/import_edition_metrics_spec.rb
```diff
@@ -90,16 +90,7 @@
 
   def stub_quality_metrics_request
```
pmanrubia: Same here, it seems to me that we don't need to stub the metrics this way...

KyleMacPherson:
@pmanrubia nice spot, thanks Pablo

PR #857, opened 2018-08-17
Handle the GdsApi::TimedOutException when calling support-api

peteglondon:
We have been getting a GdsApi::TimedOutException when
retrieving feedex comments from the support API.

The default timeout for gds-api-adapters is 4 seconds.

This PR overrides the default to allow a 15 second timeout.

It also traps any GdsApi:: errors and sends the error to sentry

Trello: https://trello.com/c/hzJF3BrN/599-2-fix-timeouterror-in-feedex

### DIFF
--- a/app/domain/etl/feedex/processor.rb
+++ b/app/domain/etl/feedex/processor.rb
@@ -27,6 +27,11 @@ private
       Events::Feedex.import(events, batch_size: BATCH_SIZE)
       batch += 1
     end
+  rescue GdsApi::TimedOutException,
+         GdsApi::HTTPNotFound,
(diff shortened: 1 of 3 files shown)


app/domain/etl/feedex/processor.rb
```diff
@@ -27,6 +27,8 @@ def extract_events
       Events::Feedex.import(events, batch_size: BATCH_SIZE)
       batch += 1
     end
+  rescue GdsApi::HTTPErrorResponse => e
```
pmanrubia: Could you please add a unit test?

PR #861, opened 2018-08-20
[do not merge] Use govuk-app-config healthcheck support

pmanrubia:
[Trello card](https://trello.com/c/jOTrrdNP/605-investigate-threads-issue-when-upgrading-to-rails-521)

We have implemented our own way of supporting healthchecks, which is ok
but we should be [using the gem that GOV.UK has created for all the
applications](https://github.com/alphagov/govuk_app_config).

This update is also related to an incidence when the existing
implementation of the healthcheck was causing issues with a memory leak
with Rails 5.2.1

In any case, we should use the gem support.

I am also adding a check to ensure that we collected daily metrics.

### DIFF
--- a/app/domain/healthchecks/daily_metrics_check.rb
+++ b/app/domain/healthchecks/daily_metrics_check.rb
@@ -1,0 +1,21 @@
+module Healthchecks
+  class DailyMetricsCheck
+    def name
+      :daily_metrics
+    end
+
+    def status
+      metrics.any? ? :ok : :critical
+    end
+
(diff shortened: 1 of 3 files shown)


pmanrubia:
Waiting to add thresholds as per @erino feedback. The current implementation would raise alarms in the period of time between 00:00 and the time the process is finished.

pmanrubia:
I am closing this PR. I am no planning in merging it now as we need to do further discovery on this matter.

PR #864, opened 2018-08-21
Remove turbolinks

pmanrubia:
There is no need to use turbolinks because the application is mainly an
API that serve JSON requests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,6 @@ gem 'rack-proxy'
 gem 'rails', '~> 5.1'
 gem 'ruby-progressbar'
 gem 'sass-rails'
-gem 'turbolinks'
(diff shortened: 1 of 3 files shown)


erino:
What about removing all of the frontend gems, such as `chartkick`, `jquery-rails`, `sass-rails` etc.?

pmanrubia:
@erino Good point, but we are currently using them for the Sandbox.

erino:
Fair enough. Will we be able to remove the sandbox and these gems once the admin app is in beta?

pmanrubia:
Not sure, but worth having that discussion. For the time being, it provides a UI with all the metrics (including the ones that are out of beta) and includes organisations...

PR #866, opened 2018-08-21
Remove raw json from dimensions item

peteglondon:
Storing the json for each Dimension Item is not scaling well and is no longer needed.

We have decided to rely on what the PublishingAPI stores in case we decide to store other attributes.

Trello: https://trello.com/c/6L4Q4gjd/598-2-remove-rawjson-column-from-dimensionsitem-etl


erino:
Does `raw_json` need to be removed from [`PublishingAPI::MessageAdapter#attributes`](https://github.com/alphagov/content-performance-manager/blob/f025f82/app/streams/publishing_api/message_adapter.rb#L65)?

pmanrubia:
I think it is needed because it is a transient attribute that is used to [retrieve the number of pdfs when processing](https://github.com/alphagov/content-performance-manager/blob/master/app/domain/etl/item/processor.rb#L36-L37).

erino:
:+1: hence adding it as a accessor.

PR #870, opened 2018-08-24
Add satisfaction score to API

chao-xian:
we want to know `satisfaction_score` based on the values of:

- `is_this_useful_yes`
- `is_this_useful_no`

but we don't want to have to calculate this every time we want to make
use of it, such as filtering or ordering by this value. So we store this value when either of the other yes/no values are set.

### DIFF
--- a/app/models/facts/metric.rb
+++ b/app/models/facts/metric.rb
@@ -60,7 +60,7 @@ class Facts::Metric < ApplicationRecord
     ]
   end
 
-  private
+private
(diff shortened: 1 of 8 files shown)


chao-xian:
[(3) Add `satisfaction-score` metric to CPM (EPIC 2)](https://trello.com/c/XBvDdloe/618-3-add-satisfaction-score-metric-to-cpm-epic-2)

PR #872, opened 2018-08-24
Allow metrics endpoint to handle multiple metrics

KyleMacPherson:
In order to display all necessary metrics for a single content item
our api needs to be able to handle multiple metrics. This PR allows that
functionality.

The time-series route is changed to allow metrics params via a query string
to allow for future expansion of the number of metrics we may use.

### DIFF
--- a/app/controllers/api/metrics_controller.rb
+++ b/app/controllers/api/metrics_controller.rb
@@ -19,20 +19,11 @@ class Api::MetricsController < Api::BaseController
 private
 
   def query_series
-    series = Reports::FindSeries.new
-               .between(from: from, to: to)
-               .by_base_path(format_base_path_param)
-               .run
-    if Metric.is_edition_metric?(metrics.first)
-      series
-        .with_edition_metrics
-        .order('dimensions_dates.date asc')
-        .pluck(:date, "facts_editions.#{metrics.first}").to_h
-    else
-      series
-        .order('dimensions_dates.date asc')
-        .pluck(:date, metrics.first).to_h
-    end
+    Reports::FindSeries.new
+       .between(from: from, to: to)
+       .by_base_path(format_base_path_param)
+       .by_metrics(params[:metrics])
+       .run
   end
 
   def format_base_path_param
@@ -40,10 +31,10 @@ private
     "/#{base_path}"
   end
 
-  delegate :from, :to, :metrics, :base_path, to: :api_request
+  delegate :from, :to, :base_path, :metrics, to: :api_request
 
   def api_request
-    @api_request ||= Api::Request.new(params.permit(:from, :to, :base_path, :format, metrics: []))
+    @api_request ||= Api::Request.new(params.permit(:from, :to, :metric, :base_path, :format, metrics: []))
   end
 
   def validate_params!
--- a/app/controllers/api/metrics_controller.rb
+++ b/app/controllers/api/metrics_controller.rb
@@ -19,7 +19,7 @@ class Api::MetricsController < Api::BaseController
 private
 
   def query_series
-    series = Reports::Series.new
+    series = Reports::FindSeries.new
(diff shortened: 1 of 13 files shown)


app/domain/api/request.rb
```diff
@@ -5,27 +5,39 @@ class Api::Request
...
 
-  attr_reader :metric, :from, :to, :base_path
+  attr_reader :metric, :from, :to, :base_path, :metrics
 
-  validates :metric, presence: true, inclusion: { in: Metric.find_all.map(&:name) }
   validates :from, presence: true, format: { with: DATE_REGEX, message: "Dates should use the format YYYY-MM-DD" }
   validates :to, presence: true, format: { with: DATE_REGEX, message: "Dates should use the format YYYY-MM-DD" }
   validates :base_path, presence: true
   validate :from_before_to, if: :have_a_date_range?
+  validate :verify_metrics, if: :have_multiple_metrics
```
pmanrubia: Why do we check if they have `multiple_metrics`? In theory we should only have at least one.

app/domain/api/request.rb
```diff
@@ -5,27 +5,39 @@ class Api::Request
...
+  end
+
   def have_a_date_range?
     from.present? && to.present? && from =~ DATE_REGEX && to =~ DATE_REGEX
   end
 
+  def verify_metrics
+    metric_params = metrics ? metrics : metric
+    metric_params.each do |metric|
+      errors.add("metric", "is not included in the list") unless Metric.find_all.map(&:name).include?(metric)
```
pmanrubia: Could you please extract a method in `Metric` to check if a metric is included?

config/routes.rb
```diff
@@ -7,7 +7,7 @@
 
   namespace :api, defaults: { format: :json } do
     get '/v1/metrics/', to: "metrics#index"
-    get '/v1/metrics/:metric/*base_path/time-series', to: "metrics#time_series"
+    get '/v1/metrics/*base_path/time-series', to: "metrics#time_series"
     get '/v1/metrics/:metric/*base_path', to: "metrics#summary"
```
pmanrubia: As per the [2nd acceptance criteria](https://trello.com/c/vClxSNYC/617-3-update-api-to-support-multiple-metrics-per-basepath-epic-2) wouldn't we need to do the same with the aggregations?

app/controllers/api/metrics_controller.rb
```diff
@@ -23,27 +23,43 @@ def query_series
...
+  end
+
+  def format_metric_names
+    metrics = params[:metrics] ? params[:metrics] : [params[:metric]]
+    metrics.map! do |metric|
+      Metric.is_edition_metric?(metric) ? "facts_editions.#{metric}" : metric
     end
   end
 
+  def dates_to_metrics(series_array)
```
pmanrubia: Doing data transformation in the controller can cause issues in the mid - long term. Could you please move this logic in the the Json view while we set up Serializers?

app/controllers/api/metrics_controller.rb
```diff
@@ -23,27 +23,43 @@ def query_series
...
-      series
-        .with_edition_metrics
-        .order('dimensions_dates.date asc')
-        .pluck(:date, "facts_editions.#{metric}").to_h
-    else
-      series
-        .order('dimensions_dates.date asc')
-        .pluck(:date, metric).to_h
+
+    metric_names = format_metric_names
```
pmanrubia: I would include this logic within `Report#Series` and would add tests accordingly

chao-xian:
[(5) Update API to support multiple metrics per basepath (EPIC 2)](https://trello.com/c/vClxSNYC/617-5-update-api-to-support-multiple-metrics-per-basepath-epic-2)

chao-xian:
Has the api documentation been updated?

PR #882, opened 2018-08-30
Add StatsD for Facts, Editions and Dimensions

pmanrubia:
[Trello card](https://trello.com/c/85P5zbF2/623-3-etl-monitor-add-data-about-dimensions)

Related to PR: https://github.com/alphagov/govuk-puppet/pull/8062

Continuation to work developed in #871. It adds StatsD support for the Dimensions and Facts. 
This is  useful because it will provide us with context about how the processing of events in PublishingAPI impacts in the different version of items that we are storing.

We are publishing to Grafana via StatsD the following metrics:

- Total number of base paths.
- Total number of `latest` base paths.
- Total number of content items.
- Total number of `latest` content items.

- Total number of metrics
- Total number of metrics for yesterday
- Total number of editions
- Total number of editions for yesterday

### DIFF
--- a/app/domain/monitor/facts.rb
+++ b/app/domain/monitor/facts.rb
@@ -7,6 +7,8 @@ class Monitor::Facts
   def run
     count_facts_metrics!
     count_facts_daily_metrics!
+
+    count_facts_editions!
   end
 
 private
@@ -21,6 +23,13 @@ private
   def count_facts_daily_metrics!
     path = path_for('daily_metrics')
     count = Facts::Metric.for_yesterday.count
+
+    GovukStatsd.count(path, count)
+  end
+
+  def count_facts_editions!
+    path = path_for('all_editions')
+    count = Facts::Edition.count
 
     GovukStatsd.count(path, count)
   end
--- a/app/domain/monitor/facts.rb
+++ b/app/domain/monitor/facts.rb
@@ -4,37 +4,37 @@ class Monitor::Facts
   end
 
   def run
-    count_facts_metrics!
-    count_facts_daily_metrics!
+    statsd_for_all_metrics!
+    statsd_for_yesterday_metrics!
 
-    count_facts_editions!
-    count_daily_facts_editions!
+    statsd_for_total_editions!
+    statsd_for_yesterday_editions!
   end
 
 private
 
-  def count_facts_metrics!
+  def statsd_for_all_metrics!
     path = path_for('all_metrics')
     count = Facts::Metric.count
 
     GovukStatsd.count(path, count)
   end
 
-  def count_facts_daily_metrics!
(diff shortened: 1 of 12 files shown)


pmanrubia:
Note: the build failure is related to the latest version that has been manually 
deployed to integration because it has dropped a column `expanded_links`
that is currently in used in this branch.

Tomorrow (31st Aug), if we rebuild this branch in integration, the error should 
have gone.

app/domain/monitor/facts.rb
```diff
@@ -17,6 +18,13 @@ def count_facts_metrics!
     GovukStatsd.count(path, count)
   end
 
+  def count_facts_daily_metrics!
```
chao-xian: Can we rename all the `count_blah_daily_thing` methods to something like `count_blah_for_day`? Daily suggests that we are aggregating more than a day's worth here, when in fact we're not.

spec/domain/monitor/dimensions_spec.rb
```diff
@@ -9,15 +9,15 @@
 
   before { allow(GovukStatsd).to receive(:count) }
 
-  it 'sends the total number of base_paths' do
+  it 'sends StatsD counter of base_paths' do
```
chao-xian: Sorry, I'm not clear how the commit message applies to the changes to the spec.

app/models/facts/metric.rb
```diff
@@ -21,7 +21,7 @@ def is_this_useful_no=(value)
     self.satisfaction_score = calculate_satisfaction_score
   end
 
-  scope :for_yesterday, -> { where(dimensions_date: Dimensions::Date.for(Date.yesterday)) }
+  scope :within_yesterday, -> { where(dimensions_date: Dimensions::Date.for(Date.yesterday)) }
```
chao-xian: I think `.for_yesterday` made more sense to my reading, but I understand why you've renamed it. I don't think `.within_yesterday` reads very well though. Perhaps `.from_yesterday` covers your meaning?

app/domain/monitor/facts.rb
chao-xian: Ah I see these get renamed in a later commit.

app/models/facts/metric.rb
pmanrubia: I will remove the commit as per our conversation.

pmanrubia:
@chao-xian thank you for your review. I have addressed your comments.

spec/domain/monitor/dimensions_spec.rb
pmanrubia: As per our conversation I have amended the commit notes explaining that
this names are consistent with other test names for the ETL monitoring.

PR #893, opened 2018-09-05
Add monitoring for publishing api events

KyleMacPherson:
Monitor the events received from publishing API

### DIFF
--- a/app/streams/publishing_api/consumer.rb
+++ b/app/streams/publishing_api/consumer.rb
@@ -1,7 +1,11 @@
 module PublishingAPI
   class Consumer
     def process(message)
-      MessageProcessorJob.perform_later(message.payload) if valid_routing_key?(message)
+      if valid_routing_key?(message)
+        MessageProcessorJob.perform_later(message.payload)
+      else
+        Monitor::Messages.increment_discarded
+      end
 
       message.ack
     rescue StandardError => e
--- a/spec/domain/monitor/messages_spec.rb
+++ b/spec/domain/monitor/messages_spec.rb
@@ -1,0 +1,41 @@
+RSpec.describe Monitor::Messages do
+  before do
+    allow(GovukStatsd).to receive(:increment)
+  end
+
+  it 'increments major if routing key ends .major' do
+    message = build(:message, routing_key: 'news_story.major')
+    expect(GovukStatsd).to receive(:increment).with("monitor.messages.major")
+    Monitor::Messages.run(message)
+  end
+
+  it 'increments minor if routing key ends .minor' do
+    message = build(:message, routing_key: 'news_story.minor')
+    expect(GovukStatsd).to receive(:increment).with("monitor.messages.minor")
+    Monitor::Messages.run(message)
+  end
+
+  it 'increments links if routing key ends .links' do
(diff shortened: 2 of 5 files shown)


KyleMacPherson:
[(3) ETL Monitor: add data about PublishingAPI integration](https://trello.com/c/XlYQfsha/624-3-etl-monitor-add-data-about-publishingapi-integration)

spec/domain/monitor/messages_spec.rb
```diff
@@ -0,0 +1,35 @@
+RSpec.describe Monitor::Messages do
+  before do
+    allow(GovukStatsd).to receive(:increment)
+  end
+
+  it 'increments major if routing key ends .major' do
+    expect(GovukStatsd).to receive(:increment).with("major")
+    Monitor::Messages.run('news_story.major')
```
pmanrubia: We shouldn't publish using the same name as the routing key. We would need to follow a similar pattern to the other monitors. Something along the lines to:

```
monitor.messages.major
```

spec/domain/monitor/messages_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  it 'increments unpublish if routing key ends .unpublish' do
+    expect(GovukStatsd).to receive(:increment).with("unpublish")
+    Monitor::Messages.run('news_story.unpublish')
+  end
+
+  it 'increments discarded if routing key ends .discarded' do
+    expect(GovukStatsd).to receive(:increment).with("discarded")
+    Monitor::Messages.run('news_story.discarded')
```
pmanrubia: `news_story.discarded` will never be triggered. This is something to be handled separately by the Job that processes the messages.

app/streams/publishing_api/consumer.rb
```diff
@@ -1,7 +1,7 @@
 module PublishingAPI
   class Consumer
     def process(message)
-      MessageProcessorJob.perform_later(message.payload) if valid_routing_key?(message)
+      MessageProcessorJob.perform_later(message.payload, message.delivery_info.routing_key) if valid_routing_key?(message)
```
pmanrubia: Not sure about passing the routing key as a second parameter.
It would be enough with inserting this code in the Consumer?

```ruby
def process(message)
  Monitor.Messages.process(message)
  MessageProcessorJob.perform_later(message.payload) if valid_routing_key?(message)
end
```

KyleMacPherson:
Puppet PR: https://github.com/alphagov/content-performance-manager/pull/893

app/streams/publishing_api/consumer.rb
```diff
@@ -1,7 +1,11 @@
 module PublishingAPI
   class Consumer
     def process(message)
-      MessageProcessorJob.perform_later(message.payload) if valid_routing_key?(message)
+      if valid_routing_key?(message)
+        MessageProcessorJob.perform_later(message.payload)
+      else
+        Monitor::Messages.increment_discarded unless valid_routing_key?(message)
```
pmanrubia: Not sure if we need to validate the routing key again? ```unless valid_routing_key?(message)```

pmanrubia:
> Puppet PR: #893

@KyleMacPherson It is pointing to a CPM PR

PR #897, opened 2018-09-06
Return only totals for summary

peteglondon:
Currently the /v1/metrics/*base_path endpoint returns results
with the format:
```
{
  "metric_name": {
    "total": 100,
    "latest": 50
  }
}
```

This commit refactors the output to be in this format:

```
{ "metric_name": 100 }
```

Returning only the total.


peteglondon:
[(3) Add metrics Charts for the Single Performance Page (EPIC 2)](https://trello.com/c/xZb8cKyy/625-3-add-metrics-charts-for-the-single-performance-page-epic-2)

PR #899, opened 2018-09-07
Update API docs to reflect multi value params

chao-xian:
A the API now accepts multiple values for the metrics parameter.

The URLs have also changed.


chao-xian:
[(5) Update API to support multiple metrics per basepath (EPIC 2)](https://trello.com/c/vClxSNYC/617-5-update-api-to-support-multiple-metrics-per-basepath-epic-2)

PR #900, opened 2018-09-07
Remove sandbox functionality

theseanything:
The sandbox was deprecated and is being removed as functionality not part of the application and only used for testing purposes. This PR removes the view, controllers and tests for the sandbox.


theseanything:
[(2) Delete Sandbox from Content Performance Manager](https://trello.com/c/6b7mYV79/640-2-delete-sandbox-from-content-performance-manager)

PR #907, opened 2018-09-10
Return a 404 error when not latest item exists

peteglondon:
Current we are getting a 500 error when
we try to get the metadata for a non-existent
base path.

This commit checks for the existence of the Dimensions::Item
and returns a 404 if it does not exist.

### DIFF
--- a/app/controllers/api/metrics_controller.rb
+++ b/app/controllers/api/metrics_controller.rb
@@ -33,7 +33,9 @@ private
   end
 
   def metadata
-    Dimensions::Item.latest_by_base_path(format_base_path_param).first.metadata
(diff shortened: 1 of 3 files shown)


app/controllers/api/metrics_controller.rb
```diff
@@ -33,7 +33,9 @@ def format_base_path_param
   end
 
   def metadata
-    Dimensions::Item.latest_by_base_path(format_base_path_param).first.metadata
+    latest_item = Dimensions::Item.latest_by_base_path(format_base_path_param).first
```
chao-xian: I'm not sure the `.first` can be called if it's `nil`?

chao-xian: It's fine as ActiveRecord returns something useful.

PR #909, opened 2018-09-11
Fix to handle race conditions in Dimensions::Date

theseanything:
This PR addresses the UniqueRecord exception thrown by the `Dimensions::Date::for` method. This is due to race conditions when one process beats another in creating record. The method now retries finding the newly created record when the exception is enccountered. The method has been renamed from `for` to `find_or_create` to clarify its purpose.

### DIFF
--- a/app/models/dimensions/date.rb
+++ b/app/models/dimensions/date.rb
@@ -33,6 +33,8 @@ class Dimensions::Date < ApplicationRecord
   def self.for(date)
     begin
       find_by(date: date) || create_with(date)
+    rescue ActiveRecord::RecordNotUnique
+      find_by(date: date)
     rescue StandardError => e
       GovukError.notify(e)
     end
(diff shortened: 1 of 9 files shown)


theseanything:
[(2) Prevent duplicate date error](https://trello.com/c/GGSXeEYe/667-2-prevent-duplicate-date-error)

app/models/dimensions/date.rb
```diff
@@ -24,11 +24,19 @@ def self.build(date)
...
+    date_dimension
+  end
+
+  def self.find_or_create(date)
+    begin
+      find_by(date: date) || create_with(date)
+    rescue ActiveRecord::RecordNotUnique
+      find_by(date: date)
+    rescue StandardError => e
+      GovukError.notify(e)
```
theseanything: Should not rescue other errors.

PR #910, opened 2018-09-11
Fix ETL not calculating satisfaction_score

chao-xian:
The ETL process was not calculating the satisfaction_score within the
SQL that we execute in order to cope with setting a huge number of
updates. Previously we were only using the the ActiveModel overrides of
the setters for `is_this_useful_yes` and `is_this_useful_no` to then
calculate and set the value of `satisfaction_score` but the SQL bypassed
that mechanism.


chao-xian:
[(2) Fix: User satisfaction score metric](https://trello.com/c/gdVxoRl4/669-2-fix-user-satisfaction-score-metric)

chao-xian:
Thanks for the feedback @pmanrubia. I've addressed your comment.

PR #916, opened 2018-09-13
Add error documentation

theseanything:
This PR adds the initial documentation page for errors presented to user of the API. All existing errors should be listed. Links supplied by error messages to documenation is also fixed, as was missing the `.html` extension.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -31,13 +31,13 @@ private
     # Type is an arbitrary URI identifying the error type
     # https://tools.ietf.org/html/rfc7807#section-3.1 recommends using
     # human-readable documentation for this, so point to our API docs.
-    error_hash[:type] = "https://content-performance-api.publishing.service.gov.uk/errors/##{type}"
+    error_hash[:type] = "https://content-performance-api.publishing.service.gov.uk/errors.html##{type}"
     render json: error_hash, status: :bad_request, content_type: "application/problem+json"
   end
 
   def not_found_response
     response_hash = {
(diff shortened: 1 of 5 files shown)


theseanything:
[(2) Document existing API errors](https://trello.com/c/zuSrLOCh/668-2-document-existing-api-errors)

pmanrubia:
👍 Good work

app/controllers/api/base_controller.rb
```diff
@@ -31,13 +31,13 @@ def error_response(type, error_hash)
...
     # human-readable documentation for this, so point to our API docs.
-    error_hash[:type] = "https://content-performance-api.publishing.service.gov.uk/errors/##{type}"
+    error_hash[:type] = "https://content-performance-api.publishing.service.gov.uk/errors.html/##{type}"
     render json: error_hash, status: :bad_request, content_type: "application/problem+json"
   end
 
   def not_found_response
     response_hash = {
-      type: "https://content-performance-api.publishing.service.gov.uk/errors/#base-path-not-found",
+      type: "https://content-performance-api.publishing.service.gov.uk/errors.html/#base-path-not-found",
```
tijmenb: Would this need to be `https://content-performance-api.publishing.service.gov.uk/errors.html#base-path-not-found`?

peteglondon:
Should we also update errors in the the API reference section to match the errors page?

app/controllers/api/base_controller.rb
theseanything: Good catch! Updated in c1f3100

PR #922, opened 2018-09-20
ADR 13 - Store Content Items JSON in Items Dimension

pmanrubia:
Each Content Item is defined by a JSON that includes all the metadata,
the links and the content details.  The links include information about
organisations, policies, taxonomies, and details including the actual
content of the page, the attachments, and other necessary details like
title, description, publishing-app, rendering app, etc.

In the Data Warehouse, we started storing the Event with the JSON, but
we decided to remove it because our database was growing too quickly.
Later on, we found out that we had a bug in our code because we were not
filtering Publishing-API events correctly. Some days we had more than 4
million events, and each one of them implied a new version of a Content
Item with no real changes.

As of today, the Content Items dimension is growing very slowly, as it
only tracks changes related to content updates. The current pace of
growth is around 600-800 rows per day, which is very small for a Data
Warehouse.

So I would like to explore this decision:

Store the JSON of the Content Items that grow the Items dimension, which
is the same as storing all the information that we currently have about
a Dimension Item.

### DIFF
--- a/doc/arch/adr-013-store-json-from-events-that-carry-content-changes.md
+++ b/doc/arch/adr-013-store-json-from-events-that-carry-content-changes.md
@@ -1,0 +1,40 @@
+# Store Content Item JSON that makes the dimensions grow
+
+20-09-2018
+
+## Context:
+
+Each Content Item is defined by a JSON that includes all the metadata, the links and the content details. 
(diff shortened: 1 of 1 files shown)


pmanrubia:
Thank you for your review @erino

PR #924, opened 2018-09-20
ADR-14: add unique identifier to the DW grain

pmanrubia:
ADR to associate a unique identifier to each item that is tracked in the
Items Dimension.

Each `base_path`, which represents the grain of Data Warehouse, will
have a unique identifier that will never change. This ID will allow us
to perform queries through time on a efficient way, and also to be
accurate with our results.

### DIFF
--- a/doc/arch/adr-014-track-attribute-changes-per-basepath.md
+++ b/doc/arch/adr-014-track-attribute-changes-per-basepath.md
@@ -1,0 +1,87 @@
+# Track attribute changes through time for basepaths
+
+21-09-2018
+
+## Context:
+
+The Data Warehouse tracks changes to Content Items through time so, for
+example, if we change the organisation attached to a Content Item, we should
+know WHEN that change happened and WHAT changed.
+
+Tracking changes not only applies to other entities associated to the Dimension
+Items, but also to core attributes like the `title`, `description`,
+`rendering-app`, or any other attribute. This is the foundation of our Data
+Warehouse: track changes and track performance at the same time.
+
+In GOV.UK we can identify a Content Item on a unique way by their `content_id`.
+
+It is the combination of a `content_id` and a `locale` what makes a Content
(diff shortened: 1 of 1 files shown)


pmanrubia:
@erino could you please review?

doc/arch/adr-014-track-attribute-changes-per-basepath.md
```diff
@@ -0,0 +1,87 @@
...
+results.
+
+On the same go, all the Contnet Items that share a locale will also be uniquely
+referenced in the Data Warehouse.
+
+### Note
+
+We would not need to add this unique identifier if 
+
+1. The sub-items for Content with multiple paths have an UID for each `base_path`.
```
erino: What about "unique id" rather then UID?

doc/arch/adr-014-track-attribute-changes-per-basepath.md
```diff
@@ -0,0 +1,87 @@
+# Track attribute changes through time for b    asepaths
```
erino: There's too many spaces in basepath `b    asepaths`

doc/arch/adr-014-track-attribute-changes-per-basepath.md
pmanrubia: Addressed in: https://github.com/alphagov/content-performance-manager/pull/924/commits/1f9f9618978f6a8546991d544b404d8a52ab93e7

doc/arch/adr-014-track-attribute-changes-per-basepath.md
pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/924/commits/1f9f9618978f6a8546991d544b404d8a52ab93e7

pmanrubia:
@erino thank you for your help and review.

PR #931, opened 2018-09-24
Use content UUID for content query [WIP][DO NOT MERGE]

peteglondon:
The query that returns data for the /content
endpoint now uses the content_uuid column
to join the latest Dimensions::Item to all
previous editions/metrics for those editions.

This means we will get metrics for all editions
even if the base_path, title, document_type have
changed.

This is implementing [the associated ADR](https://github.com/alphagov/content-performance-manager/pull/924)


peteglondon:
I've opened another one on a different branch

PR #938, opened 2018-09-26
Fix satisfaction score metric

pmanrubia:
[Trello card](https://trello.com/c/gdVxoRl4/669-2-fix-user-satisfaction-score-metric)

It removes calculation based on `attr_accessors`, as by adding constraints
to our database model and updating the values, we get to a very simple
solution to update the satisfaction score in the `user_feedback_processor`.

```
SET is_this_useful_no = s.is_this_useful_no,
    is_this_useful_yes = s.is_this_useful_yes,
    satisfaction_score = s.is_this_useful_yes / (s.is_this_useful_yes + s.is_this_useful_no::float)

```


db/migrate/20180926091756_add_default_value_to_feedback_metrics.rb
```diff
@@ -0,0 +1,21 @@
+class AddDefaultValueToFeedbackMetrics < ActiveRecord::Migration[5.2]
+  def up
```
chao-xian: I like this migration. Thanks for sorting this out. I think it might be sensible to add a Data Model, as per https://medium.com/@aweirdcraft/backfilling-data-in-rails-5-0-few-thoughts-on-writing-safe-migrations-1f4c8c4454c9

PR #942, opened 2018-09-27
Fix messsage processing job to call monitoring

chao-xian:
We also refactor slightly to accept just the routing key in the monitoring message.

Paired with @KyleMacPherson


chao-xian:
[(3) ETL Monitor: add data about PublishingAPI integration](https://trello.com/c/XlYQfsha/624-3-etl-monitor-add-data-about-publishingapi-integration)

PR #946, opened 2018-10-04
Rename Dimensions:Item to Dimensions::Edition

peteglondon:
The Dimensions:Item class actually holds the details for a single edition.

The name has caused some confusion among the developers so it's time to rename it.


peteglondon:
[(5) Index page: show 100 first content items](https://trello.com/c/HvQXxa4z/659-5-index-page-show-100-first-content-items)

PR #952, opened 2018-10-09
Add useful totals to single page api endpoint

chao-xian:
This is for calculating the total number of respondents on the front end.


chao-xian:
[(2) Display the total number of users that voted in the yes/no survey](https://trello.com/c/OrZczk9w/693-2-display-the-total-number-of-users-that-voted-in-the-yes-no-survey)

PR #959, opened 2018-10-11
Return TimeSeries objects in Finder 

pmanrubia:
The finder that returns time series returned different types based
on the filter parameters. This should never happen, so I have refactored
it to always return the same type.

On the same go, I have also cleaned up and increase coverage of the 
RSpec test for FindSeries.

### DIFF
--- a/spec/domain/queries/find_series_spec.rb
+++ b/spec/domain/queries/find_series_spec.rb
@@ -76,20 +76,22 @@ RSpec.describe Queries::FindSeries do
 
   describe '#editions' do
     it 'return the content items included in the report' do
-      edition1 = create :edition, base_path: '/path1', date: '2018-1-12'
-      edition2 = create :edition, base_path: '/path2', date: '2018-1-12'
+      edition1 = create :edition, base_path: '/path1', date: '2018-1-12', facts: { words: 1 }
+      edition2 = create :edition, base_path: '/path2', date: '2018-1-12', facts: { words: 2 }
 
-      create :metric, edition: edition1, date: '2018-1-12', pviews: 1
-      create :metric, edition: edition1, date: '2018-1-13', pviews: 2
-      create :metric, edition: edition1, date: '2018-1-14', pviews: 4
-      create :metric, edition: edition2, date: '2018-1-13', pviews: 5
-      create :metric, edition: edition2, date: '2018-1-14', pviews: 6
+      create :metric, edition: edition1, date: '2018-1-12'
+      create :metric, edition: edition1, date: '2018-1-13'
+      create :metric, edition: edition1, date: '2018-1-14'
+      create :metric, edition: edition2, date: '2018-1-13'
+      create :metric, edition: edition2, date: '2018-1-14'
 
-      result = described_class.new.by_metrics(%w(pviews)).by_base_path('/path1').run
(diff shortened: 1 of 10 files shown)


spec/domain/queries/find_series_spec.rb
```diff
@@ -1,108 +1,93 @@
...
-      metric1 = create :metric, date: today, edition: edition1
-      edition2 = create :edition, base_path: '/path/2', date: tomorrow, facts: { words: 20_000 }
-      metric2 = create :metric, date: tomorrow, edition: edition2
-      metric3 = create :metric, date: tomorrow
-
-      series = described_class.new.run
+      create :metric, date: Date.today, pviews: 1
 
-      expect(series).to match_array([metric1, metric2, metric3])
+      expect(described_class.new.by_metrics(%w(pviews)).run.length).to eq(1)
```
theseanything: Should we test that we get the actual filtered metrics (e.g. `pviews`) back?

spec/domain/queries/find_series_spec.rb
```diff
@@ -1,108 +1,93 @@
...
-
-      expect(series).to match_array([metric1, metric2])
-    end
+      create :metric, edition: edition1, date: '2018-1-12', pviews: 1
+      create :metric, edition: edition1, date: '2018-1-13', pviews: 2
+      create :metric, edition: edition2, date: '2018-1-14', pviews: 1
 
-    it 'ignores parameters when blank' do
-      metric = create(:metric, date: Date.today)
+      series = described_class.new.by_metrics(%w(pviews)).by_organisation_id('org-1').run
```
theseanything: Do we need to call `by_metrics`?

spec/domain/queries/find_series_spec.rb
```diff
@@ -1,108 +1,93 @@
...
+      create :metric, edition: edition1, date: '2018-1-14', pviews: 4
+      create :metric, edition: edition2, date: '2018-1-13', pviews: 5
+      create :metric, edition: edition2, date: '2018-1-14', pviews: 6
+
+      result = described_class.new.by_metrics(%w(pviews)).by_base_path('/path1').run
+      expect(result.first.time_series).to eq([
+        { date: "2018-01-12", value: 1 },
+        { date: "2018-01-13", value: 2 },
+        { date: "2018-01-14", value: 4 },
+      ])
```
theseanything: Not sure how this test is different from the one previous?

spec/domain/queries/find_series_spec.rb
pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/959/commits/55942bf20560782583f88c08b5c7e5e685edb877

spec/domain/queries/find_series_spec.rb
pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/959/commits/2861f50b1e9093caf4fc69a58042c88ab8653087

spec/domain/queries/find_series_spec.rb
pmanrubia: Addressed in https://github.com/alphagov/content-performance-manager/pull/959/commits/55942bf20560782583f88c08b5c7e5e685edb877

PR #960, opened 2018-10-12
Add document types end point

KyleMacPherson:
### What?
Add an end point to render all unique document types.

[Trello card](https://trello.com/c/7zjqCznt/726-2-index-page-filter-by-document-type)

### Why?
So that we can render a dropdown with all document types in our front end to allow users to filter by document type

screen shot 2018-10-12 at 13 06 30

### DIFF
--- a/app/domain/queries/find_all_document_types.rb
+++ b/app/domain/queries/find_all_document_types.rb
@@ -1,0 +1,12 @@
+class Queries::FindAllDocumentTypes
+  def self.retrieve
+    new.retrieve
+  end
+
+  def retrieve
+    Dimensions::Edition.latest
+      .select(:document_type)
+      .distinct
+      .order(:document_type).to_a
+  end
+end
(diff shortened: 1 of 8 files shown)


KyleMacPherson:
[(2) Index page - filter by document type](https://trello.com/c/7zjqCznt/726-2-index-page-filter-by-document-type)

theseanything:
Maybe include a test, so we know are getting the appropiate document types from the DB i.e. only the latest?

app/domain/queries/find_all_document_types.rb
```diff
@@ -0,0 +1,12 @@
+class Queries::FindAllDocumentTypes
```
pmanrubia: We would need a Unit test for this class

spec/domain/queries/find_all_document_types_spec.rb
```diff
@@ -0,0 +1,21 @@
+RSpec.describe Queries::FindAllDocumentTypes do
+  context 'when we have data' do
+    before do
+      create :user
+      create :edition, document_type: 'guide'
+      create :edition, document_type: 'travel_advice'
+      create :edition, document_type: 'manual'
+    end
+
+    it 'returns distinct organisations ordered by primary_organisation_title' do
```
pmanrubia: What about adding an edition that is not latest and checking that it returns the correct value?

PR #961, opened 2018-10-15
Fix Satisfaction Score calculation 

pmanrubia:
[Trello card](https://trello.com/c/rqS4X7zm/737-fix-satisfaction-score-issue)

The Satisfaction Score was not being calculated properly because it 
was just adding up each individual percentage.

I have solved it by calculating the aggregations of all metrics before
building the response, and recalculating in that moment the metric.

We are now passing a `Hash` with all the aggregations to the view, so
it is easier to just render the proper aggregated metric.

As a side improvement, I have also removed the limitation of hardcoding
a set of metrics. There is no real benefit in having this constraint, so we 
can safely remove it.

### DIFF
--- a/spec/domain/queries/find_aggregations_spec.rb
+++ b/spec/domain/queries/find_aggregations_spec.rb
@@ -1,7 +1,7 @@
 RSpec.describe Queries::FindAggregations do
   it 'has aggregations for all metric names' do
     result = Queries::FindAggregations.new.run
-    expect(result.keys).to eq(Metric.find_all_names)
+    expect(result.keys).to eq(Metric.find_all_names.map(&:to_sym))
   end
 
   it 'return aggregated values for daily metrics' do
@@ -10,7 +10,6 @@ RSpec.describe Queries::FindAggregations do
     create :metric, edition: edition, date: '2018-01-02', pviews: 3
 
     result = Queries::FindAggregations.new.run
-    expect(result.fetch('pviews')).to eq(5)
     expect(result.fetch(:pviews)).to eq(5)
   end
 
@@ -77,6 +76,6 @@ RSpec.describe Queries::FindAggregations do
   it 'returns nil values when no results' do
     result = Queries::FindAggregations.new.run
 
-    expect(result).to match(hash_including('pviews' => nil, 'upviews' => nil))
+    expect(result).to match(hash_including(pviews: nil, upviews: nil))
   end
 end
--- a/app/domain/queries/find_aggregations.rb
+++ b/app/domain/queries/find_aggregations.rb
@@ -39,7 +39,7 @@ private
 
   def build_response(metrics)
     metric_names = Metric.find_all_names
-    Hash[metric_names.zip(metrics)].with_indifferent_access
+    Hash[metric_names.zip(metrics)].deep_symbolize_keys
   end
 
   def aggregations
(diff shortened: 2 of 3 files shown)


spec/domain/queries/find_aggregations_spec.rb
```diff
@@ -0,0 +1,83 @@
+RSpec.describe Queries::FindAggregations do
+  it 'A hash with a Key with each metric name' do
+    resut = Queries::FindAggregations.new.run
```
theseanything: spelling mistake - result?

spec/domain/queries/find_aggregations_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    expect(resut).to be_a(Hash)
+    expect(resut.keys).to eq(Metric.find_all_names)
```
theseanything: If we are checking the keys - do we need to check its a Hash? (As this would fail anyway)

spec/domain/queries/find_aggregations_spec.rb
```diff
@@ -0,0 +1,83 @@
+RSpec.describe Queries::FindAggregations do
+  it 'A hash with a Key with each metric name' do
```
theseanything: I feel this test title has implementation details - maybe something like `has aggregations for all metric names`?

spec/domain/queries/find_aggregations_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  it 'return aggregated values for daily metrics' do
+    edition = create :edition, base_path: '/path/1', date: '2018-01-01'
+    create :metric, edition: edition, date: '2018-01-01', pviews: 2
+    create :metric, edition: edition, date: '2018-01-02', pviews: 3
+
+    result = Queries::FindAggregations.new.run
+    expect(result.fetch('pviews')).to eq(5)
+    expect(result.fetch(:pviews)).to eq(5)
```
theseanything: Feel this is testing 2 different things - that the aggregated value is correct and that we can access the result using a string or symbol? Maybe separate them for clarity?

spec/domain/queries/find_aggregations_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    it 'return 0 if no responses' do
+      edition = create :edition, base_path: '/path/1', date: '2018-01-01'
+      create :metric, edition: edition, date: '2018-01-01', useful_yes: 0, useful_no: 0
+      create :metric, edition: edition, date: '2018-01-02', useful_yes: 0, useful_no: 0
+
+      result = Queries::FindAggregations.new.run
+
+      expect(result.fetch(:satisfaction)).to be_zero
+    end
```
theseanything: Should we add a test to make sure we are calculating a average correctly? (And not the average of averages like we did before)

spec/requests/api/single_page_spec.rb
```diff
@@ -110,6 +110,40 @@
...
+            'name' => 'upviews',
+            'total' => 30
+            )
+          )
+      }
+      expect(body).to include(expected)
+      expect(response).to have_http_status(:ok)
+    end
+
+    it 'returns the aggregated recalculated value for Satisfaction Score' do
```
theseanything: `aggregated recalculated value` is confusing -  maybe simplify the test title to `returns the satisfaction score for time period`

spec/domain/queries/find_aggregations_spec.rb
pmanrubia: I have finally decided to go with `:symbols` in the response.

PR #964, opened 2018-10-18
Allow filter by document type

peteglondon:
This PR introduces the `document_type` optional parameter to the /content endpoint

When the `document_type` parameter is supplied, the endpoint only returns data where
the latest edition of a content item has that `document_type`

Url to test: http://content-performance-manager.dev.gov.uk/content?document_type=corporate_report&from=2018-09-19&organisation_id=b8a6a8f2-d04f-4346-90c1-d28898565866&to=2018-10-19

### DIFF
--- a/app/controllers/content_controller.rb
+++ b/app/controllers/content_controller.rb
@@ -2,11 +2,7 @@ class ContentController < Api::BaseController
   before_action :validate_params!
 
   def show
-    @content = Queries::FindContent.retrieve(
-      from: from,
-      to: to,
-      organisation_id: params[:organisation_id]
-    )
+    @content = Queries::FindContent.retrieve(api_request: api_request)
     @organisation_id = params[:organisation_id]
     render json: { results: @content, organisation_id: @organisation_id }.to_json
   end
--- a/app/controllers/content_controller.rb
+++ b/app/controllers/content_controller.rb
@@ -2,7 +2,7 @@ class ContentController < Api::BaseController
   before_action :validate_params!
 
   def show
-    @content = Queries::FindContent.retrieve(api_request: api_request)
+    @content = Queries::FindContent.retrieve(content_filters: api_request)
     @organisation_id = params[:organisation_id]
     render json: { results: @content, organisation_id: @organisation_id }.to_json
   end
@@ -14,7 +14,7 @@ class ContentController < Api::BaseController
 private
 
   def permitted_params
-    params.permit(:from, :to, :organisation_id, :format)
+    params.permit(:from, :to, :organisation_id, :document_type, :format)
   end
 
   def validate_params!
--- a/app/controllers/content_controller.rb
+++ b/app/controllers/content_controller.rb
@@ -2,7 +2,7 @@ class ContentController < Api::BaseController
   before_action :validate_params!
 
   def show
-    @content = Queries::FindContent.retrieve(filter: api_request.to_filter)
+    @content = Queries::FindContent.call(filter: api_request.to_filter)
     @organisation_id = params[:organisation_id]
 
     render json: { results: @content, organisation_id: @organisation_id }.to_json
--- a/app/controllers/content_controller.rb
+++ b/app/controllers/content_controller.rb
@@ -2,17 +2,20 @@ class ContentController < Api::BaseController
   before_action :validate_params!
 
   def show
-    content = Queries::FindContent.call(filter: api_request.to_filter)
-    organisation_id = params[:organisation_id]
+    filter = api_request.to_filter
+    content = Queries::FindContent.call(filter: filter)
 
-    render json: { results: content, organisation_id: organisation_id }.to_json
+    render json: {
+      results: content,
+      organisation_id: params[:organisation_id]
+    }.to_json
   end
+
+private
 
   def api_request
     @api_request ||= Api::ContentRequest.new(permitted_params)
   end
-
-private
 
   def permitted_params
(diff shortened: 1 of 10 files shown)


peteglondon:
[(2) Index page - filter by document type](https://trello.com/c/7zjqCznt/726-2-index-page-filter-by-document-type)

app/controllers/content_controller.rb
```diff
@@ -2,11 +2,7 @@ class ContentController < Api::BaseController
   before_action :validate_params!
 
   def show
-    @content = Queries::FindContent.retrieve(
-      from: from,
-      to: to,
-      organisation_id: params[:organisation_id]
-    )
+    @content = Queries::FindContent.retrieve(api_request: api_request)
```
theseanything: I feel that using an API request object obsfactates what `FindContent.retrieve` actually needs.

peteglondon: Good point, actually it doesn't even need to do that.
Fixed now.

spec/requests/api/v1/content_spec.rb
```diff
@@ -97,6 +95,23 @@
       json = JSON.parse(response.body).deep_symbolize_keys
       expect(json[:organisation_id]).to eq primary_org_id
     end
+
+    it 'filters by document_type' do
+      get '/content', params: { from: '2018-01-01', to: '2018-09-01', organisation_id: primary_org_id, document_type: 'latest_doc_type' }
+      expect(response.status).to eq(200)
+      json = JSON.parse(response.body).deep_symbolize_keys
+      expect(json[:results]).to eq(
```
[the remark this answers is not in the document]
peteglondon: Good point, actually it doesn't even need to do that. Fixed now -  asserting the count is fine.

app/controllers/content_controller.rb
peteglondon: It's all in a name.  `retrieve` now takes `content_filters`.

pmanrubia:
@theseanything could you review again? Pete seems to have worked after your comments

app/controllers/content_controller.rb
```diff
@@ -2,11 +2,7 @@ class ContentController < Api::BaseController
   before_action :validate_params!
 
   def show
-    @content = Queries::FindContent.retrieve(
-      from: from,
-      to: to,
-      organisation_id: params[:organisation_id]
-    )
+    @content = Queries::FindContent.retrieve(content_filters: api_request)
```
theseanything: I still feel this function arguments are still hidden. It's not clear what parts of an "API request" this function will use or what will be used as filters.

app/domain/queries/find_content.rb
```diff
@@ -62,10 +60,12 @@ def array_to_hash(array)
   end
 
   def slice_editions
-    Dimensions::Edition.by_organisation_id(@organsation_id)
+    editions = Dimensions::Edition.by_organisation_id(@content_filters.organisation_id)
+    editions = editions.where('latest.document_type = ?', @content_filters.document_type) if @content_filters.document_type
```
theseanything: This may exclude editions that are latest for a given time period, but not the latest of all time. E.g. "last-month"
We should also be consistent with how we filter by `organisation_id`.

spec/requests/document_types_spec.rb
```diff
@@ -11,4 +11,33 @@
...
+
+    it "can be cached for up to a day" do
+      Timecop.freeze(Time.zone.local(2020, 1, 1, 1, 0, 1)) do
+        get '/document_types'
+
+        expect(response.headers['ETag']).to be_present
+        expect(response.headers['Cache-Control']).to eq "max-age=86399, public"
+      end
+    end
+  end
```
theseanything: We should use shared examples for these common tests.

pmanrubia:
@theseanything I have finished addressing your comments on top of @peteglondon's PR. Could you please have another look?

spec/domain/queries/find_content_spec.rb
```diff
@@ -84,48 +70,53 @@
...
+          upviews: 200,
+          document_type: 'press_release',
+          satisfaction: 0.5,
+          satisfaction_score_responses: 120,
+          searches: 20,
+        }]
+      )
+    end
+
+    it 'returns items matching the document_type of the latest version' do
```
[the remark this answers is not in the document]
pmanrubia: Not sure how to improve its name. I have followed the same pattern as the previous one. If you have any ideas, feel free to add a commit. In any case, the whole test need some refactoring that will come when we start aggregating content.

spec/domain/api/content_request_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    )
+
+    expect(request.to_filter).to eq(
+      document_type: 'guide',
+      organisation_id: 'the-id',
+      from: '2018-01-01',
+      to: '2018-01-31'
+    )
+  end
+end
```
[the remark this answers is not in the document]
pmanrubia: I have added the missing fields

spec/domain/queries/find_content_spec.rb
```diff
@@ -84,49 +70,49 @@
...
+          satisfaction_score_responses: 120,
+          searches: 20,
+        }]
       )
     end
+
+    it 'returns items matching the document_type of the latest version' do
+      results = described_class.call(filter: filter.merge(document_type: 'press_release'))
+
+      expect(results).to eq([{
```
theseanything: Probably need to use the `include` matcher

PR #965, opened 2018-10-18
Include content_id in Dimensions::Edition metadata

cbaines:
This is needed in the Content Data Admin to generate links to
publishing applications to edit the content (for example Whitehall
Publisher).

https://trello.com/c/pamIZVPi/716-2-page-data-add-link-to-correct-page-in-publishing-tool

### DIFF
--- a/app/models/dimensions/edition.rb
+++ b/app/models/dimensions/edition.rb
@@ -45,6 +45,7 @@ class Dimensions::Edition < ApplicationRecord
     {
       title: title,
       base_path: base_path,
+      content_id: content_id,
       first_published_at: first_published_at,
       public_updated_at: public_updated_at,
       publishing_app: publishing_app,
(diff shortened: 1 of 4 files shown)


app/models/dimensions/edition.rb
```diff
@@ -45,6 +45,7 @@ def metadata
     {
       title: title,
       base_path: base_path,
+      content_id: content_id,
```
pmanrubia: Could you please add a tests to it?
You would need to update:
1. https://github.com/alphagov/content-performance-manager/blob/master/spec/domain/queries/find_metadata_spec.rb
2. https://github.com/alphagov/content-performance-manager/blob/master/spec/requests/api/single_page_spec.rb

pmanrubia: Could you please add a tests to it?
You would need to update:
1. https://github.com/alphagov/content-performance-manager/blob/master/spec/domain/queries/find_metadata_spec.rb
2. https://github.com/alphagov/content-performance-manager/blob/master/spec/requests/api/single_page_spec.rb

cbaines: I've amended this commit to update the relevant tests, it might not show up on GitHub yet, as GitHub still a bit broken...

PR #966, opened 2018-10-22
Sort time series by date

theseanything:
The API needs to respond with the time series in chronological order. Previously data was just returned as ordered within thedatabase. We now sort the time series data by date from oldest to latest.


theseanything:
[(1) Bug - fix date range in charts and table in integration & staging environments](https://trello.com/c/eb2aLLSo/756-1-bug-fix-date-range-in-charts-and-table-in-integration-staging-environments)

PR #968, opened 2018-10-23
Cut down the parts of Rails that are loaded

cbaines:
Some parts of Rails loaded by default are not being used, and probably
won't ever be, like actionmailer for example. So just require the
parts that are used.

The motivation behind this change is to remove activestorage from
being required, as it requires secrets when precompiling assets, which
breaks it in the GOV.UK Mini Environment Admin.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,6 +1,12 @@
 require_relative 'boot'
 
-require 'rails/all'
+require 'rails'
+require 'action_controller/railtie'
+require 'action_view/railtie'
+require 'active_job/railtie'
+require 'active_model/railtie'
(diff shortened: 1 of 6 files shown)


config/application.rb
```diff
@@ -1,6 +1,12 @@
 require_relative 'boot'
 
 require 'rails/all'
```
pmanrubia: Would you need to remove anything here?

cbaines: Haha, yep, need to remove that line!

I've done this now, and made some more changes to make this actually pass the tests. Thanks for taking a look Pablo :)

PR #974, opened 2018-10-30
Add pagination

peteglondon:
Add pagination to /content endpoint.

the /content endpoint now accepts `page` and `page_size' parameters.

The default `page_size` is 100.

Also removed the redundant `organisation_id` from the response.

There is a [Related PR in content-data-admin](https://github.com/alphagov/content-data-admin/pull/169)

## Before
screen shot 2018-10-30 at 11 00 39

## After
screen shot 2018-10-30 at 11 00 50

Url to test: http://content-performance-manager.dev.gov.uk/content?from=2018-09-26&to=2018-10-26&organisation_id=25aacd68-dc0c-4041-9c3a-df59a5357c23&page=1&page_size=10

### DIFF
--- a/app/domain/queries/find_content.rb
+++ b/app/domain/queries/find_content.rb
@@ -1,6 +1,6 @@
 class Queries::FindContent
   def self.call(filter:)
-    filter.assert_valid_keys :from, :to, :organisation_id, :document_type
+    filter.assert_valid_keys :from, :to, :organisation_id, :document_type, :page, :page_size
 
     new(filter).call
   end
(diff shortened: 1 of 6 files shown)


spec/domain/queries/find_content_spec.rb
```diff
@@ -113,6 +113,36 @@
     end
   end
 
+  context 'when pagination parameters are provided' do
```
pmanrubia: Should we test results when . no pagination parameters are provided? We should return first page with default page size.

app/domain/queries/find_content.rb
```diff
@@ -26,6 +27,8 @@ def initialize(filter)
     @to = filter.fetch(:to)
     @organisation_id = filter.fetch(:organisation_id)
     @document_type = filter.fetch(:document_type)
+    @page = filter[:page] || 1
+    @page_size = filter[:page_size] || 100
```
pmanrubia: Shall we extract a constant with the default page size, so we don't have magic numbers?

spec/domain/queries/find_content_spec.rb
peteglondon: Absolutely, That test was on the list but then I forgot to write it.

PR #976, opened 2018-10-30
Add not NULL constraints to withrawn and historical

theseanything:
This adds the migrations to apply not NULL constraints for withdrawn and historical columns in Editions table. We are apply this constraint as we should always know the status of a content item.


theseanything:
[(2) Page data - Display 'withdrawn' and 'history mode'](https://trello.com/c/IFVmpXNs/671-2-page-data-display-withdrawn-and-history-mode)

PR #977, opened 2018-10-31
Return total_pages from /content endpoint

peteglondon:


peteglondon:
[(5) Index page - Add pagination](https://trello.com/c/9SpR0P1I/662-5-index-page-add-pagination)

PR #978, opened 2018-10-31
Improve the performance of /organisations

peteglondon:
The /organisations endpoint has unacceptable performance.

The performance is getting worse as the database grows.

This commit retrieves organisation by returning documents
with a content_type of `organisation`.


peteglondon:
[(5) Index page - Add pagination](https://trello.com/c/9SpR0P1I/662-5-index-page-add-pagination)

PR #992, opened 2018-11-06
Improve readability, maintainability and increase test coverage for the integration with Publishing API

peteglondon:
### What

We need to improve the readability and test coverage of the code that handles the integration with Publishing API

### Why

The code for the integration with Publishing API is key for our Data Warehouse; there are many reasons for it, but the most important one is that is rules how the Editions dimension grow over time.

We need to have code that is easy to read, to maintain and extend; as of today, our code is not good enough, and need some time and attention.


peteglondon:
[(5) Improve readability, maintainability and increase test coverage for the integration with Publishing API](https://trello.com/c/Dfcm3Xzn/793-5-improve-readability-maintainability-and-increase-test-coverage-for-the-integration-with-publishing-api)

PR #997, opened 2018-11-09
Fixes schema.db: materialized views

pmanrubia:
Updates `schema.rb` which was left in an invalid state 
in https://github.com/alphagov/content-performance-manager/pull/995


cbaines:
Do you know if this change get overwritten when the schema is dumped out or updated?

pmanrubia:
@cbaines I have re-run all the migrations from scratch, so my assumption is that I didn't include these changes in my PR, but I might be wrong?

PR #1005, opened 2018-11-16
Improve handling of messages without a locale

cbaines:


cbaines:
This has now been replaced by https://github.com/alphagov/content-performance-manager/pull/1009

PR #1009, opened 2018-11-19
Improve handling of messages without a locale (reworked)

cbaines:
Following on from https://github.com/alphagov/content-performance-manager/pull/1005

Now that the Publishing API will include the locale more often [1], shift to guarding against a missing locale.

1: https://github.com/alphagov/publishing-api/pull/1390
https://github.com/alphagov/publishing-api/pull/1391
https://github.com/alphagov/publishing-api/pull/1392

### DIFF
--- a/app/domain/streams/handlers/single_item_handler.rb
+++ b/app/domain/streams/handlers/single_item_handler.rb
@@ -1,4 +1,7 @@
 class Streams::Handlers::SingleItemHandler < Streams::Handlers::BaseHandler
+  class MissingLocaleError < StandardError
+  end
+
   def self.process(*args)
     new(*args).process
   end
@@ -16,6 +19,8 @@ class Streams::Handlers::SingleItemHandler < Streams::Handlers::BaseHandler
 private
 
   def find_old_edition(content_id, locale)
+    raise MissingLocaleError unless locale
+
     Dimensions::Edition.find_by(content_id: content_id, locale: locale, latest: true)
   end
 end
--- a/spec/domain/streams/handlers/single_item_handler_spec.rb
+++ b/spec/domain/streams/handlers/single_item_handler_spec.rb
@@ -1,0 +1,66 @@
+RSpec.describe Streams::Handlers::SingleItemHandler do
+  let(:content_id) { SecureRandom.uuid }
+  let(:base_path) { '/base-path' }
+  let(:locale) { 'en' }
+
+  let(:old_edition) do
+    create(
+      :edition,
+      content_id: content_id,
+      base_path: base_path,
+      locale: 'en',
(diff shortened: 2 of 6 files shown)


app/domain/streams/handlers/single_item_handler.rb
```diff
@@ -16,6 +19,8 @@ def process
 private
 
   def find_old_edition(content_id, locale)
+    raise MissingLocaleError unless locale
```
pmanrubia: 👍 I like the approach of guarding against unexpected data and deal with the errors in Sentry. It would be good to count how many content items that are not `redirect` or `gone(s)`don't have a locale. It the number is too big, we might have issues when repopulating the Editions dimension.

spec/integration/streams/single/grow_dimension_spec.rb
```diff
@@ -115,4 +115,41 @@
       expect(new_edition).to have_attributes(warehouse_item_id: warehouse_item_id, base_path: new_base_path)
     end
   end
+
+  context 'when handling unpublishing related messages' do
+    it 'it does not grow dimensions_editions unnecessarily' do
+      edition = create :edition
+
+      first_gone_message = build(
```
pmanrubia: What about extracting this `gone` message into a factory so we can reuse it?

spec/domain/streams/handlers/single_item_handler_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+  context 'when attrs differ from the old edition' do
+    context 'in the case of a redirect or gone with no locale' do
+      let(:attrs) do
+        {
+          content_id: content_id,
+          base_path: base_path
+        }
+      end
+
+      it 'finds the old edition' do
```
pmanrubia: The `it` copy does not look right.

app/domain/streams/handlers/single_item_handler.rb
cbaines: > It would be good to count how many content items that are not `redirect` or `gone(s)`don't have a locale. It the number is too big, we might have issues when repopulating the Editions dimension.

I'm pretty sure this is 0. Everything in the Publishing API has a locale.

spec/integration/streams/single/grow_dimension_spec.rb
cbaines: Sure, I've made an attempt at that now :)

spec/domain/streams/handlers/single_item_handler_spec.rb
cbaines: Ah, yep, I've fixed this now :+1:

PR #1011, opened 2018-11-20
Show correct number of PDFs

KyleMacPherson:
[Trello card](https://trello.com/c/7PsDeH5Q/832-3-number-of-pdfs-in-content-data-does-not-match-govuk-page)

We have a bug which shows an incorrect number of PDFs attachments for some content.

This is because we use regex to search for pdf but it is case sensitive.

This PR makes the regex expression case insensitive because some content has
attachments which have the pdf extension in upper case.


KyleMacPherson:
[(3) Number of PDFs in Content Data does not match GOV.UK page](https://trello.com/c/7PsDeH5Q/832-3-number-of-pdfs-in-content-data-does-not-match-govuk-page)

PR #1012, opened 2018-11-20
Allow filtering of index page by date range

KyleMacPherson:
### What?

Allow content endpoint to filter content by date range.

### How?

This PR allows us to select the view we wish to show based on the date range param we pass to the endpoint.

### DIFF
--- a/app/domain/queries/find_content.rb
+++ b/app/domain/queries/find_content.rb
@@ -9,8 +9,9 @@ class Queries::FindContent
   end
 
   def call
-    results = Aggregations::SearchLastThirtyDays.all
-                .joins('INNER JOIN dimensions_editions ON aggregations_search_last_thirty_days.dimensions_edition_id = dimensions_editions.id')
+    view = Queries::SelectView.new(date_range).run
+    results = view[:model_name].all
+                .joins("INNER JOIN dimensions_editions ON aggregations_search_#{view[:table_name]}.dimensions_edition_id = dimensions_editions.id")
                 .merge(slice_editions)
                 .order(order_by)
                 .page(@page)
@@ -37,6 +38,7 @@ private
     @search_term = filter[:search_term]
     @page = filter[:page] || 1
     @page_size = filter[:page_size] || DEFAULT_PAGE_SIZE
+    @date_range = filter.fetch(:date_range)
   end
 
   def aggregates
@@ -44,15 +46,15 @@ private
   end
 
   def array_to_hash(array)
-    satisfaction_responses = array[4] + array[5]
+    satisfaction_responses = array[4].to_i + array[5].to_i
     {
       base_path: array[0],
       title: array[1],
       document_type: array[2],
-      upviews: array[3],
-      satisfaction: satisfaction_responses.zero? ? nil : array[4].to_f / satisfaction_responses,
-      satisfaction_score_responses: satisfaction_responses,
-      searches: array[6],
+      upviews: array[3].to_i,
+      satisfaction: satisfaction_responses.zero? ? nil : (array[4].to_f / satisfaction_responses).to_f,
+      satisfaction_score_responses: satisfaction_responses.to_i,
+      searches: array[6].to_i,
     }
   end
 
--- a/app/domain/queries/select_view.rb
+++ b/app/domain/queries/select_view.rb
@@ -1,0 +1,51 @@
+class Queries::SelectView
+  attr_reader :date_range
+
+  def initialize(date_range)
+    @date_range = date_range
+  end
+
+  def run
+    return GovukError.notify(InvalidDateRangeError.new("Invalid date range: #{date_range}")) unless valid_date_range?
+    { model_name: model_name, table_name: table_name }
+  end
+
+private
+
+  def model_name
+    case date_range
+    when 'last-month'
+      Aggregations::SearchLastMonth
+    when 'last-3-months'
+      Aggregations::SearchLastThreeMonths
+    when 'last-6-months'
+      Aggregations::SearchLastSixMonths
+    when 'last-year'
+      Aggregations::SearchLastTwelveMonths
+    else
+      Aggregations::SearchLastThirtyDays
+    end
+  end
+
+  def table_name
+    case date_range
+    when 'last-month'
+      'last_months'
+    when 'last-3-months'
+      'last_three_months'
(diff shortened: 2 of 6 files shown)


app/domain/queries/find_content.rb
```diff
@@ -9,8 +11,9 @@ def self.call(filter:)
   end
 
   def call
-    results = Aggregations::SearchLastThirtyDays.all
-                .joins('INNER JOIN dimensions_editions ON aggregations_search_last_thirty_days.dimensions_edition_id = dimensions_editions.id')
+    view = select_view
```
pmanrubia: I would isolate this behaviour in another class that can be tested in isolation.
What about something like this:

```ruby
 view_name = QueryPlanner.find_name(date_range)
```

This seems like something that don't belong to the finder; as a result we would also have less tests in the finder, and even the planner could be reused in the future.

app/domain/queries/find_content.rb
```diff
@@ -61,4 +69,20 @@ def slice_editions
...
+    return Aggregations::SearchLastSixMonths if date_range == 'last-6-months'
+    Aggregations::SearchLastTwelveMonths
+  end
+
+  def table_name
+    return 'last_thirty_days' if date_range == 'last-30-days'
+    return 'last_months' if date_range == 'last-month'
+    return 'last_three_months' if date_range == 'last-3-months'
+    return 'last_six_months' if date_range == 'last-6-months'
+    'last_twelve_months' if date_range == 'last-year'
```
pmanrubia: If the `date_range` is none of the expected, I would raise an exception because we are not using the query planner properly.

app/domain/queries/find_content.rb
```diff
@@ -61,4 +69,20 @@ def slice_editions
     editions = editions.where('document_type = ?', document_type) if document_type
     editions
   end
+
+  def model_name
+    return Aggregations::SearchLastThirtyDays if date_range == 'last-30-days'
+    return Aggregations::SearchLastMonth if date_range == 'last-month'
+    return Aggregations::SearchLastThreeMonths if date_range == 'last-3-months'
+    return Aggregations::SearchLastSixMonths if date_range == 'last-6-months'
+    Aggregations::SearchLastTwelveMonths
```
pmanrubia: I would not default to a particular view. It is better to fail than silently returning defaults

spec/domain/queries/find_content_spec.rb
```diff
@@ -58,6 +58,112 @@
     )
   end
 
+  describe 'Other date ranges' do
```
pmanrubia: What about extracting the following in the describe: `Other date ranges`

```ruby
let(:this_month_date) { Date.today }
let(:edition1) { create :edition, base_path: '/path1', date: 2.months.ago, organisation_id: primary_org_id}
let(:edition2) { create :edition, base_path: '/path2', date: 2.months.ago, organisation_id: primary_org_id}
```

Shouldn't we also include here there 30 days?

spec/domain/queries/find_content_spec.rb
```diff
@@ -156,4 +262,51 @@
       expect(-> { described_class.call(filter: filter) }).to raise_error(ArgumentError)
     end
   end
+
+  describe '#select_view' do
```
pmanrubia: As per my comment below I would extract this into the unit tests for QueryPlanner.

pmanrubia:
> The 3 months and longer views return strings but the last month and last 30 days do not.

[You could use this method](https://github.com/alphagov/content-performance-manager/blob/1070f6ff767c976082f8f544110788774a6be7db/app/domain/queries/find_content.rb#L53-L64) to cast into the proper type with: `to_f` and `to_i`.

app/domain/queries/find_content.rb
KyleMacPherson: 👍

app/domain/queries/find_content.rb
KyleMacPherson: I have defaulted the date_range to `last-30-days` because I think we need a default so that we don't always have to pass a date range, for instance when the user firsts visits the page, but I also raise errors if we don't give a valid date_range

app/domain/queries/find_content.rb
KyleMacPherson: 👍 see above

spec/domain/queries/find_content_spec.rb
KyleMacPherson: 👍

spec/domain/queries/find_content_spec.rb
KyleMacPherson: 👍

app/domain/queries/select_view.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def run
+    return GovukError.notify(InvalidDateRangeError.new("Invalid date range: #{date_range}")) unless valid_date_range?
+    { model_name: model_name, table_name: table_name }
+  end
+
+private
+
+  def model_name
+    return Aggregations::SearchLastMonth if date_range == 'last-month'
```
pmanrubia: could you please use a `case` statement?

app/domain/queries/select_view.rb
```diff
@@ -0,0 +1,37 @@
...
+    return Aggregations::SearchLastThreeMonths if date_range == 'last-3-months'
+    return Aggregations::SearchLastSixMonths if date_range == 'last-6-months'
+    return Aggregations::SearchLastTwelveMonths if date_range == 'last-year'
+    Aggregations::SearchLastThirtyDays if date_range == 'last-30-days'
+  end
+
+  def table_name
+    return 'last_months' if date_range == 'last-month'
```
pmanrubia: could you please use a case statement?

db/schema.rb
```diff
@@ -13,6 +13,8 @@
 ActiveRecord::Schema.define(version: 2018_11_22_105601) do
 
   # These are extensions that must be enabled in order to support this database
+  enable_extension "fuzzystrmatch"
```
pmanrubia: Hi @KyleMacPherson you are including changes to schema RB that don't apply to your PR

KyleMacPherson:
[(3) Index page - Make date range filters work on index page](https://trello.com/c/52Imsnnq/841-3-index-page-make-date-range-filters-work-on-index-page)

PR #1013, opened 2018-11-20
Filter out irrelevant content

peteglondon:
We store in the Data Warehouse events that are for "unpublished" state, or that are not relevant for content designers. We need to filter these out from the API's query to the Content Data Admin.

this includes the `/content` and `/document_types` endpoints.

Trello: https://trello.com/c/B0Wt3wFj/685-3-index-page-filter-out-non-relevant-content-items-from-displaying-in-content-data-admin

### DIFF
--- a/app/models/dimensions/edition.rb
+++ b/app/models/dimensions/edition.rb
@@ -21,7 +21,7 @@ class Dimensions::Edition < ApplicationRecord
     latest_by_content_id(content_id, locale)
       .where.not(base_path: exclude_paths)
   end
-
+  scope :live_content, -> { where.not(document_type: %w[redirect gone]) }
   def promote!(old_edition)
     if old_edition
       old_edition.deprecate!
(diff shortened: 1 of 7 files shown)


app/models/dimensions/edition.rb
```diff
@@ -21,7 +21,7 @@ class Dimensions::Edition < ApplicationRecord
     latest_by_content_id(content_id, locale)
       .where.not(base_path: exclude_paths)
   end
-
+  scope :live_content, -> { where.not(document_type: %w[redirect gone]) }
```
cbaines: Maybe keep the newline separating the scopes from the methods?

cbaines: Also, I'm a bit hesitant about "live" content, mainly because the Publishing API uses the term "live" to talk about the "draft" and "live" content stores. Maybe "real" could be used instead?

pmanrubia: You have a point Chris; what I like a bout `live` is that it is use by the product owner, and people around him (non tech people).

When it comes to naming it `real`... not sure about this name to be honest. I would like to explore other names if possible?

What about `editable_content`? this would represent content that is of interest for content designers because they can act on it?

spec/requests/api/v1/content_spec.rb
```diff
@@ -77,6 +77,34 @@
     end
   end
 
+  describe 'Live content' do
```
pmanrubia: Relevant content?

PR #1017, opened 2018-11-22
Search by base_path and title in the Editions table

pmanrubia:
[Trello card](https://trello.com/c/l2T9VoVG/725-5-index-page-filter-by-title-url)

**As a** Content Designer
**I want** to filter the content by title / url
**So that** I can be more precise when searching 

It uses PostgreSQL full text search capabilities to search though all
the base_paths. I have tried with [pg_search][1], but I have finally
decided not to use it because there is no way to skip the sorting by
ranking.

In our scenario, we don't want the results to be sorted by ranking
because we already have an existing sorting criteria. It is just better
to rely on simple text searching with no aditional sorting.

I have created two indexes that significantly improves the performance of
the query.

Important to note here is that when creating the index on the base_path 
I am removing  the `/` because this was issues because the `/` where 
considered letters and making the whole match to fail.

Also important to note how the index and the query share the same way
to remove the `/` so the query planner can effectively choose the right index.

paired with @cbaines 

[1]: https://github.com/Casecommons/pg_search

### DIFF
--- a/app/controllers/content_controller.rb
+++ b/app/controllers/content_controller.rb
@@ -15,7 +15,7 @@ private
   end
 
   def permitted_params
-    params.permit(:from, :to, :organisation_id, :document_type, :format, :page, :page_size, :date_range)
+    params.permit(:from, :to, :organisation_id, :document_type, :format, :page, :page_size, :date_range, :q)
   end
 
   def validate_params!
--- a/app/models/dimensions/edition.rb
+++ b/app/models/dimensions/edition.rb
@@ -23,6 +23,10 @@ class Dimensions::Edition < ApplicationRecord
   end
   scope :relevant_content, -> { where.not(document_type: %w[redirect gone]) }
 
+  def self.search(query)
+    where("to_tsvector('english',title) @@ plainto_tsquery('english', :q)", q: query)
+  end
+
   def promote!(old_edition)
     if old_edition
       old_edition.deprecate!
--- a/app/models/dimensions/edition.rb
+++ b/app/models/dimensions/edition.rb
@@ -24,7 +24,12 @@ class Dimensions::Edition < ApplicationRecord
   scope :relevant_content, -> { where.not(document_type: %w[redirect gone]) }
 
   def self.search(query)
-    where("to_tsvector('english',title) @@ plainto_tsquery('english', :q)", q: query)
+    sql = <<~SQL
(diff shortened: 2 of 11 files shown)


db/migrate/20181121135744_add_full_text_index_title.rb
```diff
@@ -0,0 +1,9 @@
+class AddFullTextIndexTitle < ActiveRecord::Migration[5.2]
+  def up
+    execute "create index dimensions_editions_title on dimensions_editions using gin(to_tsvector('english'::regconfig, COALESCE((title)::text, ''::text)))"
```
cbaines: The expression for the index (`COALESCE((title)::text, ''::text)`) here doesn't exactly match what the search function is using. It would be good to double check that the index is still used when searching?

app/controllers/content_controller.rb
```diff
@@ -15,7 +15,7 @@ def api_request
   end
 
   def permitted_params
-    params.permit(:from, :to, :organisation_id, :document_type, :format, :page, :page_size, :date_range)
+    params.permit(:from, :to, :organisation_id, :document_type, :format, :page, :page_size, :date_range, :q)
```
theseanything: What's q?

pmanrubia: Query for searching? I thought it was quite an standard way... but I might be wrong?

app/models/dimensions/edition.rb
```diff
@@ -23,6 +23,15 @@ class Dimensions::Edition < ApplicationRecord
   end
   scope :relevant_content, -> { where.not(document_type: %w[redirect gone]) }
 
+  def self.search(query)
+    sql = <<~SQL
+      to_tsvector('english',title) @@ plainto_tsquery('english', :q) or
+      to_tsvector('english',replace(base_path, '/', ' ')) @@ plainto_tsquery('english', :q2)
```
theseanything: Will using the 'English' dictionary' effect search with content written in other locales?

pmanrubia: For the time being we are focused on english. For other locales we won't have stemming, but it doesn't feel like something prioritised by the business. Something to discuss in the future, though.

db/migrate/20181121135744_add_full_text_index_title.rb
pmanrubia: @cbaines I haver updated the index. Thank you for spotting this issue.

PR #1018, opened 2018-11-23
Remove defaults from facts_metrics and events_gas

peteglondon:
Why
When repopulating data we are getting zeros in metrics columns
across the dates we have repopulated. Some of these are from before
the item was published.

These columns should only have data when we have actually received it
from the source (i.e. it's been published and we collected some data).

This fixes https://trello.com/c/Wxq2pQxH/890-data-bug-repopulating-1-oct-to-11-nov-has-added-zero-values-for-pages-that-hadnt-been-published-yet

### DIFF
--- a/db/migrate/20181123120108_disable_extensions.rb
+++ b/db/migrate/20181123120108_disable_extensions.rb
@@ -1,0 +1,6 @@
+class DisableExtensions < ActiveRecord::Migration[5.2]
+  def change
+    disable_extension "fuzzystrmatch"
+    disable_extension "pg_trgm"
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_11_23_120108) do
+ActiveRecord::Schema.define(version: 2018_11_23_142713) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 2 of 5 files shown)


db/schema.rb
```diff
@@ -87,11 +87,12 @@
     t.string "previous_version"
     t.string "update_type"
     t.datetime "last_edited_at"
-    t.string "warehouse_item_id", null: false
     t.json "raw_json"
+    t.string "warehouse_item_id", null: false
     t.boolean "withdrawn", null: false
     t.boolean "historical", null: false
     t.index ["base_path"], name: "index_dimensions_editions_on_base_path"
+    t.index ["content_id", "latest"], name: "idx_latest_content_id"
```
pmanrubia: This index is duplicated in the diff?

db/schema.rb
```diff
@@ -200,22 +201,6 @@
   add_foreign_key "facts_metrics", "dimensions_dates", primary_key: "date"
   add_foreign_key "facts_metrics", "dimensions_editions"
 
-  create_view "aggregations_search_last_thirty_days", materialized: true,  sql_definition: <<-SQL
```
pmanrubia: Why do we have changes on the `schema.rb` for the materialized views?

peteglondon:
[Data bug - Repopulating 1 Oct to 11 Nov has added zero values for pages that hadn't been published yet](https://trello.com/c/Wxq2pQxH/890-data-bug-repopulating-1-oct-to-11-nov-has-added-zero-values-for-pages-that-hadnt-been-published-yet)

db/schema.rb
peteglondon: I've no idea. Everything is still there it's generated things in a different order.

db/migrate/20181123120108_disable_extensions.rb
```diff
@@ -0,0 +1,5 @@
+class DisableExtensions < ActiveRecord::Migration[5.2]
+  def change
+    disable_extension "fuzzystrmatch"
+    disable_extension "pg_trgm"  end
```
pmanrubia: `end` should be on the next line

theseanything:
Do we want this to merged in with the `db-schema-updates` feature branch rather than master?

PR #1019, opened 2018-11-23
Renames parameter from `q` to `search_term`

pmanrubia:
[Trello card](https://trello.com/c/l2T9VoVG/725-5-index-page-filter-by-title-url)

Using q is causing issues:

1. It is not obvious to all developers
2. Lint is complaining about being too short.

### DIFF
--- a/app/models/dimensions/edition.rb
+++ b/app/models/dimensions/edition.rb
@@ -25,11 +25,11 @@ class Dimensions::Edition < ApplicationRecord
 
   def self.search(query)
     sql = <<~SQL
-      to_tsvector('english',title) @@ plainto_tsquery('english', :q) or
-      to_tsvector('english'::regconfig, replace((base_path)::text, '/'::text, ' '::text)) @@ plainto_tsquery('english', :q2)
+      to_tsvector('english',title) @@ plainto_tsquery('english', :search_term) or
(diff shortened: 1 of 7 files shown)


app/models/dimensions/edition.rb
```diff
@@ -25,11 +25,11 @@ class Dimensions::Edition < ApplicationRecord
...
   def self.search(query)
     sql = <<~SQL
-      to_tsvector('english',title) @@ plainto_tsquery('english', :q) or
-      to_tsvector('english'::regconfig, replace((base_path)::text, '/'::text, ' '::text)) @@ plainto_tsquery('english', :q2)
+      to_tsvector('english',title) @@ plainto_tsquery('english', :search_term) or
+      to_tsvector('english'::regconfig, replace((base_path)::text, '/'::text, ' '::text)) @@ plainto_tsquery('english', :title_url_2)
     SQL
 
-    where(sql, q: query, q2: query.tr('/', ' '))
+    where(sql, search_term: query, title_url_2: query.tr('/', ' '))
```
cbaines: Maybe use `search_term_without_slash` instead of `title_url_2` here, as I think that would be clearer.

pmanrubia: Thank you. I have addressed your comment 👍

PR #1021, opened 2018-11-26
Add Publishing API events table

theseanything:
This add a Publishing API events table to store messages sent from the Publishing API from which dimension editions are made from. 

Essentially moving the `Dimensions::Edition.raw_json` to `Events::PublishingAPI.payload`. The Publishing API events table also stores the routing key of the original message.


theseanything:
[(3) Extract events table (blocker for populating the data backwards)](https://trello.com/c/2vatqOxj/871-3-extract-events-table-blocker-for-populating-the-data-backwards)

PR #1030, opened 2018-11-28
Add all orgs and no orgs to filter

peteglondon:
# What

Allow organisation_id parameter to accept `all` and `none`.

`all` will not apply the organisation filter.
`none` will return content that has a nil organisation_id.

This is a non-breaking change.

Trello: https://trello.com/c/oWvlfbPR/878-3-index-page-add-all-organisations-and-no-organisations-to-the-organisation-filter

### DIFF
--- a/app/domain/queries/find_content.rb
+++ b/app/domain/queries/find_content.rb
@@ -1,5 +1,7 @@
 class Queries::FindContent
   DEFAULT_PAGE_SIZE = 100
+  ALL = 'all'.freeze
+  NONE = 'none'.freeze
 
   def self.call(filter:)
     raise ArgumentError unless filter.has_key?(:organisation_id) && filter.has_key?(:date_range)
@@ -60,10 +62,10 @@ private
 
   def slice_editions
     editions = Dimensions::Edition.relevant_content
-    if organisation_id == 'none'
+    if organisation_id == NONE
       editions = editions.where('organisation_id IS NULL')
     else
-      editions = editions.where('organisation_id = ?', organisation_id) unless organisation_id == 'all'
+      editions = editions.where('organisation_id = ?', organisation_id) unless organisation_id == ALL
     end
     editions = editions.where('document_type = ?', document_type) if document_type
     editions = editions.search(search_term) if search_term.present?
--- a/app/domain/queries/find_content.rb
+++ b/app/domain/queries/find_content.rb
@@ -64,8 +64,8 @@ private
     editions = Dimensions::Edition.relevant_content
     if organisation_id == NONE
       editions = editions.where('organisation_id IS NULL')
-    else
-      editions = editions.by_organisation_id(organisation_id) unless organisation_id == ALL
+    elsif  organisation_id != ALL
+      editions = editions.by_organisation_id(organisation_id)
     end
     editions = editions.where('document_type = ?', document_type) if document_type
     editions = editions.search(search_term) if search_term.present?
--- a/app/domain/queries/find_content.rb
+++ b/app/domain/queries/find_content.rb
@@ -60,7 +60,11 @@ private
 
   def slice_editions
     editions = Dimensions::Edition.relevant_content
-    editions = editions.where('organisation_id = ?', organisation_id) unless organisation_id == 'all'
+    if organisation_id == 'none'
+      editions = editions.where('organisation_id IS NULL')
+    else
+      editions = editions.where('organisation_id = ?', organisation_id) unless organisation_id == 'all'
+    end
     editions = editions.where('document_type = ?', document_type) if document_type
     editions = editions.search(search_term) if search_term.present?
     editions
--- a/app/domain/queries/find_content.rb
+++ b/app/domain/queries/find_content.rb
@@ -65,7 +65,7 @@ private
     if organisation_id == NONE
       editions = editions.where('organisation_id IS NULL')
     else
-      editions = editions.where('organisation_id = ?', organisation_id) unless organisation_id == ALL
+      editions = editions.by_organisation_id(organisation_id) unless organisation_id == ALL
     end
     editions = editions.where('document_type = ?', document_type) if document_type
     editions = editions.search(search_term) if search_term.present?
--- a/app/domain/queries/find_content.rb
+++ b/app/domain/queries/find_content.rb
@@ -60,7 +60,7 @@ private
 
   def slice_editions
     editions = Dimensions::Edition.relevant_content
-    editions = editions.where('organisation_id = ?', organisation_id)
+    editions = editions.where('organisation_id = ?', organisation_id) unless organisation_id == 'all'
     editions = editions.where('document_type = ?', document_type) if document_type
     editions = editions.search(search_term) if search_term.present?
     editions
--- a/spec/domain/queries/find_content_spec.rb
+++ b/spec/domain/queries/find_content_spec.rb
@@ -156,6 +156,33 @@ RSpec.describe Queries::FindContent do
     end
   end
 
+  describe 'Filter by all organisations' do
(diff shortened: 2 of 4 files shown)


app/domain/queries/find_content.rb
```diff
@@ -60,7 +60,11 @@ def array_to_hash(array)
 
   def slice_editions
     editions = Dimensions::Edition.relevant_content
-    editions = editions.where('organisation_id = ?', organisation_id)
+    if organisation_id == 'none'
+      editions = editions.where('organisation_id IS NULL')
+    else
+      editions = editions.where('organisation_id = ?', organisation_id) unless organisation_id == 'all'
```
pmanrubia: Can you extract a constant for 'all'?

app/domain/queries/find_content.rb
```diff
@@ -60,7 +60,11 @@ def array_to_hash(array)
 
   def slice_editions
     editions = Dimensions::Edition.relevant_content
-    editions = editions.where('organisation_id = ?', organisation_id)
+    if organisation_id == 'none'
```
pmanrubia: What about adding the filter by organisation as a `scope` for editions? It feels like something to be reused in other places?

peteglondon:
[(3) Index page - Add 'all organisations' and 'no organisations' to the organisation filter](https://trello.com/c/oWvlfbPR/878-3-index-page-add-all-organisations-and-no-organisations-to-the-organisation-filter)

app/domain/queries/find_content.rb
peteglondon: We already had one - I just didn't use it.

spec/requests/api/single_page_spec.rb
```diff
@@ -29,6 +29,20 @@
...
+
+      it 'returns data for the homepage' do
+        get "/single_page", params: { from: '2018-01-01', to: '2018-01-31' }
+
+        body = JSON.parse(response.body).deep_symbolize_keys
+        expect(response).to have_http_status(200)
+        expect(body[:metadata][:title]).to eq('GOV.UK homepage')
+      end
+    end
+
```
theseanything: Out of scope for this PR.

spec/routing/single_page_routing_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  it 'routes /single_page (for the homepage)' do
+    expect(get: '/single_page/').to route_to(
+      controller: 'single_item',
+      action: 'show',
+      format: :json
+    )
+  end
+end
```
theseanything: Again out of scope.

spec/domain/queries/find_content_spec.rb
```diff
@@ -156,6 +156,59 @@
...
+  describe 'Filter by all organisations' do
+    let(:filter) do
+      {
+        date_range: 'past-30-days',
+        organisation_id: 'all',
+        document_type: nil
+      }
+    end
+
+    let(:edition1) { create :edition, date: 15.days.ago }
```
theseanything: You can add facts directly to an edition.
```suggestion
    let(:edition1) { create :edition, date: 15.days.ago, facts: { upviews: 20 } }
```

spec/domain/queries/find_content_spec.rb
```diff
@@ -156,6 +156,59 @@
...
+        document_type: nil
+      }
+    end
+
+    let(:edition1) { create :edition, date: 15.days.ago, organisation_id: nil }
+
+    before do
+      create :metric, edition: edition1, date: 15.days.ago
+      edition2 = create :edition, date: 15.days.ago, organisation_id: primary_org_id
+      create :metric, edition: edition2, date: 15.days.ago
```
theseanything: This code isn't consistent with code in the previous test - an edition is being created in the `let` block and the `before` block . Again you can also specify facts directly in the edition.

spec/domain/queries/find_content_spec.rb
```diff
@@ -156,6 +156,59 @@
...
+    let(:filter) do
+      {
+        date_range: 'past-30-days',
+        organisation_id: 'all',
+        document_type: nil
+      }
+    end
+
+    let(:edition1) { create :edition, date: 15.days.ago }
+    let(:edition2) { create :edition, date: 15.days.ago }
```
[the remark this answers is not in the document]
theseanything: We should also be creating the different editions inside the example case (`it` block) as the groups of editions are specific to each example.

spec/requests/api/v1/content_spec.rb
```diff
@@ -240,6 +240,48 @@
...
   end
 
+  describe 'Filter by all organisations' do
+    let(:another_org_id) { 'c97bbc95-967a-4678-848b-6f393171f194' }
+    let(:edition1) { create :edition, organisation_id: organisation_id, date: 15.days.ago }
+    let(:edition2) { create :edition, organisation_id: another_org_id, date: 15.days.ago }
+
+    subject { get '/content', params: { date_range: 'past-30-days', organisation_id: 'all' } }
+
+    before do
```
theseanything: Do we need a `before` block for a single example?

spec/requests/api/v1/content_spec.rb
```diff
@@ -240,6 +240,48 @@
...
+  describe 'Filter by no organisation' do
+    let(:edition1) { create :edition, organisation_id: nil, date: 15.days.ago }
+
+    subject { get '/content', params: { date_range: 'past-30-days', organisation_id: 'none' } }
+
+    before do
+      create :metric, edition: edition1, date: 15.days.ago
+      edition2 = create :edition, organisation_id: organisation_id, date: 15.days.ago
+      create :metric, edition: edition2, date: 15.days.ago
+      recalculate_aggregations!
```
theseanything: This could be tidied up by specify facts when creating the edition and making the code consistent with the test example above.

theseanything:
@pmanrubia @peteglondon I feel the behaviour for ALL org-ids isn't consistent with other filters. As they are filters, the absence of an organisation filter should indicate ALL - (this the behaviour of document_type.)

spec/requests/api/single_page_spec.rb
peteglondon: Shouldn't be in this PR - rebased now.

spec/routing/single_page_routing_spec.rb
peteglondon: Shouldn't be in this PR - rebased now.

spec/requests/api/v1/content_spec.rb
peteglondon: I don't see this as a problem.

spec/requests/api/v1/content_spec.rb
peteglondon: Again we don't need a Facts:Edition for this test.

peteglondon:
> @pmanrubia @peteglondon I feel the behaviour for ALL org-ids isn't consistent with other filters. As they are filters, the absence of an organisation filter should indicate ALL - (this the behaviour of document_type.)

I think this is easier to understand than having a blank parameter. If we want to be consistent we should change the document_type filter.

app/domain/queries/find_content.rb
```diff
@@ -60,7 +62,11 @@ def array_to_hash(array)
 
   def slice_editions
     editions = Dimensions::Edition.relevant_content
-    editions = editions.where('organisation_id = ?', organisation_id)
+    if organisation_id == NONE
+      editions = editions.where('organisation_id IS NULL')
+    else
+      editions = editions.by_organisation_id(organisation_id) unless organisation_id == ALL
```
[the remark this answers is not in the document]
peteglondon: Changed to `elsif`

PR #1031, opened 2018-11-28
Make TransformPath more resilient

cbaines:
The pviews and upviews fields are nullable in the database, so don't
assume that they won't be nil here.

I think it's useful to avoid inventing a 0 value out of nowhere, so
the 0 value is only used as a default if either of the fields actually
have a value.


pmanrubia:
Thank you @cbaines for your work on this 👍 . We are finally adding database constraints so we don't have `nil` on the metrics, so we would not need to guard agains it.

PR #1033, opened 2018-11-29
Add a new route to handle the homepage.

peteglondon:
Now we have allowed filtering by no organisation or
all organisations, the GOV.UK homepage now appears in the
results. This has highlighted a bug. The homepage's base_path
`/` doesn't map to a route in either app.

This commit introduces a route without the base path, which allows
the GOV.UK homepage to routed.

There is a [Related PR in content-data-admin](https://github.com/alphagov/content-data-admin/pull/248)

Trello: https://trello.com/c/q9YG0Ftm/915-fix-bug-with-rendering-the-govuk-homepage-in-content-data-admin

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,7 +13,7 @@ Rails.application.routes.draw do
   end
 
   get '/content', to: 'content#show'
-  get '/single_page/*base_path', to: 'single_item#show', defaults: { format: :json }
+  get '/single_page/(*base_path)', to: 'single_item#show', defaults: { format: :json }
   get '/organisations', to: 'organisation#index', defaults: { format: :json }
   get '/document_types', to: 'document_type#index', defaults: { format: :json }
 end
--- a/spec/routing/single_page_routing_spec.rb
+++ b/spec/routing/single_page_routing_spec.rb
@@ -1,0 +1,18 @@
+RSpec.describe 'single page endpoint routing' do
+  it 'routes /single-item/very/long/base/path' do
+    expect(get: '/single_page/very/long/base/path').to route_to(
+      controller: 'single_item',
+      action: 'show',
+      format: :json,
+      base_path: 'very/long/base/path'
+    )
+  end
+
+  it 'routes /single_page (for the homepage)' do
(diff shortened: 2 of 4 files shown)


config/routes.rb
```diff
@@ -14,6 +14,7 @@
 
   get '/content', to: 'content#show'
   get '/single_page/*base_path', to: 'single_item#show', defaults: { format: :json }
+  get '/single_page', to: 'single_item#show', defaults: { base_path: '', format: :json }
```
chao-xian: What are the downsides to setting a default base_path value of "/"?

peteglondon: The code will pre-pend the leading `/` before querying by :base_path

config/routes.rb
```diff
@@ -14,6 +14,7 @@
 
   get '/content', to: 'content#show'
   get '/single_page/*base_path', to: 'single_item#show', defaults: { format: :json }
```
cbaines: How about changing this to `get '/single_page*base_path'`, would that work?

All the base_path's in the Publishing API start with `/`, so this would be more consistent generally.

peteglondon: It didn't work but now I've found a better solution.

peteglondon:
[Fix bug with rendering the GOV.UK homepage in content-data-admin](https://trello.com/c/q9YG0Ftm/915-fix-bug-with-rendering-the-govuk-homepage-in-content-data-admin)

spec/routing/single_page_routing_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+      format: :json,
+      base_path: 'very/long/base/path'
+    )
+  end
+
+  it 'routes /single_page (for the homepage)' do
+    expect(get: '/single_page/').to route_to(
+      controller: 'single_item',
+      action: 'show',
+      format: :json
```
theseanything: should we add `base_path: ""`?

peteglondon: The assertion checks for the exact params. the base_path param is missing in this instance.

PR #1042, opened 2018-12-04
Filter out irrelevant content part 2

peteglondon:
#What
We show some document types in the filter list that aren't relevant for publishers. We need to filter these out from the API's query to the Content Data Admin and remove them from the document type filter list

We need to filter out:

- `vanish`
- `unpublishing`
- `need`

#Why
Content Data should only show content that is relevant to publishers.

This commit filters the following document types
from the /content and /document_types endpoints.

* vanish
* unpublishing
* need

### Before
screen shot 2018-12-04 at 14 28 54
### After
screen shot 2018-12-04 at 14 36 14

### DIFF
--- a/app/models/dimensions/edition.rb
+++ b/app/models/dimensions/edition.rb
@@ -22,7 +22,7 @@ class Dimensions::Edition < ApplicationRecord
     latest_by_content_id(content_id, locale)
       .where.not(base_path: exclude_paths)
   end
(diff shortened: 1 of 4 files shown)


peteglondon:
[(2) Index page - Filter out non-relevant content items from displaying in Content Data Admin (part 2)](https://trello.com/c/hxo6rgz5/905-2-index-page-filter-out-non-relevant-content-items-from-displaying-in-content-data-admin-part-2)

PR #1043, opened 2018-12-04
Unique index for views: concurrent refresh of data

pmanrubia:
Add a unique index to the view to allow concurrent rebuild of the data while
the users are reading from it. This is important in order to not blocking users
access.

See Scenic[1] documentation[2]

[1]: https://github.com/scenic-views/scenic
[2]: https://github.com/scenic-views/scenic#what-about-materialized-views

### DIFF
--- a/spec/support/shared/shared_masterialized_views.rb
+++ b/spec/support/shared/shared_masterialized_views.rb
@@ -1,6 +1,6 @@
 RSpec.shared_examples 'a materialized view' do |table_name|
   it 'refresh the materialized view' do
-    expect(Scenic.database).to receive(:refresh_materialized_view).with(table_name, concurrently: false, cascade: false)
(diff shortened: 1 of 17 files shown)


spec/support/shared/shared_masterialized_views.rb
```diff
@@ -1,6 +1,6 @@
 RSpec.shared_examples 'a materialized view' do |table_name|
   it 'refresh the materialized view' do
-    expect(Scenic.database).to receive(:refresh_materialized_view).with(table_name, concurrently: false, cascade: false)
+    expect(Scenic.database).to receive(:refresh_materialized_view).with(table_name, concurrently: false, cascade: true)
```
theseanything:
```suggestion
    expect(Scenic.database).to receive(:refresh_materialized_view).with(table_name, concurrently: true, cascade: false)
```

pmanrubia: I have addressed it, thank you!

PR #1046, opened 2018-12-05
Fix date ranges to accurately reflect dates

theseanything:
Date ranges started a day early and ended on the current day. Fixed date ranges to generate the start date a day later and end date to be yesterday as expected.


theseanything:
[Fix date ranges to start one day later (eg 30 days not 31 days)](https://trello.com/c/Lgzah2uQ/922-fix-date-ranges-to-start-one-day-later-eg-30-days-not-31-days)

PR #1049, opened 2018-12-06
Fix date ranges in views

theseanything:
The views for past 3, 6, 12 months and 30 days contained a previous day at the beginning of the time period. This PR includes migrations to update the views and tests to makes sure views return the correct time period.


theseanything:
[Fix date ranges to start one day later (eg 30 days not 31 days)](https://trello.com/c/Lgzah2uQ/922-fix-date-ranges-to-start-one-day-later-eg-30-days-not-31-days)

PR #1052, opened 2018-12-07
Remove metrics request

theseanything:
No longer used as Aggregation and TimeSeries controllers were removed.


theseanything:
Wrong branch.

PR #1055, opened 2018-12-10
Fix total_count returned from /content

peteglondon:
We have an error where the total_results returned from
this endpoint does not match the number of rows returned.

The cause of this is that `latest: true` is not included
in the criteria in `Queries::FindContent#slice_editions`
and `slice_editions.count` is how we retreive the `total_results`
returned.

Refactored the code to use `slice_editions.latest.count` and
added a test to check that old editions are excluded from the
count.

Trello: https://trello.com/c/cx3X7pIT/934-2-filter-description-says-3-results-instead-of-2-results

# Screenshots

## Before
before

## After
after

### DIFF
--- a/app/domain/queries/find_content.rb
+++ b/app/domain/queries/find_content.rb
@@ -22,7 +22,7 @@ class Queries::FindContent
       results: results.pluck(*aggregates).map(&method(:array_to_hash)),
       page: @page,
       total_pages: results.total_pages,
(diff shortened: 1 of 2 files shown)


peteglondon:
[(2) Filter description says '3 results' instead of '2 results'](https://trello.com/c/cx3X7pIT/934-2-filter-description-says-3-results-instead-of-2-results)

PR #1057, opened 2018-12-11
Fix incorrect pdf_count for mainstream documents

peteglondon:
Mainstream documents are showing 0 for pdf_count.

The cause if this is that other content has the PDF links
under details > documents in the json and are nested under a div
with the class `attachment-details`.

Mainstream documents have the PDF links under details > body
and are nested under a div with the class `form-download`.

This commit changes the number_of_pdfs processor to find
pdf links in this location.

Trello: https://trello.com/c/YH8MUtuv/936-2-no-of-pdfs-not-working-on-mainstream

### DIFF
--- a/app/domain/etl/edition/metadata/number_of_files.rb
+++ b/app/domain/etl/edition/metadata/number_of_files.rb
@@ -5,12 +5,12 @@ module Etl::Edition::Metadata::NumberOfFiles
     filter_links all_links, extensions_regex
   end
 
-  ALL_LINKS_XPATH = "//*[contains(@class, 'attachment-details')]//a/@href".freeze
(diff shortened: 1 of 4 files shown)


app/domain/etl/edition/metadata/number_of_files.rb
```diff
@@ -5,12 +5,12 @@ def self.number_of_files(body, extensions_regex)
     filter_links all_links, extensions_regex
   end
 
-  ALL_LINKS_XPATH = "//*[contains(@class, 'attachment-details')]//a/@href".freeze
+  ALL_LINKS_XPATH = "//*[contains(@class, 'attachment-details') or contains(@class, 'form-download')]//a/@href".freeze
```
pmanrubia: Can you create Migration to fix existing data?

peteglondon: Now changed to a rake task.

PR #1059, opened 2018-12-13
Document import data into VM

peteglondon:
The CPM database is not included in the data import by default.

This commit documents how to import it.

### DIFF
--- a/doc/import_production_data.md
+++ b/doc/import_production_data.md
@@ -1,0 +1,29 @@
+# Import Production Data
+
+
+To import production data, follow the instructions at 
+[the GOV.UK developer docs](https://docs.publishing.service.gov.uk/manual/get-started.html#8-import-production-data)
+
+## Note
+The content_performance_manager database has grown very large and is no longer imported by default.
+
+There are some swiches you can provide to the replication script which are documented when you run:
+
+```
+./replicate-data-local.sh -h
+```
+
+If you want to download/import only the content_performance_manager database, add the following arguments to 
+those given in the instructions for both download and import:
+
+```
+-m -e -q -t -r -i "service-manual-publisher publishing_api link_checker_api email-alert-monitor content_data_admin content_tagger content_publisher organisations-publisher email-alert-api content-register policy-publisher content_audit_tool ckan local-links-manager support_contacts transition"
+```
+
+The important switches are: 
+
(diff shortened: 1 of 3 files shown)


doc/import_production_data.md
```diff
@@ -0,0 +1,23 @@
+# Import Production Data
+
+The content_performance_manager database is no longer imported by default.
+
+To import production data, first either download the data as documented in 
+[the GOV.UK developer docs](https://docs.publishing.service.gov.uk/manual/get-started.html#7-import-production-data)
```
pmanrubia: The link seems to be wrong.
Could you please detail what instructions need to be run in order to download a database that is not downloaded by default?

doc/import_production_data.md
```diff
@@ -0,0 +1,23 @@
+# Import Production Data
+
+The content_performance_manager database is no longer imported by default.
```
theseanything: This is sentence has no context and only makes sense with pre-existing knowledge. I think it would be important to start this documentation with a link to the import production data guide in the tech docs.

doc/import_production_data.md
```diff
@@ -0,0 +1,23 @@
...
+
+To import production data, first either download the data as documented in 
+[the GOV.UK developer docs](https://docs.publishing.service.gov.uk/manual/get-started.html#7-import-production-data)
+
+OR
+
+Get someone to give you a database dump if you don't have access.
+
+Then from inside the development VM run
```
theseanything: This seems to be repeating the tech docs guide - which will be confusing if process changes.

doc/import_production_data.md
```diff
@@ -0,0 +1,23 @@
...
+
+```
+mac$ cd ~/govuk/govuk-puppet/development-vm/replication
+mac$ ./replicate-data-local.sh -s -r -q -t -m -e -d backups/YYYY-MM-DD/ -i "service-manual-publisher publishing_api link_checker_api email-alert-monitor content_data_admin content_tagger content_publisher organisations-publisher email-alert-api content-register policy-publisher content_audit_tool ckan local-links-manager"
+```
```
theseanything: Need to state what this command is doing and how it replaces the instructions within the tech docs.

theseanything:
We should also make a note about the size of download and the time required by the process.

PR #1073, opened 2019-01-02
Search with full protocol

pmanrubia:
[Trello card](https://trello.com/c/FrWQ0ON2/941-2-index-page-search-by-whole-url-in-search-box)

The base_path of the editions does not include the protocol or the domain,
so queries passing full URLs don’t return the expected results.

This commit removes the protocol and the domain from the search terms as
they are not needed.

### DIFF
--- a/app/domain/queries/find_content.rb
+++ b/app/domain/queries/find_content.rb
@@ -1,8 +1,4 @@
 class Queries::FindContent
-  DEFAULT_PAGE_SIZE = 100
-  ALL = 'all'.freeze
-  NONE = 'none'.freeze
-
   def self.call(filter:)
     raise ArgumentError unless filter.has_key?(:organisation_id) && filter.has_key?(:date_range)
 
@@ -28,6 +24,10 @@ class Queries::FindContent
   end
 
 private
+
+  DEFAULT_PAGE_SIZE = 100
+  ALL = 'all'.freeze
+  NONE = 'none'.freeze
 
   def order_by
     'upviews desc'
--- a/app/domain/queries/find_content.rb
+++ b/app/domain/queries/find_content.rb
@@ -38,10 +38,19 @@ private
   def initialize(filter)
     @organisation_id = filter.fetch(:organisation_id)
     @document_type = filter.fetch(:document_type)
-    @search_term = filter[:search_term]
+    @search_term = parse_search_term(filter[:search_term]) if filter[:search_term]
     @page = filter[:page] || 1
     @page_size = filter[:page_size] || DEFAULT_PAGE_SIZE
     @date_range = filter.fetch(:date_range)
+  end
+
+  def parse_search_term(search_term)
+    protocol = /http(s)?:\/\//
+    domain = /(www\.)?gov\.uk/
+
+    search_term.
+      gsub(protocol, '').
+      gsub(domain, '')
   end
 
   def aggregates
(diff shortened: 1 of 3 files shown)


app/domain/queries/find_content.rb
```diff
@@ -38,12 +38,21 @@ def order_by
...
-    @search_term = filter[:search_term]
+    @search_term = parse_search_term(filter[:search_term]) if filter[:search_term]
     @page = filter[:page] || 1
     @page_size = filter[:page_size] || DEFAULT_PAGE_SIZE
     @date_range = filter.fetch(:date_range)
   end
 
+  def parse_search_term(search_term)
+    protocol = /http(s)?:\/\//
+    domain = /gov.uk/
```
tijmenb: This might not work because GOV.UK URLs are always start with `www.gov.uk`. In content-tagger we've used `URI` to parse out the path instead:

https://github.com/alphagov/content-tagger/blob/1c4ab6fb7d99558250139a19db6702fadeff1bcf/app/forms/content_lookup_form.rb#L31-L36

pmanrubia: Thank you @tijmenb, I have added this commit (3b526f01e472a3697adbb54b75bf5bbf18421d5a) to ensure that we are considering `www.gov.uk` as a prefix that needs to be removed.

theseanything: Could you combine the into a single regex like `/(www.)?gov.uk/` ?

theseanything: I think you also need to escape `.` - `/(www\.)?gov\.uk/`

spec/domain/queries/find_content_spec.rb
```diff
@@ -348,5 +347,57 @@
...
+
+    describe 'search by base_path with full URLs' do
+      before do
+        edition1 = create :edition, base_path: '/base-path', date: 2.months.ago, organisation_id: primary_org_id
+
+        create :metric, edition: edition1, date: 15.days.ago
+        recalculate_aggregations!
+      end
+
+      let(:results) { described_class.call(filter: filter.merge(search_term: search_term))[:results] }
```
theseanything: I think this should be a `subject` if you use it directly in your expectations

spec/domain/queries/find_content_spec.rb
```diff
@@ -348,5 +347,57 @@
...
+      context 'when search terms include domain www.gov.uk' do
+        let(:search_term) { 'https://www.gov.uk/base-path' }
+
+        it 'return editions matching the base_path' do
+          expect(results).to contain_exactly(
+            hash_including(base_path: '/base-path'),
+          )
+        end
+      end
+    end
```
theseanything: Maybe tests for just `gov.uk`/ `www.gov.uk` without the protocol?

spec/domain/queries/find_content_spec.rb
theseanything: Would it be simpler to unit test the parse_search_term function?

pmanrubia: I would rather keeping `parse_search_term` as `private`. To test it I would need to make it public.

pmanrubia:
@theseanything thank you for your review. I have addressed your comments.

app/domain/queries/find_content.rb
```diff
@@ -38,12 +38,21 @@ def order_by
...
-    @search_term = filter[:search_term]
+    @search_term = parse_search_term(filter[:search_term]) if filter[:search_term]
     @page = filter[:page] || 1
     @page_size = filter[:page_size] || DEFAULT_PAGE_SIZE
     @date_range = filter.fetch(:date_range)
   end
 
+  def parse_search_term(search_term)
+    protocol = /http(s)?:\/\//
+    domain = /(www\.)?gov.uk/
```
theseanything: Think you need to escape the second `.`
```suggestion
    domain = /(www\.)?gov\.uk/
```

PR #1074, opened 2019-01-02
Add /organisations endpoint in v1

theseanything:
This PR addresses the following:
- Add organisations endpoint to the /v1 path. 
- Restructured the organisations controller to use ActiveModel serialization.
- Now have the option to specify locales, with english being returned by default.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ gem 'plek'
 
 # Third party gems
 gem 'activerecord-import'
+gem 'active_model_serializers'
 gem 'awesome_print'
 gem 'chartkick'
 gem 'google-api-client', '~> 0.27'
--- a/spec/models/organisation_spec.rb
+++ b/spec/models/organisation_spec.rb
@@ -1,0 +1,41 @@
+RSpec.describe Organisation do
+  describe '.find_all' do
+    context 'sorting' do
+      before do
+        create(:edition, document_type: 'organisation', title: 'Organisation C')
+        create(:edition, document_type: 'organisation', title: 'Organisation A')
+        create(:edition, document_type: 'organisation', title: 'Organisation B')
+      end
+
+      it 'returns a list of organisations' do
+        organisations = Organisation.find_all
+        expect(organisations).to all be_a(Organisation)
+      end
+
+      it 'returns a list sorted by title' do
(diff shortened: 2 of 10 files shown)


spec/models/organisation_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      it 'returns a list of organisations' do
+        organisations = Organisation.find_all
+        expect(organisations).to all be_a(Organisation)
+      end
+
+      it 'returns a list sorted by title' do
+        organisations = Organisation.find_all
+
+        expect(organisations).to match_array([
```
pmanrubia: `match_array` ignores the order. So `[1,2,3]` would match `[2,3,1]`. Can you think of a better way to ensure that the order is being kept.

spec/models/organisation_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+          have_attributes(name: 'Organisation B'),
+          have_attributes(name: 'Organisation C')
+        ])
+      end
+    end
+
+    context 'locales' do
+      before do
+        create(:edition, document_type: 'organisation', content_id: '1', title: 'English Name', locale: 'en')
+        create(:edition, document_type: 'organisation', content_id: '1', title: 'Enw Cymraeg', locale: 'cy')
```
pmanrubia: Shouldn't they have a different `content_id`?

theseanything: We infer `organisation id` from the `content_id`. So these tests represent when we have the same organisation (i.e. `organisation_id = 1`) with multiple locales.

Gemfile
```diff
@@ -12,6 +12,7 @@ gem 'plek'
 
 # Third party gems
 gem 'activerecord-import'
+gem 'active_model_serializers', '~> 0.10.0'
```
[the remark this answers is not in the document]
theseanything: Removed the version specifier.

spec/models/organisation_spec.rb
theseanything: Implemented sorted_by matcher.

theseanything:
Split into separate commits.

pmanrubia:
👍

PR #1087, opened 2019-01-10
Trap errors in etl processes

peteglondon:
# What
Traps errors during Etl processing

# Why

We want the master processor to continue if there are any errors
with the individual processes.

Trello: https://trello.com/c/8PqhrsdK/972-3-master-process-do-not-stop-if-a-single-etl-process-fails-to-run


peteglondon:
[(3) Master process: do not stop if a single ETL process fails to run](https://trello.com/c/8PqhrsdK/972-3-master-process-do-not-stop-if-a-single-etl-process-fails-to-run)

PR #1088, opened 2019-01-10
Fix /organisations endpoint returning Welsh versions

peteglondon:
# What
Filter out non `en` versions from `/content` endpoint

# Why

We are returning Welsh versions of organisations from the CPM /organisations endpoint. This is causing CDA to display the welsh version in the table header if it's higher in the list (alphabetically) as it just picks the first one with the correct content_id.

Trello: https://trello.com/c/LnIAKNN5/1021-fix-bug-bug-with-welsh-organisation-names

### DIFF
--- a/app/domain/queries/find_all_organisations.rb
+++ b/app/domain/queries/find_all_organisations.rb
@@ -5,7 +5,7 @@ class Queries::FindAllOrganisations
 
   def retrieve
     Dimensions::Edition.latest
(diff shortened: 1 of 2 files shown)


peteglondon:
[Fix bug bug with welsh organisation names.](https://trello.com/c/LnIAKNN5/1021-fix-bug-bug-with-welsh-organisation-names)

PR #1098, opened 2019-01-16
Remove timestamps from facts_metrics

cbaines:
These are unused, and removing them might reduce the disk space used
by this table (~44GB in production).


cbaines:
I haven't tested this, just wanted to put the idea out there, and get people's thoughts on it...

pmanrubia:
I think that removing them is a good idea but I am more keen towards keeping those attributes until we are un public beta so we can use them in case we want to debug / trace issues with the ETL processes. 

I don't have a strong view though, so happy to remove them if everyone agrees.

cbaines:
> I think that removing them is a good idea but I am more keen towards keeping those attributes until we are un public beta so we can use them in case we want to debug / trace issues with the ETL processes.

That sounds sensible, I'll close this until we're ready to look at it again.

PR #1111, opened 2019-01-26
Add indexes to optimise searches across all orgs

chao-xian:
# What/Why

We are seeing a timeout on production when we search for an item
from all organisations. The slowest aspect of the view we have on
the index page is sorting by upviews. Adding an index speeds it up
overall.


chao-xian:
[(3) Bug - timeout when querying all content filtering by search_term](https://trello.com/c/dL9n9KJb/1081-3-bug-timeout-when-querying-all-content-filtering-by-searchterm)

theseanything:
Do we have performance issues if we sort by other metrics i.e. other than unique pageviews?

chao-xian:
Closing this PR as it's not likely to work. Have another PR incoming. Thanks for your feedback @theseanything

PR #1113, opened 2019-01-28
Remove custom sql for all_document_types

chao-xian:
There is a timeout issue on the production database where the query
hangs on the RECURSIVE command. We are removing this to get rid of
timeouts that users are seeing.


chao-xian:
[(3) Bug - timeout when querying all content filtering by search_term](https://trello.com/c/dL9n9KJb/1081-3-bug-timeout-when-querying-all-content-filtering-by-searchterm)

PR #1123, opened 2019-02-04
Remove supergroups and subgroups

theseanything:
This removes the supergroup and subgroup attributes from the dimensions editions table and prevent the stream process from extracting the attributes from publishing API messages.


theseanything:
[(2) Remove Supertypes in Content Performance Manager](https://trello.com/c/FmZsmUim/1098-2-remove-supertypes-in-content-performance-manager)

PR #1129, opened 2019-02-06
Add reading time

chao-xian:
Calculates and returns the reading time for content based on the number of words and an assumed reading rate of 200 wpm.

Note: I initially set the db storage type to integer, but as we need to round up < 1 minute values to 1, and also because we may want to store smaller values in future (I'd already redone the migration a couple of times) it's easier to store the value as a float at this point.

### DIFF
--- a/app/models/facts/metric.rb
+++ b/app/models/facts/metric.rb
@@ -9,7 +9,9 @@ class Facts::Metric < ApplicationRecord
     :readability,
     :chars,
     :sentences,
-    :words, to: :facts_edition
+    :words,
+    :reading_time,
+    to: :facts_edition
 
   validates :dimensions_date, presence: true
   validates :dimensions_edition, presence: true
--- a/db/migrate/20190204150122_add_reading_time_to_facts_edition.rb
+++ b/db/migrate/20190204150122_add_reading_time_to_facts_edition.rb
@@ -1,0 +1,5 @@
+class AddReadingTimeToFactsEdition < ActiveRecord::Migration[5.2]
+  def change
+    add_column :facts_editions, :reading_time, :integer
+  end
+end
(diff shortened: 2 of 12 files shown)


chao-xian:
[(2) Calculate reading time metric (backend)](https://trello.com/c/24kx1Ytt/1107-2-calculate-reading-time-metric-backend)

db/migrate/20190204150122_add_reading_time_to_facts_edition.rb
```diff
@@ -0,0 +1,5 @@
+class AddReadingTimeToFactsEdition < ActiveRecord::Migration[5.2]
+  def change
+    add_column :facts_editions, :reading_time, :float
```
theseanything: Why use float for discrete values? - I assume the smallest resolution we care about is per second, which we could use an integer - Or could we use the Postgres interval type?

chao-xian: We always need to round up to 1 minutes if the actual reading time is less than that (from conversation with PA and CD) but I'm now thinking just storing as seconds in integer is a better option. Will rework when I get a chance after 2ndline.

app/models/facts/metric.rb
```diff
@@ -9,7 +9,8 @@ class Facts::Metric < ApplicationRecord
     :readability,
     :chars,
     :sentences,
-    :words, to: :facts_edition
+    :words,
```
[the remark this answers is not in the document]
chao-xian: It's ok! But I'll drop it to a newline.

db/migrate/20190204150122_add_reading_time_to_facts_edition.rb
chao-xian: Stored as minutes in integer.

PR #1131, opened 2019-02-06
Healthcheck: Google Analytics core metrics

pmanrubia:
[Trello card](https://trello.com/c/lJ0YzNDt/1145-3-raise-2ndline-alarms-when-the-core-google-analytics-metrics-are-not-collected-from-google)

Raises `2ndline` alarms, following the similar approach to `daily_metrics_check.rb`, to notify when metrics collected from `Etl::GA::ViewsAndNavigationProcessor` are not present.

### Why

If the metrics are not present, then the ETL master process failed, and 2nd line needs to take an action

**Note**

I am adding migrations in a separate PR

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -148,7 +148,7 @@ GEM
       bootstrap-sass (= 3.3.5.1)
       jquery-rails (~> 4.3.1)
       rails (>= 3.2.0)
-    govuk_app_config (1.11.2)
+    govuk_app_config (1.11.3)
       aws-xray-sdk (~> 0.10.0)
       logstasher (~> 1.2.2)
       sentry-raven (~> 2.7.1)
@@ -264,7 +264,7 @@ GEM
       rack (>= 1.2, < 3)
     odyssey (0.2.0)
       require_all
-    oj (3.7.6)
+    oj (3.7.8)
     omniauth (1.9.0)
       hashie (>= 3.4.6, < 3.7.0)
       rack (>= 1.6.2, < 3)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -359,6 +359,9 @@ GEM
     rspec-expectations (3.8.2)
       diff-lcs (>= 1.2.0, < 2.0)
       rspec-support (~> 3.8.0)
+    rspec-its (1.2.0)
+      rspec-core (>= 3.0.0)
+      rspec-expectations (>= 3.0.0)
     rspec-mocks (3.8.0)
       diff-lcs (>= 1.2.0, < 2.0)
       rspec-support (~> 3.8.0)
@@ -516,6 +519,7 @@ DEPENDENCIES
   rack-proxy
   rails (~> 5.2)
   rails-controller-testing
+  rspec-its
   rspec-rails
   ruby-progressbar
   sass-rails
--- a/app/domain/healthchecks/concerns/time_range.rb
+++ b/app/domain/healthchecks/concerns/time_range.rb
@@ -1,0 +1,27 @@
+require 'active_support/concern'
+
+module Healthchecks::Concerns::TimeRange
+  extend ActiveSupport::Concern
+
+  included do
+    def enabled?
+      healthchecks_enabled? && within_time_range?
+    end
+
+  private
+
+    def adition_of_metric_values
+      Facts::Metric.for_yesterday.sum(:pviews)
+    end
+
+    def healthchecks_enabled?
+      ENV['ETL_HEALTHCHECK_ENABLED'] == '1'
+    end
+
+    def within_time_range?
+      time = Time.zone.now
+
+      time.hour >= Integer(ENV['ETL_HEALTHCHECK_ENABLED_FROM_HOUR'])
+    end
+  end
+end
--- a/app/domain/healthchecks/daily_metrics_check.rb
+++ b/app/domain/healthchecks/daily_metrics_check.rb
@@ -1,5 +1,7 @@
 module Healthchecks
   class DailyMetricsCheck
+    include Concerns::TimeRange
+
     def name
       :daily_metrics
     end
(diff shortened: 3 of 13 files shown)


spec/domain/healthchecks/database_connection_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  describe '#status' do
+    context 'when the connection is valid' do
+      before { Dimensions::Date.create_with(Date.today) }
+
+      its(:status) { is_expected.to eq(:ok) }
+      its(:message) { is_expected.to be_blank }
+    end
+
+    context 'when the connection is not valid' do
```
theseanything: Not clear what conditions create this context?

spec/domain/healthchecks/etl_google_analytics_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      edition = create :edition
+      create :metric, edition: edition, date: yesterday, pviews: 0
+      create :metric, edition: edition, date: Date.today, pviews: 10
+    end
+
+    its(:status) { is_expected.to eq(:critical) }
+    its(:message) { is_expected.to eq('ETL :: no pviews for yesterday') }
+  end
+
+  context 'when there are no pviews for yesterday' do
```
theseanything: Not sure this context is correct? - Should this be `when there are pviews for yesterday`?

spec/domain/healthchecks/etl_google_analytics_spec.rb
```diff
@@ -0,0 +1,31 @@
+RSpec.describe Healthchecks::EtlGoogleAnalytics do
+  its(:name) { is_expected.to eq(:etl_google_analytics_pviews) }
+
+  include_examples 'Healthcheck enabled/disabled within time range'
+
+  subject { described_class.build(:pviews) }
+
+  let(:yesterday) { Date.yesterday }
+
+  context 'when there are no pviews' do
```
theseanything:
```suggestion
  context 'when there are no pviews for yesterday' do
```

spec/support/shared/shared_healthcheck_enabled.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+
+      context 'out of time range' do
+        let(:today) { Time.new(2018, 1, 15, 1, 0, 0) }
+
+        it { is_expected.to_not be_enabled }
+      end
+    end
+
+    context 'when ETL checks are not enabled' do
```
theseanything: Can we add a test for when `ETL_HEALTHCHECK_ENABLED` is set to another value other than `1`?

Gemfile.lock
```diff
@@ -359,6 +359,9 @@ GEM
     rspec-expectations (3.8.2)
       diff-lcs (>= 1.2.0, < 2.0)
       rspec-support (~> 3.8.0)
+    rspec-its (1.2.0)
```
theseanything: Whats the advantages of using this gem over writing example like `it { is_expected.to have_attributes(status: ok) }`?

spec/domain/healthchecks/database_connection_spec.rb
pmanrubia: It can be invalid because of the database is down, because it has no available connections, because of an issue with the pool configuration,... 

For the sake of this test, if I can't get the first `Date` of the `Dimensions::Date` I don't have a valid connection. This approach is quite common in Connection Pools when they run a very simple query to validate each connection before making it available. 

I didn't worry too much about the performance of this query because it is actually called every 15 seconds, by the way.

If you can think of a better name for the context, just let me know.

app/domain/healthchecks/concerns/time_range.rb
```diff
@@ -0,0 +1,27 @@
+require 'active_support/concern'
+
+module Healthchecks::Concerns::TimeRange
```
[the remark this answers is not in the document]
pmanrubia: I have renamed it to `Deactivable`

app/domain/healthchecks/etl_google_analytics.rb
```diff
@@ -0,0 +1,34 @@
+module Healthchecks
+  class EtlGoogleAnalytics
+    include ActiveModel::Model
+    include Concerns::TimeRange
```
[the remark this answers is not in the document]
pmanrubia: They are used by the `healthcheck` framework defined in `govuk_app_config`. Before raising the alarm it checks for the `enabled?` method.

app/domain/healthchecks/daily_metrics_check.rb
```diff
@@ -1,5 +1,7 @@
 module Healthchecks
   class DailyMetricsCheck
+    include Concerns::TimeRange
```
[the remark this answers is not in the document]
pmanrubia: They are used by the `healthcheck` framework defined in `govuk_app_config`. Before raising the alarm it checks for the `enabled?` method.

app/domain/healthchecks/concerns/time_range.rb
```diff
@@ -0,0 +1,27 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    def enabled?
+      healthchecks_enabled? && within_time_range?
+    end
+
+  private
+
+    def adition_of_metric_values
```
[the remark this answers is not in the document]
pmanrubia: Good spot I have removed it

spec/support/shared/shared_healthcheck_enabled.rb
pmanrubia: I have added the following test case. 

```ruby
    context 'when ETL check is misconfigured' do
      before do
        ENV['ETL_HEALTHCHECK_ENABLED'] = '2'
        ENV['ETL_HEALTHCHECK_ENABLED_FROM_HOUR'] = '9'
      end

      let(:today) { Time.new(2018, 1, 15, 1, 0, 0) }

      it { is_expected.to_not be_enabled }
    end
```

Gemfile.lock
pmanrubia: Not a huge difference, but they read a bit better to me. 
It allows you to do things like:

```ruby
its("phone_numbers.size") { should_not eq(0) }
```

or, it also provide handy methods to pass parameters in a more advanced way than `have_attributes`. [Have a look at the docs](https://github.com/rspec/rspec-its). I personally like it.

I decided to add it, as it is a gem maintain by RSpec team.

theseanything: Sounds good!

spec/domain/healthchecks/database_connection_spec.rb
theseanything: Sorry, I meant there isn't anything in the test setup to signify an error in the database connection. We should probably leave a comment explaining that the absence of records is what we are testing.

PR #1134, opened 2019-02-07
Add reading time metric to /content endpoint

theseanything:
Adding the reading time metric to the /content endpoint. Has only been added to the /single_page endpoint, but is needed by the content-data-admin to export reading time in the search results CSV.

### DIFF
--- a/app/views/content/show.json.jbuilder
+++ b/app/views/content/show.json.jbuilder
@@ -13,6 +13,7 @@ json.results @content_items do |content_item|
   json.searches content_item[:searches]
   json.pdf_count content_item[:pdf_count]
   json.word_count content_item[:words]
+  json.reading_time content_item[:reading_time]
 end
 
 json.page @page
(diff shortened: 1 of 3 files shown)


theseanything:
[(2) Add 'reading time' to the CSV](https://trello.com/c/vJ0z4kvU/1121-2-add-reading-time-to-the-csv)

app/views/content/show.json.jbuilder
```diff
@@ -13,6 +13,7 @@ json.results @content_items do |content_item|
   json.searches content_item[:searches]
   json.pdf_count content_item[:pdf_count]
   json.word_count content_item[:words]
+  json.reading_time content_item[:reading_time]
```
pmanrubia: Can we add a test for this?

theseanything: Extended content tests to check for attribute in reponse

PR #1137, opened 2019-02-08
Update total_results calculation

chao-xian:
Users are seeing a discrepancy in the number provided by
total_results and the actual total returned.

This scenario occurs when a content item has been newly published, but
the ETL process has not yet run. The result set will exclude the new item
until the ETL process has run by the next day, but the total_result
still counts that Edition.

## Before
screenshot at feb 08 11-17-40 pm

## After
screenshot at feb 08 11-18-12 pm

### DIFF
--- a/app/domain/finders/content.rb
+++ b/app/domain/finders/content.rb
@@ -8,20 +8,11 @@ class Finders::Content
   end
 
   def call
-    view = Finders::SelectView.new(date_range).run
-    results = view[:model_name].all
-                .joins("INNER JOIN dimensions_editions ON aggregations_search_#{view[:table_name]}.dimensions_edition_id = dimensions_editions.id")
-                .joins("INNER JOIN facts_editions ON dimensions_editions.id = facts_editions.dimensions_edition_id")
-                .merge(slice_editions)
-                .order(sanitized_order(@sort_attribute, @sort_direction))
-                .page(@page)
-                .per(@page_size)
-                .select(*aggregates)
     {
       results: results.map(&method(:array_to_hash)),
       page: @page,
       total_pages: results.total_pages,
-      total_results: slice_editions.latest.count
+      total_results: total_results
     }
   end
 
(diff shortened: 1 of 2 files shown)


chao-xian:
[(2) Bug - newly published content isn't handled on the index page](https://trello.com/c/KFOltTvC/1066-2-bug-newly-published-content-isnt-handled-on-the-index-page)

app/domain/finders/content.rb
```diff
@@ -44,6 +35,29 @@ def initialize(filter)
...
+      .per(@page_size)
+      .select(*aggregates)
+  end
+
+  def total_results
+    view[:model_name].all
+      .joins("INNER JOIN dimensions_editions ON aggregations_search_#{view[:table_name]}.dimensions_edition_id = dimensions_editions.id")
+      .joins("INNER JOIN facts_editions ON dimensions_editions.id = facts_editions.dimensions_edition_id")
+      .merge(slice_editions)
+      .length
```
pmanrubia: Not 100% if this is still the case, but if you use `length` it will cast to an array and perform a count in memory. I would use `count`.

app/domain/finders/content.rb
```diff
@@ -44,6 +35,29 @@ def initialize(filter)
     @sort_direction = filter.fetch(:sort_direction) || 'desc'
   end
 
+  def view
+    @view ||= Finders::SelectView.new(date_range).run
+  end
+
+  def results
+    view[:model_name].all
```
pmanrubia:
```ruby
    view[:model_name].all
      .joins("INNER JOIN dimensions_editions ON aggregations_search_#{view[:table_name]}.dimensions_edition_id = dimensions_editions.id")
      .joins("INNER JOIN facts_editions ON dimensions_editions.id = facts_editions.dimensions_edition_id")
      .merge(slice_editions)
```

This is the same query for retrieving the metrics and for counting. What about extracting a method?

chao-xian: Thanks Pablo. [Fixex](https://github.com/alphagov/content-performance-manager/pull/1137/commits/b5c918aa9e32fc1e1b230cfaa8bf0c9a9f29c735)

app/domain/finders/content.rb
chao-xian: Thanks Pablo. [Fixed](https://github.com/alphagov/content-performance-manager/pull/1137/commits/b5c918aa9e32fc1e1b230cfaa8bf0c9a9f29c735)

PR #1148, opened 2019-02-13
Upgrade Ruby to 2.6.1

edwardkerry:
Upgrade Ruby to 2.6.1, see commits for more details.


cbaines:
Given these changes are dependant on each other, for example, I would guess that the Dockerfile won't build on 2d7b360014f6f6873f9164e52435fd2c70cb9129 as the Ruby version won't match the `.ruby-version` file, it might be neater to have a single commit, but I'm fine with this as well.

PR #1149, opened 2019-02-13
Inline attributes in joins to speed up performance in `/content` endpoint

pmanrubia:
[Trello card](https://trello.com/c/BeS6Ulyt/1150-5-bug-searching-on-production-times-out)

Queries when finding content performs well, but in some situations, they take 
a bit too much time. This happens when the amount of content items that
are returned is too high.

This is the query that is run on our views (there is a join with dimensions_editions
and facts_editions)

```sql
explain analyze SELECT     base_path,
           title,
           organisation_id,
           document_type,
           "aggregations_search_last_thirty_days"."upviews",
           "aggregations_search_last_thirty_days"."pviews",
           "aggregations_search_last_thirty_days"."useful_yes",
           "aggregations_search_last_thirty_days"."useful_no",
           "aggregations_search_last_thirty_days"."searches",
           "aggregations_search_last_thirty_days"."feedex",
           pdf_count,
           words
FROM       "aggregations_search_last_thirty_days"
INNER JOIN dimensions_editions
ON         aggregations_search_last_thirty_days.dimensions_edition_id = dimensions_editions.id
INNER JOIN facts_editions
ON         dimensions_editions.id = facts_editions.dimensions_edition_id
WHERE      "dimensions_editions"."document_type" NOT IN ('redirect',
                                                         'gone')
AND        (
                      To_tsvector('english',title) @@ plainto_tsquery('english', '/government/publication')
           or         to_tsvector('english'::regconfig, replace((base_path)::text, '/'::text, ' '::text)) @@ plainto_tsquery('english', ' government publication'))
ORDER BY   upviews DESC limit 100 offset 1;
```

I have been running some explain plans for the queries and they are not
performing well because of the Hash Join operations. I have tried to increase
the parameter `work_mem` but the performance is still not optimal.

## Before 

```
 Limit  (cost=204057.75..204058.00 rows=100 width=252) (actual time=9940.531..9940.552 rows=100 loops=1)
   ->  Sort  (cost=204057.74..204272.82 rows=86029 width=252) (actual time=9940.529..9940.545 rows=101 loops=1)
         Sort Key: aggregations_search_last_thirty_days.upviews DESC
         Sort Method: top-N heapsort  Memory: 59kB
         ->  Hash Join  (cost=169052.41..200763.60 rows=86029 width=252) (actual time=9524.722..9902.169 rows=136683 loops=1)
               Hash Cond: (aggregations_search_last_thirty_days.dimensions_edition_id = dimensions_editions.id)
               ->  Seq Scan on aggregations_search_last_thirty_days  (cost=0.00..16052.38 rows=514538 width=56) (actual time=0.022..115.362 rows=514538 loops=1)
               ->  Hash  (cost=165069.85..165069.85 rows=93085 width=220) (actual time=9524.364..9524.364 rows=150004 loops=1)
                     Buckets: 16384 (originally 16384)  Batches: 16 (originally 8)  Memory Usage: 3969kB
                     ->  Hash Join  (cost=134531.15..165069.85 rows=93085 width=220) (actual time=9117.982..9460.695 rows=150004 loops=1)
                           Hash Cond: (facts_editions.dimensions_edition_id = dimensions_editions.id)
                           ->  Seq Scan on facts_editions  (cost=0.00..19457.62 rows=555262 width=16) (actual time=0.012..115.665 rows=556829 loops=1)
                           ->  Hash  (cost=130720.50..130720.50 rows=93332 width=204) (actual time=9116.496..9116.496 rows=150004 loops=1)
                                 Buckets: 32768 (originally 32768)  Batches: 16 (originally 8)  Memory Usage: 3841kB
                                 ->  Bitmap Heap Scan on dimensions_editions  (cost=1101.58..130720.50 rows=93332 width=204) (actual time=125.869..8976.190 rows=150004 loops=1)
                                       Recheck Cond: ((to_tsvector('english'::regconfig, (title)::text) @@ '''/government/publication'''::tsquery) OR (to_tsvector('english'::regconfig, replace((base_path)::text, '/'::text, ' '::text)) @@ '''govern'' & ''public'''::tsquery))
                                       Rows Removed by Index Recheck: 129819
                                       Filter: ((document_type)::text <> ALL ('{redirect,gone}'::text[]))
                                       Rows Removed by Filter: 14840
                                       Heap Blocks: exact=26930 lossy=26487
                                       ->  BitmapOr  (cost=1101.58..1101.58 rows=104922 width=0) (actual time=113.139..113.139 rows=0 loops=1)
                                             ->  Bitmap Index Scan on dimensions_editions_title  (cost=0.00..42.44 rows=325 width=0) (actual time=0.344..0.344 rows=0 loops=1)
                                                   Index Cond: (to_tsvector('english'::regconfig, (title)::text) @@ '''/government/publication'''::tsquery)
                                             ->  Bitmap Index Scan on dimensions_editions_base_path  (cost=0.00..1012.48 rows=104598 width=0) (actual time=112.793..112.793 rows=164874 loops=1)
                                                   Index Cond: (to_tsvector('english'::regconfig, replace((base_path)::text, '/'::text, ' '::text)) @@ '''govern'' & ''public'''::tsquery)
 Planning time: 2.003 ms
 Execution time: 9940.714 ms
(27 rows)

```

## After

```
 Limit  (cost=0.42..3417.66 rows=25 width=252) (actual time=2.416..11.286 rows=25 loops=1)
   ->  Index Scan Backward using search_last_thirty_days_gin_base_path_upviews on aggregations_search_last_thirty_days  (cost=0.42..307415.48 rows=2249 width=252) (actual time=2.415..11.282 rows=25 loops=1)
         Filter: ((to_tsvector('english'::regconfig, (title)::text) @@ '''/government/publication'''::tsquery) OR (to_tsvector('english'::regconfig, replace((base_path)::text, '/'::text, ' '::text)) @@ '''govern'' & ''public'''::tsquery))
         Rows Removed by Filter: 504
 Planning time: 0.996 ms
 Execution time: 11.312 ms
(6 rows)
```

Improving the performance has proven difficult, so I have decided to remove
the joins each time we perform a query and inline the attributes in the views,

This adds a bit of complexity to the views, but the performance improvement
is very good.


app/domain/finders/content.rb
```diff
@@ -22,42 +22,75 @@ def call
...
   def view
-    @view ||= Finders::SelectView.new(date_range).run
+    @view ||= Finders::SelectView.new(filter[:date_range]).run
+  end
+
+  def do_filter
+    scope = view[:model_name].all
+    scope = find_by_search_term(scope)
+    scope = find_by_document_type(scope)
+    scope = find_by_organisation(scope)
```
theseanything: Could we add the conditionals about applying filters within this method, rather than the individual methods? I feel it will make the logic easier to read, as it currently reads that filters are **always** being applied.

app/domain/finders/content.rb
```diff
@@ -22,42 +22,75 @@ def call
   ALL = 'all'.freeze
   NONE = 'none'.freeze
 
-  attr_reader :organisation_id, :document_type, :date_range, :search_term
+  attr_reader :filter
```
theseanything: Is this attribute access from outside this class? if not, can we remove the attr_reader and use the `@filter` instance variable, so we are consistent with date_range, page, etc...

spec/models/dimensions/edition_spec.rb
```diff
@@ -42,17 +42,6 @@
...
-      edition = create :edition, document_type: 'news_story'
-      create :edition, document_type: 'redirect'
-      create :edition, document_type: 'gone'
-      create :edition, document_type: 'vanish'
-      create :edition, document_type: 'unpublishing'
-      create :edition, document_type: 'need'
-      results = subject.relevant_content
-      expect(results).to match_array(edition)
-    end
-
```
theseanything: Does the `relevant_content` method still exist for dimension edition model?

spec/models/dimensions/edition_spec.rb
```diff
@@ -263,93 +252,4 @@
       expect(-> { create :edition, base_path: 'value', latest: false }).to_not raise_error
     end
   end
-
-  describe '.search' do
```
theseanything: Does the `search` method for the dimension editions model still exist?

lib/tasks/repopulate_pdf_count.rake
```diff
@@ -1,15 +0,0 @@
...
-        if edition.publishing_api_event
-          pdf_count = Etl::Edition::Metadata::NumberOfPdfs.parse(edition.publishing_api_event.payload)
-          edition.facts_edition.update!(pdf_count: pdf_count)
-        end
-      end
-      putc '.'
-    end
-    puts '\nDone!\n'
-  end
-end
```
theseanything: Why is this being removed?

app/domain/finders/content.rb
pmanrubia: It is a `private` `attr_reader`. It feels like a good practice to me. It is not accessible from the outside, and we don't overuse instance variables within the class.

spec/models/dimensions/edition_spec.rb
pmanrubia: Not sure if I get your point here. The `relevant_content` is no longer used so I have removed it.

spec/models/dimensions/edition_spec.rb
pmanrubia: Same as above, not sure if I get your point here. I have removed it because the method no longer exists in the `Dimensions::Edition`

lib/tasks/repopulate_pdf_count.rake
pmanrubia: This was a one-off `rake task` to repopulate the pdf count. It does not have a test so I am not feeling comfortable with having rake tasks with no associates tests as we can't guarantee that they are going to work in the mid - long run. 

I would rather removing it, and it we think we need a rake task to repopulate PDF, or other attributes, they we can create one with that purpose in mind (and document and test accordingly)

app/domain/finders/content.rb
```diff
@@ -22,42 +22,75 @@ def call
...
-    @search_term = parse_search_term(filter[:search_term]) if filter[:search_term]
-    @page = filter[:page] || 1
-    @page_size = filter[:page_size] || DEFAULT_PAGE_SIZE
+    @filter = filter
+
+    @page = filter.fetch(:page, 1)
+    @page_size = filter.fetch(:page_size, DEFAULT_PAGE_SIZE)
     @date_range = filter.fetch(:date_range)
     @sort_attribute = filter.fetch(:sort_attribute) || 'upviews'
     @sort_direction = filter.fetch(:sort_direction) || 'desc'
```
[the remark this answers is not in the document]
pmanrubia: Instead of following the same pattern, I am reverting my initial refactor.  This is why: 

The default attribute for `fetch` (the second argument) only applies when the `key` is not present, not when it is `nil`. In our tests we set default values to `nil` so they are failing if I am consistent with the changes to `page` and `page_size`. 

The purpose of this PR is not to refactor this class, so I will keep the scope small by leaving it in its initial state. Thank for your raising this issue.

db/migrate/20190213154809_add_indexes_view_aggregations_search_last_thirty_days.rb
```diff
@@ -0,0 +1,21 @@
+class AddIndexesViewAggregationsSearchLastThirtyDays < ActiveRecord::Migration[5.2]
+  def up
+    execute "create index aggregations_search_last_thirty_days_gin_base_path on aggregations_search_last_thirty_days using gin(to_tsvector('english'::regconfig, replace((base_path)::text, '/'::text, ' '::text)))"
+    execute "create index aggregations_search_last_thirty_days_gin_title on aggregations_search_last_thirty_days  using gin(to_tsvector('english', title ))"
+
+    add_index :aggregations_search_last_thirty_days, :upviews, name: :search_last_thirty_days_gin_base_path_upviews
+    add_index :aggregations_search_last_thirty_days, [:organisation_id, :upviews], name: :search_last_thirty_days_gin_base_path_organisation_id
+    add_index :aggregations_search_last_thirty_days, [:document_type, :upviews], name: :search_last_thirty_days_gin_base_path_document_type
```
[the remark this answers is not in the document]
pmanrubia: We had no indices for the other columns, before. I would rather keeping the scope for this PR as small as possible. 

It seems to me that they should have been added when we added the sorting capabilities for the api endpoint `/content`. I am happy to add new indexes in a separate PR later on, with the goal of keeping this PR as small as possible.

app/domain/finders/content.rb
pmanrubia: I have tried, but it does not read as well (although this is totally subjective). If you have a strong view I will do it, but can you please confirm that you would rather the following?:

```ruby
  def apply_filter
    scope = view[:model_name].all
    scope = find_by_search_term(scope) if filter[:search_term].present?
    scope = find_by_document_type(scope) if filter[:document_type].present?
    scope = find_by_organisation(scope)

    scope
  end
```

lib/tasks/repopulate_pdf_count.rake
theseanything: Just making sure I understand - you would have need modify this rake task to make it compatible with the rest of changes in this PR - but because it doesn't have any tests, you'd rather remove it than make the changes and add tests?

spec/models/dimensions/edition_spec.rb
theseanything: Sorry I missed where you removed them! Makes sense now

spec/models/dimensions/edition_spec.rb
theseanything: Sorry I missed where you removed them! Makes sense now

app/domain/finders/content.rb
theseanything: I'd rather this, as it will also improve the readability of the individual functions themselves.

app/domain/finders/content.rb
```diff
@@ -36,28 +33,62 @@ def initialize(filter)
   end
 
   def view
-    @view ||= Finders::SelectView.new(date_range).run
+    @view ||= Finders::SelectView.new(@filter[:date_range]).run
```
theseanything: Should we use `@date_range`?

lib/tasks/repopulate_pdf_count.rake
pmanrubia: The main reason is not having rake tasks that were created for a migration (a one-off thing) and left without a clear purpose nor proper tests.

I think we should remove any rake task that has no tests, nor is intended for a clear use case.

app/domain/finders/content.rb
pmanrubia: I have updated it. I honestly can't see the obvious benefit, but I would rather updating it and not having a long discussion about it.

lib/tasks/repopulate_pdf_count.rake
theseanything: Okay, in which case it seems out of scope for this PR

PR #1151, opened 2019-02-14
Sort by `warehouse_item_id` to guarantee order in /content endpoint

pmanrubia:
## Issue

When you use the 'Download all data in CSV format' link on the index page to download all content for a selected organisation, there are some duplicate rows in the CSV.

## Solution

We can’t rely on attributes that contain duplicates to sort the data. 
The duplication in those values can make our pagination to return the 
same rows multiple times.

This is the case for content items with no views. For example, if we 
List all the HTML publications, we have more than 8K content items with 
no views so pagination through them won’t be consistent.

This is fixed by sorting by `warehouse_item_id` as a second criteria.

### DIFF
--- a/db/migrate/20190214113305_add_index_views_using_warehouse_item_id.rb
+++ b/db/migrate/20190214113305_add_index_views_using_warehouse_item_id.rb
@@ -1,0 +1,36 @@
+class AddIndexViewsUsingWarehouseItemId < ActiveRecord::Migration[5.2]
+  def change
+    remove_index :aggregations_search_last_twelve_months, name: :search_last_twelve_months_gin_base_path_upviews
+    remove_index :aggregations_search_last_twelve_months, name: :search_last_twelve_months_gin_base_path_organisation_id
+    remove_index :aggregations_search_last_twelve_months, name: :search_last_twelve_months_gin_base_path_document_type
+    remove_index :aggregations_search_last_six_months, name: :search_last_six_months_gin_base_path_upviews
+    remove_index :aggregations_search_last_six_months, name: :search_last_six_months_gin_base_path_organisation_id
+    remove_index :aggregations_search_last_six_months, name: :search_last_six_months_gin_base_path_document_type
+    remove_index :aggregations_search_last_three_months, name: :search_last_three_months_gin_base_path_upviews
+    remove_index :aggregations_search_last_three_months, name: :search_last_three_months_gin_base_path_organisation_id
+    remove_index :aggregations_search_last_three_months, name: :search_last_three_months_gin_base_path_document_type
+    remove_index :aggregations_search_last_months, name: :search_last_month_gin_base_path_upviews
+    remove_index :aggregations_search_last_months, name: :search_last_month_gin_base_path_organisation_id
+    remove_index :aggregations_search_last_months, name: :search_last_month_gin_base_path_document_type
+    remove_index :aggregations_search_last_thirty_days, name: :search_last_thirty_days_gin_base_path_upviews
+    remove_index :aggregations_search_last_thirty_days, name: :search_last_thirty_days_gin_base_path_organisation_id
+    remove_index :aggregations_search_last_thirty_days, name: :search_last_thirty_days_gin_base_path_document_type
+
+
+    add_index :aggregations_search_last_twelve_months, %i[upviews warehouse_item_id], name: :search_last_twelve_months_upviews
+    add_index :aggregations_search_last_twelve_months, %i[organisation_id upviews warehouse_item_id], name: :search_last_twelve_months_organisation_id
+    add_index :aggregations_search_last_twelve_months, %i[document_type upviews warehouse_item_id], name: :search_last_twelve_months_document_type
+    add_index :aggregations_search_last_six_months, %i[upviews warehouse_item_id], name: :search_last_six_months_upviews
+    add_index :aggregations_search_last_six_months, %i[organisation_id upviews warehouse_item_id], name: :search_last_six_months_organisation_id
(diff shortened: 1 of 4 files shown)


db/migrate/20190214113305_add_index_views_using_warehouse_item_id.rb
```diff
@@ -0,0 +1,36 @@
...
+    add_index :aggregations_search_last_three_months, [:document_type, :upviews, :warehouse_item_id], name: :search_last_three_months_gin_base_path_document_type
+    add_index :aggregations_search_last_months, [:upviews, :warehouse_item_id], name: :search_last_month_gin_base_path_upviews
+    add_index :aggregations_search_last_months, [:organisation_id, :upviews, :warehouse_item_id], name: :search_last_month_gin_base_path_organisation_id
+    add_index :aggregations_search_last_months, [:document_type, :upviews, :warehouse_item_id], name: :search_last_month_gin_base_path_document_type
+    add_index :aggregations_search_last_thirty_days, [:upviews, :warehouse_item_id], name: :search_last_thirty_days_gin_base_path_upviews
+    add_index :aggregations_search_last_thirty_days, [:organisation_id, :upviews, :warehouse_item_id], name: :search_last_thirty_days_gin_base_path_organisation_id
+    add_index :aggregations_search_last_thirty_days, [:document_type, :upviews, :warehouse_item_id], name: :search_last_thirty_days_gin_base_path_document_type
```
theseanything: I don't understand the naming behind these indexes? i.e. `search_last_twelve_months_gin_base_path_upviews` what does `gin_base_path` stand for?

db/schema.rb
```diff
@@ -472,9 +472,9 @@
...
   add_index "aggregations_search_last_twelve_months", "to_tsvector('english'::regconfig, replace((base_path)::text, '/'::text, ' '::text))", name: "aggregations_search_last_twelve_months_gin_base_path", using: :gin
-  add_index "aggregations_search_last_twelve_months", ["document_type", "upviews"], name: "search_last_twelve_months_gin_base_path_document_type"
-  add_index "aggregations_search_last_twelve_months", ["organisation_id", "upviews"], name: "search_last_twelve_months_gin_base_path_organisation_id"
-  add_index "aggregations_search_last_twelve_months", ["upviews"], name: "search_last_twelve_months_gin_base_path_upviews"
+  add_index "aggregations_search_last_twelve_months", ["document_type", "upviews", "warehouse_item_id"], name: "search_last_twelve_months_gin_base_path_document_type"
+  add_index "aggregations_search_last_twelve_months", ["organisation_id", "upviews", "warehouse_item_id"], name: "search_last_twelve_months_gin_base_path_organisation_id"
+  add_index "aggregations_search_last_twelve_months", ["upviews", "warehouse_item_id"], name: "search_last_twelve_months_gin_base_path_upviews"
```
theseanything: Why only upviews and not for other metrics?

spec/domain/finders/content_spec.rb
```diff
@@ -235,36 +235,56 @@
...
+        titles = response.fetch(:results).map { |result| result.fetch(:title) }
+        expect(titles).to eq(%w(last middle first))
+      end
     end
 
-    it 'order in ascending' do
-      response = described_class.call(filter: filter.merge(sort_direction: 'asc'))
+    context 'when values repeat' do
+      before do
+        edition1 = create :edition, title: 'middle', warehouse_item_id: 'w2', organisation_id: primary_org_id
```
theseanything: So its less confusing can we use `w1` and `w2` as warehouse_item_ids and `first` and `last` as titles?

spec/domain/finders/content_spec.rb
```diff
@@ -235,36 +235,56 @@
...
+        create :metric, edition: edition2, date: 15.days.ago, upviews: 1
+        recalculate_aggregations!
+      end
+
+      it 'use column `warehouse_item_id` to ensure unique order' do
+        response = described_class.call(filter: filter.merge(sort_attribute: 'upviews'))
+
+        titles = response.fetch(:results).map { |result| result.fetch(:title) }
+        expect(titles).to eq(%w(last middle))
+      end
```
theseanything: Can we test that the order of the warehouse item id also respects the sort direction?

db/migrate/20190214113305_add_index_views_using_warehouse_item_id.rb
pmanrubia: Good spot. I have removed it as they were copy pasted.

db/schema.rb
pmanrubia: Same explanation as in https://github.com/alphagov/content-performance-manager/pull/1149#discussion_r. Please, do let me know if this does not make sense.

spec/domain/finders/content_spec.rb
pmanrubia: The `warehouse_item_id` is not part of the finder response, so asserting on the `title` felt like a sensible way to go. 

I am not very keen to change what the finder returns in order to make a test more readable.

theseanything: Sorry - Just in the scope of this test - I meant it's not clear why edition1's title is `middle` rather than `first` and why we start counting warehouse_item_ids at `w2`.

```ruby
edition1 = create :edition, title: 'first', warehouse_item_id: 'w1', organisation_id: primary_org_id
edition2 = create :edition, title: 'last', warehouse_item_id: 'w2', organisation_id: primary_org_id
```

pmanrubia: I have renamed them to `first`, `last` and `w1`, `w2`

PR #1155, opened 2019-02-18
Add ability to sort by satisfaction

theseanything:
This PR adds the following changes:
- Adds satisfaction column to aggregation views, where satisfaction is calculated within Postgres. For rows with no survey responses, the satisfaction is NULL.
- Removes satisfaction_score_responses from content response as no longer used
- Adds ordering constraint whereby NULLs are always ranked last regardless of direction.

These changes are needed to support sorting by satisfaction score, which previously threw an error. Coercing NULLs to always be last, helps users see valuable results without having to request subsequent pages.

Adding the calculation for satisfaction score for each aggregation view is not ideal - future work should be done to remove this repeated calculation.

### DIFF
--- a/spec/domain/finders/content_spec.rb
+++ b/spec/domain/finders/content_spec.rb
@@ -245,6 +245,33 @@ RSpec.describe Finders::Content do
   end
 
   describe 'Order' do
+    context 'when there are NULLS' do
+      before do
+        edition1 = create :edition, title: 'first', organisation_id: primary_org_id
+        edition2 = create :edition, title: 'second', organisation_id: primary_org_id
+        edition3 = create :edition, title: 'null', organisation_id: primary_org_id
+
+        create :metric, edition: edition1, date: 15.days.ago, useful_yes: 1, useful_no: 0 # satisfaction = 1.0
+        create :metric, edition: edition2, date: 15.days.ago, useful_yes: 0, useful_no: 1 # satisfaction = 0.0
+        create :metric, edition: edition3, date: 15.days.ago, useful_yes: 0, useful_no: 0 # satisfaction = NULL
+        recalculate_aggregations!
+      end
+
+      it 'orders NULL last when in ascending direction' do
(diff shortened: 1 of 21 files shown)


spec/domain/finders/content_spec.rb
```diff
@@ -235,6 +235,33 @@
   end
 
   describe 'Order' do
+    context 'when there are NULLS' do
```
pmanrubia: It is not obvious that we are testing the `satisfaction` attribute. I first thought that we were sorting by `title` and having `'null'` in the title was a test issue.

It is a bit tricky because the `satisfaction` will be `null` only when `useful_yes` and `useful_no` are 0. 

```ruby
create :metric, edition: edition1, date: 15.days.ago, useful_yes: 1, useful_no: 0 # satisfaction: 1
create :metric, edition: edition2, date: 15.days.ago, useful_yes: 0, useful_no: 1 # satisfaction: 0
create :metric, edition: edition3, date: 15.days.ago, useful_yes: 0, useful_no: 0 # satisfaction: NULL
```

If Lint is happy with the code above, it could be an option. Although it would be nice to find a different one that does not depend on coments in code.

theseanything:
> We are calculating in several places the satisfaction score (in the views multiple times, and in the Aggregations finder). I am happy to merge it because we are moving the product forward, but could you please add a note to the PR explaining that we need to address this issue?
> and draft a Story in the planning board to perform the calculation just once?

Added to description and there is a new story on the trello board.
https://trello.com/c/QmCF4qDS/1167-5-fix-repeated-satisfaction-score-calculation-in-views

spec/domain/finders/content_spec.rb
theseanything: Yup agree - I've added the comments like you suggested. I couldn't see an clearer way. I think because no columns in fact_edition are NULLable, the only way to get a NULL in the aggregation views is indirectly via useful_yes/no.

PR #1157, opened 2019-02-20
Add migration to change the upviews and warehouse_item_id indexes

cbaines:
For the content page, including the CSV download, the ordering for the
queries uses DESC NULLS LAST for upviews, and DESC for
warehouse_item_id. This migration changes the btree index on upviews
and warehouse_item_id to match, so that these indexes can be used for
the content endpoint query.


pmanrubia:
👍

PR #1160, opened 2019-02-22
Add docs on how to requeue all content

chao-xian:
This rake task exists in the Publishing API and was created to
bootstrap data in the Data Warehouse. It shouldn't need re-running,
but we need to document how to use it.

### DIFF
--- a/doc/requeue_all_content.md
+++ b/doc/requeue_all_content.md
@@ -1,0 +1,14 @@
+# Re-queue all content from the Publishing API
+
+This is a rake task in the Publishing API. It will queue messages to RabbitMQ that the Content Data API will consume. This was initially created to bootstrap the data in the Content Data API, and shouldn't need re-running.
+
+To run the rake task, use the `run_rake_task` job on the deployment Jenkins and provide the routing key as below:
+
+```bash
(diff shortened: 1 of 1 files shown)


chao-xian:
[(1) Document how to republish all content from publishing-api](https://trello.com/c/N3flndby/943-1-document-how-to-republish-all-content-from-publishing-api)

doc/republish_all_content.md
```diff
@@ -0,0 +1,14 @@
+# Republish all content from the Publishing API
```
cbaines: It would be great if this could be described without using "republish", to avoid confusion with the existing meanings of "republish" (for example, [1][2]). How about "Re-send", "Re-queue" or "Re-present"?

1: https://github.com/alphagov/publishing-api/blob/master/app/commands/v2/republish.rb
2: https://github.com/alphagov/whitehall/blob/master/lib/tasks/publishing_api.rake#L109-L113

chao-xian: Updated - thanks for the suggestion! 1bddfdafc8a44ce5a67e5eb0f832944c72cac10f

PR #1165, opened 2019-02-27
[Do not merge] Decode characters base paths

pmanrubia:
[Trello card](https://trello.com/c/EHrM18LF/1198-2-bug-some-foreign-language-slugs-stretch-the-page-title-column-width-in-the-table)

Some URLs have the following format:

/government/publications/apply-for-a-uk-visa-in-thailand/%E0%B8%81%E0%B8%B2%E0%B8%A3%E0%B8%A2%E0%B8%B7%E0%B9%88%E0%B8%99%E0%B8%84%E0%B8%B3%E0%B8%A3%E0%B9%89%E0%B8%AD%E0%B8%87%E0%B8%82%E0%B8%AD%E0%B8%A7%E0%B8%B5%E0%B8%8B%E0%B9%88%E0%B8%B2%E0%B8%AA%E0%B8%AB%E0%B8%A3%E0%B8%B2%E0%B8%8A%E0%B8%AD%E0%B8%B2%E0%B8%93%E0%B8%B2%E0%B8%88%E0%B8%B1%E0%B8%81%E0%B8%A3

This seems to be related with unescaped characters.

### DIFF
--- a/app/domain/streams/messages/multipart_message.rb
+++ b/app/domain/streams/messages/multipart_message.rb
@@ -51,9 +51,9 @@ module Streams
       part_sub_path = sub_path_for_part(part, index)
 
       if part_sub_path
-        "#{base_path}/#{part_sub_path}"
+        CGI::unescape "#{base_path}/#{part_sub_path}"
       else
-        base_path
+        CGI::unescape base_path
       end
     end
 
--- a/app/domain/streams/messages/single_item_message.rb
+++ b/app/domain/streams/messages/single_item_message.rb
@@ -26,7 +26,7 @@ module Streams
   private
 
     def base_path
-      @payload.fetch('base_path')
+      CGI::unescape @payload.fetch('base_path')
     end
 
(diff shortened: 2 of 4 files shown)


cbaines:
Functionally this looks fine, but I think escaping/unescaping is backwards in the commits and the tests.

The `base_path` coming from the Publishing API is percent encoded (which is described in more detail here [1]), a form of "escaping". So the way to get the actual `base_path` is to unescape the `base_path` to get the non-percent encoded representation. So maybe change the commits and tests to refer to the handling of "percent encoded characters", rather than "unescaped characters" (as they're actually escaped).

1: https://tools.ietf.org/html/rfc3986#page-11

pmanrubia:
I am closing this temporarily as explained in https://github.com/alphagov/content-data-admin/pull/402. Agree with being explicit on the encoding representation in tests and commits (I will keep that in mind once we track this issue).

PR #1166, opened 2019-02-28
Add Postgres calculate satisfaction function

theseanything:
This PR addresses the issue of having multiple statements that calculate satisfaction score within the codebase. The following has been changed:
- Adds migrations to add Postgres function for calculating satisfaction score. Currently all calculations happen within Postgres and this will reduce repetition of the same logic. 
- Remove default and nulls constraint to satisfaction columns in facts_metrics and monthly_aggregation tables. If we are unable to calculate the score, due to insufficient data (e.g. 0 useful yes and 0 useful_no) we should not default the satisfaction score to 0.0. We should use NULL to represent this unknown.
- Updates SQL queries that calculated satisfaction score to use the Postgres function.

### DIFF
--- a/app/domain/etl/aggregations/monthly.rb
+++ b/app/domain/etl/aggregations/monthly.rb
@@ -65,7 +65,7 @@ private
         sum(entrances),
         sum(searches),
         sum(feedex),
-        sum(satisfaction),
+        calculate_satisfaction(sum(useful_yes), sum(useful_no)),
         sum(useful_yes),
         sum(useful_no),
         sum(exits),
--- a/app/domain/etl/ga/user_feedback_processor.rb
+++ b/app/domain/etl/ga/user_feedback_processor.rb
@@ -45,7 +45,7 @@ private
       UPDATE facts_metrics
       SET useful_no = s.useful_no,
           useful_yes = s.useful_yes,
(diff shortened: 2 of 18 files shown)


theseanything:
The biggest issue is satisfaction score needs to be calculated when we refresh the materialised views - if we want sorting/filtering functionality.

Satisfaction score has to be calculated directly from the useful_yes, useful_no for that time period and we can't aggregate existing scores for each day (average of averages problem).

I could separate out the logic for daily metrics into our ETL process, but we'd still have duplicated logic. And seeing as we were already do the calculation in the database at the various stages, I felt encapsulating it in a function was at least better than what we had. 

But agreed - it doesn't feel great having this logic in the database. Its harder to test and ActiveRecord isn't aware of functions, hence doesn't show up in schema.rb and functions aren't restored with `db:schema:load`/`db:reset` (re: failing tests).

pmanrubia:
Hi Sean, thank you. I understand now why you have gone for that approach. 

Thinking about it, I am not keen to merge this PR, as the cost benefit ratio of `duplications` vs `logic in the database` is not worth in my opinion. I would rather living with duplicated logic. We could revisit the issue if the existing code is causing issues.

Sorry about the time that you have put into this PR. It feels like the right solution to prevent duplication, but the cost does not feel worth to me.

PR #1169, opened 2019-03-01
Chomp off long page_paths from GA reports

chao-xian:
During the ETL process, page_paths that have very long query params
(suspicious looking query stuffing) will cause the import to fail as the
index can't store a string value so long. This pragmatic fix will allow
the ETL to complete, as we do not treat URLs with query params as
separate instances of page visits as Google Analytics does so we can
mitigate against this problem.

### DIFF
--- a/app/domain/etl/ga/views_and_navigation_service.rb
+++ b/app/domain/etl/ga/views_and_navigation_service.rb
@@ -1,4 +1,6 @@
 class Etl::GA::ViewsAndNavigationService
+  PAGE_PATH_LENGTH_LIMIT = 1500
+
   def self.find_in_batches(*args, &block)
     new.find_in_batches(*args, &block)
   end
@@ -22,6 +24,8 @@ private
 
   def append_data_labels(values)
     page_path, pviews, upviews, entrances, exits, bounces, page_time = *values
+
+    page_path = page_path.slice(PAGE_PATH_LENGTH_LIMIT) if page_path.length > PAGE_PATH_LENGTH_LIMIT
 
     {
       'page_path' => page_path,
--- a/spec/domain/etl/ga/views_and_navigation_service_spec.rb
+++ b/spec/domain/etl/ga/views_and_navigation_service_spec.rb
@@ -11,23 +11,23 @@ RSpec.describe Etl::GA::ViewsAndNavigationService do
   describe "#find_in_batches" do
     let(:date) { Date.new(2018, 2, 20) }
 
-    before do
-      allow(google_client).to receive(:fetch_all) do
-        [
-          build_report_data(
-            build_report_row(dimensions: %w(/foo), metrics: %w(1 1 1 1 1 1 1 1))
-          ),
-          build_report_data(
-            build_report_row(dimensions: %w(/bar), metrics: %w(2 2 2 2 2 2 2 2))
-          ),
-          build_report_data(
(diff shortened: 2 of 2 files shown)


app/domain/etl/ga/views_and_navigation_service.rb
```diff
@@ -23,6 +23,8 @@ def client
   def append_data_labels(values)
     page_path, pviews, upviews, entrances, exits, bounces, page_time = *values
 
+    page_path = page_path.slice(1500) if page_path.length > 1500
```
pmanrubia: Could you please extract a constant with a meaningful name for the max path length?

spec/domain/etl/ga/views_and_navigation_service_spec.rb
```diff
@@ -68,5 +68,43 @@
...
+
+        expect { |probe| subject.find_in_batches(date: date, batch_size: 1, &probe) }
+          .to yield_with_args(arg)
+      end
+    end
+  end
+
+private
+
+  def junk_page_path
```
pmanrubia: What about generating the string with ruby?

```ruby
str = "/".concat("a" * 1600)
```

app/domain/etl/ga/views_and_navigation_service.rb
chao-xian: Yes sorry for the magic number - [fixed](https://github.com/alphagov/content-performance-manager/commit/cc0c35e8add2d9086b41d0b57b018c967951d20d)

spec/domain/etl/ga/views_and_navigation_service_spec.rb
chao-xian: That's definitely nicer... [fixed](https://github.com/alphagov/content-performance-manager/commit/cc0c35e8add2d9086b41d0b57b018c967951d20d)

chao-xian:
[(3) Bug - Missing GA data for 3 days](https://trello.com/c/7p3mxLsW/1159-2-bug-missing-ga-data-for-3-days)

cbaines:
If I understand the problem correctly, there are errors doing the insert in to the temporary `events_gas` table for the `page_path` field (that contains not only the path (in the URI sense), but query parameters as well), for data in Google Analytics with long query parameter components?

If so, given that the `base_path`s in the Content Performance Manager don't contain query parameters, perhaps we could avoid this issue, and potentially speed up the ETL process by dropping any entry in the Google Analytics data where `page_path` contains query parameters?

Also, as a separate point, I'm a little uncertain about what's the intended and actual behaviour of the `ViewsAndNavigation` service and processor. From an initial reading of the code, it seems like we're only recording metrics where no query parameters are used, so we're ignoring Google Analytics data where query parameters are present. But I don't think this matches what you say in the commit message "we do not treat URLs with query params as separate instances of page visits as Google Analytics does". Have I missed something?

chao-xian:
@cbaines this is a really good point - I was under the misconception that we were dealing with the query parameters in the page_path on load but that doesn't seem to be the [case](https://github.com/alphagov/content-performance-manager/blob/master/app/domain/etl/ga/views_and_navigation_processor.rb#L63) which I think means we are ignoring stats for the same page but not if they have query params. Which to me begs the question that we probably should be stripping the query param. However in the meantime this PR will address the short term problem while we decide if we want to make a change to this behaviour.

PR #1172, opened 2019-03-05
Update Search Aggregation models to inherit from MaterializedView

theseanything:
This PR aims to simplify the logic for the search aggregations models and adds the ability for batch requests within Rails.

The following classes are now subclasses for `MaterializedViews`:
- `SearchLastThirtyDaysMonths`
- `SearchLastMonth`
- `SearchLastThreeMonths`
- `SearchLastSixMonths`
- `SearchLastTwelveMonths`

This allows use to push repeated logic into the parent class.

Also added a the primary_key attribute to the parent class required by Rails to perform batch requests.

### DIFF
--- a/app/models/aggregations/materialized_view.rb
+++ b/app/models/aggregations/materialized_view.rb
@@ -1,6 +1,8 @@
 class Aggregations::MaterializedView < ActiveRecord::Base
   self.abstract_class = true
 
+  self.primary_key = 'warehouse_item_id'
+
   def self.prepare
     increase_work_mem_for_current_transaction
   end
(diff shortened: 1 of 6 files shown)


app/models/aggregations/materialized_view.rb
```diff
@@ -1,4 +1,6 @@
-class Aggregations::MaterializedView
+class Aggregations::MaterializedView < ActiveRecord::Base
```
chao-xian: We should only inherit from `ApplicationRecord` as it's Rails standards - see https://blog.bigbinary.com/2015/12/28/application-record-in-rails-5.html

theseanything: The MaterializeView acts as the "views" ApplicationRecord.

PR #1173, opened 2019-03-05
Enable null values for satisfaction score

chao-xian:
Null values are a valid representation of satisfaction scores, as it
denotes that we did not get any responses for the yes/no survey. This
also fixes the representation of data in the graphs, which until now
have been misleading - setting 0 as a satisfacion score (the default
value) for periods even before the rollout of the yes/no survey is
misleading.

## After
The code changes and running the rake task results in the Content Data Admin's graphs handling null values as blanks on the graph, instead of defaulting to zeros:

screenshot at mar 08 9-54-21 am

### DIFF
--- a/app/domain/etl/ga/user_feedback_processor.rb
+++ b/app/domain/etl/ga/user_feedback_processor.rb
@@ -45,7 +45,7 @@ private
       UPDATE facts_metrics
       SET useful_no = s.useful_no,
           useful_yes = s.useful_yes,
-          satisfaction= s.useful_yes / (s.useful_yes + s.useful_no::float)
+          satisfaction= s.useful_yes / NULLIF((s.useful_yes + s.useful_no::float),0)
       FROM (
         SELECT useful_no,
                useful_yes,
--- a/db/migrate/20190304151614_remove_default_satisfaction.rb
+++ b/db/migrate/20190304151614_remove_default_satisfaction.rb
@@ -1,0 +1,17 @@
+class RemoveDefaultSatisfaction < ActiveRecord::Migration[5.2]
+  def up
+    change_column :facts_metrics, :satisfaction, :float, null: true
+    change_column_default :facts_metrics, :satisfaction, nil
+
+    change_column :aggregations_monthly_metrics, :satisfaction, :float, null: true
+    change_column_default :aggregations_monthly_metrics, :satisfaction, nil
+  end
+
+  def down
+    change_column :facts_metrics, :satisfaction, :float, null: false
+    change_column_default :facts_metrics, :satisfaction, 0.0
+
+    change_column :aggregations_monthly_metrics, :satisfaction, :float, null: false
+    change_column_default :aggregations_monthly_metrics, :satisfaction, 0.0
+  end
+end
--- a/lib/tasks/data_migrations.rake
+++ b/lib/tasks/data_migrations.rake
@@ -1,0 +1,15 @@
+namespace :data_migrations do
+  desc 'Sets all metrics with no useful yes/no responses to have null for satisfaction'
+  task satisfaction_defaults_to_null: :environment do
+    puts "Setting default satisfaction to nil for pages with no responses"
+
+    zero_responses = Facts::Metric.where('useful_yes = 0').where('useful_no = 0')
+
+    zero_responses.in_batches.each_with_index do |metrics, batch_index|
+      puts "Updating #{batch_index} records"
+      metrics.update_all(satisfaction: nil)
+    end
+
+    puts "END"
+  end
+end
--- a/lib/tasks/data_migrations.rake
+++ b/lib/tasks/data_migrations.rake
@@ -3,12 +3,10 @@ namespace :data_migrations do
   task satisfaction_defaults_to_null: :environment do
     puts "Setting default satisfaction to nil for pages with no responses"
 
-    zero_responses = Facts::Metric.where('useful_yes = 0').where('useful_no = 0')
-
-    zero_responses.in_batches.each_with_index do |metrics, batch_index|
-      puts "Updating #{batch_index} records"
-      metrics.update_all(satisfaction: nil)
-    end
+    Facts::Metric
+      .where('useful_yes = 0')
+      .where('useful_no = 0')
+      .update_all(satisfaction: nil)
 
     puts "END"
   end
--- a/spec/domain/etl/ga/user_feedback_processor_spec.rb
+++ b/spec/domain/etl/ga/user_feedback_processor_spec.rb
@@ -87,6 +87,13 @@ RSpec.describe Etl::GA::UserFeedbackProcessor do
 
       expect(fact.reload.satisfaction).to be_within(0.1).of(0.0)
     end
+
+    it 'set `satisfaction = nil` with `useful_yes:0` and `no: 0`' do
+      allow(Etl::GA::UserFeedbackService).to receive(:find_in_batches).and_yield(ga_response(useful_yes: 0, useful_no: 0))
+      described_class.process(date: date)
+
+      expect(fact.reload.satisfaction).to be_nil
+    end
   end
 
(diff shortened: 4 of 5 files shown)


chao-xian:
[(2) User satisfaction chart have a line along zero rather than no line for dates before data exists](https://trello.com/c/r365wSrf/1178-2-user-satisfaction-chart-have-a-line-along-zero-rather-than-no-line-for-dates-before-data-exists)

app/domain/etl/ga/user_feedback_processor.rb
```diff
@@ -45,7 +45,7 @@ def load_metrics_query(date_to_s)
       UPDATE facts_metrics
       SET useful_no = s.useful_no,
           useful_yes = s.useful_yes,
-          satisfaction= s.useful_yes / (s.useful_yes + s.useful_no::float)
+          satisfaction= s.useful_yes / NULLIF((s.useful_yes + s.useful_no::float),0)
```
cbaines: What will happen for pages where the actual value for the day is 0 interactions with the usefulness metrics? In my mind, they should be recorded as 0, not NULL, but I think this would record them as NULL?

theseanything: I'd argue that for days we have zero interactions, we have no data, if we have no data, the satisfaction score is unknown. Saying a page has 0% satisfaction based on no data would be misleading and inaccurate.

db/migrate/20190304151614_remove_default_satisfaction.rb
```diff
@@ -0,0 +1,17 @@
...
+    change_column_default :aggregations_monthly_metrics, :satisfaction, nil
+  end
+
+  def down
+    change_column :facts_metrics, :satisfaction, :float, null: false
+    change_column_default :facts_metrics, :satisfaction, 0.0
+
+    change_column :aggregations_monthly_metrics, :satisfaction, :float, null: false
+    change_column_default :aggregations_monthly_metrics, :satisfaction, 0.0
+  end
```
theseanything: I think Rails see this as an irreversible migration. In which case you could simplify the migration to something like:
```ruby
  def change
    change_column_default :aggregations_monthly_metrics, :satisfaction, nil
    change_column_null :aggregations_monthly_metrics, :satisfaction, true

    change_column_default :facts_metrics, :satisfaction, nil
    change_column_null :facts_metrics, :satisfaction, true
  end
```

or maybe worth leaving a comment?

spec/domain/etl/ga/user_feedback_processor_spec.rb
```diff
@@ -87,6 +87,13 @@
 
       expect(fact.reload.satisfaction).to be_within(0.1).of(0.0)
     end
+
+    it 'set `satisfaction = nil` with `useful_yes:0` and `no: 0`' do
+      allow(Etl::GA::UserFeedbackService).to receive(:find_in_batches).and_yield(ga_response(useful_yes: 0, useful_no: 0))
+      described_class.process(date: date)
+# byebug
```
theseanything: Remove comment

app/domain/etl/ga/user_feedback_processor.rb
cbaines: Ok, yeah, I'm not really sure what was happening before though.

To avoid a division by 0 error, somehow `useful_yes` in `events_gas` must be 0, but `useful_no` must be > 0...

cbaines:
> This also fixes the representation of data in the graphs, which until now
> have been misleading - setting 0 as a satisfacion score (the default
> value) for periods even before the rollout of the yes/no survey is
> misleading.

As far as I can see, these changes only affect new data right? This makes it sound like these changes will also fix the existing values in the `facts_metrics` table.

app/domain/etl/ga/user_feedback_processor.rb
theseanything: missing a few spaces -
```suggestion
          satisfaction = s.useful_yes / NULLIF((s.useful_yes + s.useful_no::float), 0)
```

theseanything:
> > This also fixes the representation of data in the graphs, which until now
> > have been misleading - setting 0 as a satisfacion score (the default
> > value) for periods even before the rollout of the yes/no survey is
> > misleading.
> 
> As far as I can see, these changes only affect new data right? This makes it sound like these changes will also fix the existing values in the `facts_metrics` table.

I'm inclined that we should fix previous data as well. Should be straight forward as a migration (?)

app/domain/etl/ga/user_feedback_processor.rb
chao-xian: @cbaines The NULLIF handles most cases, except `NULL\NULL` which I'll need to think a bit more about.

db/migrate/20190304151614_remove_default_satisfaction.rb
chao-xian: I hit the irreversible problem when it was only within a `change`, which is why explicitly setting what the defaults will be in `up` and `down` will work (and I tested locally).

spec/domain/etl/ga/user_feedback_processor_spec.rb
chao-xian: Yikes

chao-xian:
@cbaines @theseanything sorry my commit message was misleading which I'll address, but I had intended to run a rake task to address the previous data, but add the rake task in a separate PR. Probably better to write that now though.

app/domain/etl/ga/user_feedback_processor.rb
theseanything: If you want to be consistent with the aggregation views:
```sql
CASE
  WHEN useful_yes + useful_no = 0 THEN NULL
  ELSE useful_yes / (useful_yes + useful_no)::float
END
```

theseanything: NULL / NULL should be okay because of the type cast

chao-xian: Ah my local testing omitted the casting. Awesome!

spec/domain/etl/ga/user_feedback_processor_spec.rb
chao-xian: [Fixed](https://github.com/alphagov/content-performance-manager/commit/a5a86fc09a93c3af69e0546605a1d0972eb78c6e)

lib/tasks/data_migrations.rake
```diff
@@ -0,0 +1,15 @@
...
+  desc 'Sets all metrics with no useful yes/no responses to have null for satisfaction'
+  task satisfaction_defaults_to_null: :environment do
+    puts "Setting default satisfaction to nil for pages with no responses"
+
+    zero_responses = Facts::Metric.where('useful_yes = 0').where('useful_no = 0')
+
+    zero_responses.in_batches.each_with_index do |metrics, batch_index|
+      puts "Updating #{batch_index} records"
+      metrics.update_all(satisfaction: nil)
+    end
```
cbaines: Going through in batches is going to be really slow given the number of records in this table, I think the default batch size is 1000, so given that, do you know how many batches are required?

This should work without any batching, and thus should be a lot quicker:
``` ruby
Facts::Metric
  .where(useful_yes: 0, useful_no: 0)
  .update_all(satisfaction: nil)
```

pmanrubia: I am interested in getting your thoughts on how we should tackle these migrations. 

I find value in having a longer rake task with a progress indicator that takes say 30 minutes vs a SQL query that takes 20 silently. 

We could explore also creating a temporary index for queries that expand all the metrics table, then do the change, and then remove the index. 

I am happy with this to go through, but It is not obvious for me that this is a better solution, and it would be nice to explore other options for the future.

cc/ @chao-xian @theseanything @cbaines

chao-xian: I had difficulty testing an index locally as I did try adding it and then running the migration, but my VM ran out of disk space...

PR #1177, opened 2019-03-07
Ignore pages with long paths from GA

theseanything:
Previously we were using slice. Slice uses the character at that position of the string and caused multiple pages to have the same age_path which violated the GA Events table primary key constraint. 

This is a quick fix, based on the assumption long base paths are due to long query strings and the fact we don't use data about paths with query strings. 

We should look for a solution to deal with query strings.

### DIFF
--- a/app/domain/etl/ga/views_and_navigation_service.rb
+++ b/app/domain/etl/ga/views_and_navigation_service.rb
@@ -12,6 +12,7 @@ class Etl::GA::ViewsAndNavigationService
       .flat_map(&method(:extract_rows))
       .map(&method(:extract_dimensions_and_metrics))
       .map(&method(:append_data_labels))
+      .reject(&method(:long_query_string?))
       .map { |h| h['date'] = date.strftime('%F'); h }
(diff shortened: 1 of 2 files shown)


app/domain/etl/ga/views_and_navigation_service.rb
```diff
@@ -39,6 +38,10 @@ def append_data_labels(values)
     }
   end
 
+  def long_page_path?(data)
+    data['page_path'].length > PAGE_PATH_LENGTH_LIMIT
+  end
```
cbaines: I was thinking more along the lines of rejecting anything with query params, e.g.

```ruby
!URI.parse(data['page_path']).query.nil?
```

That's a bit more explicit with respect to our intent, but I'm fine with this approach as well.

cbaines:
There's a type on the commit message `age_path`.

theseanything:
[(2) Repopulate missing data for 4th March 2019](https://trello.com/c/xtZtwfWD/1234-2-repopulate-missing-data-for-4th-march-2019)

PR #1178, opened 2019-03-07
Update Readme for Content Performance Manager

pmanrubia:
[Trello card](https://trello.com/c/gzdx8rs5/1237-1-update-readmemd-in-content-performance-manager)

### What

Update Readme in order to:

- Remove all the information that no longer apply
- Make it smaller and link to the information in other documents
- [Make it more compliant with GOV.UK guidelines for Readme documents](https://docs.publishing.service.gov.uk/manual/readmes.html).


README.md
```diff
@@ -35,12 +35,7 @@ In the development VM, go to:
 
 ```bash
 cd /var/govuk/govuk-puppet/development-vm
-```
-
-Then run:
-
- ```bash
- bowl content-performance-manager
+bundle exec bowl content-performance-manager
```
cbaines: I don't think you should use `bundle exec` with `bowl`, as then things get a bit funny. However, I'm not the best one to ask about this, as I gave up using bowler several years ago...

README.md
```diff
@@ -5,7 +5,7 @@ A data warehouse that stores content and content metrics, to help content owners
 This repository contains:
 - Extract, transform, load (ETL) processes for populating the data warehouse
 - An internal tool for exploring the data (AKA the sandbox)
-- Content performance API ([docs](content-performance-api.publishing.service.gov.uk/#gov-uk-content-performance-api))
+- [An API that exposes metrics and content changes][api-doc]  (content-performance-api.publishing.service.gov.uk/#gov-uk-content-performance-api))
```
cbaines: Is the URL like thing at the end of this line still correct?

pmanrubia:
Thank you for your review @cbaines. I have addressed your comments.

PR #1179, opened 2019-03-07
Add missing `requires` for ETL rake tasks

theseanything:
The etl:repopulate_feedex and the etl:repopulate_searches rake task are missing trace_and_recoverable dependency. This adds in the require.

### DIFF
--- a/app/domain/etl/ga/internal_search_processor.rb
+++ b/app/domain/etl/ga/internal_search_processor.rb
@@ -1,3 +1,4 @@
+require_dependency 'concerns/trace_and_recoverable'
 class Etl::GA::InternalSearchProcessor
   include Concerns::TraceAndRecoverable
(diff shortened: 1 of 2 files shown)


app/domain/etl/ga/internal_search_processor.rb
```diff
@@ -1,3 +1,4 @@
+require_dependency 'concerns/trace_and_recoverable'
```
cbaines: If you think this is the right change, then that's fine :+1: 

But as I don't know much about Rails and Ruby's behaviour in this regard, I'd be interested to know why Rails autoloading isn't handling this, and why this is `require_dependency` rather than just `require`?

theseanything:
[(2) Repopulate missing data for 4th March 2019](https://trello.com/c/xtZtwfWD/1234-2-repopulate-missing-data-for-4th-march-2019)

PR #1184, opened 2019-03-12
Add search aggregations healthcheck

chao-xian:
We generate materialised views for searching acrosss the content. A refresh of the views is triggered when the latest Dimensions::Edition items are generated. This tests that the ETL process for this occurred. If the materialized views are not fresh, and alert is triggered in icinga.


chao-xian:
[(2) Raise 2ndline alarms when the Search aggregations are not collected from Google](https://trello.com/c/66Z6PHSZ/1148-2-raise-2ndline-alarms-when-the-search-aggregations-are-not-collected-from-google)

pmanrubia:
💯

db/views/aggregations_search_last_thirty_days_v08.sql
```diff
@@ -0,0 +1,41 @@
...
+  dimensions_editions.document_type as document_type,
+  dimensions_editions.base_path as base_path,
+  dimensions_editions.organisation_id as organisation_id,
+  dimensions_editions.id as dimensions_edition_id,
+  aggregations.upviews::bigint as upviews,
+  aggregations.pviews::bigint as pviews,
+  aggregations.feedex::bigint as feedex,
+  aggregations.useful_yes::bigint as useful_yes,
+  aggregations.useful_no::bigint as useful_no,
+  CURRENT_DATE as updated_at,
```
cbaines: I'm wondering what the space trade off is here, given that this value is being stored for every single entry in this view (and I see this being added to other views as well).

How big are the views already, and how much space does this look to add?

cbaines: I've come up with a couple of possible alternatives...

PostgreSQL apparently can record some information about commit times, https://www.postgresql.org/docs/9.6/functions-info.html#FUNCTIONS-COMMIT-TIMESTAMP which could be useful, but would require some configuration changes, and I'm also uncertain about the space/performance overhead.

Another option would be to have another table which records the last updated time for each view. That would only require as many rows as we have views, and if we update the table in the same transaction as the materialised view, it should be accurate.

chao-xian: @cbaines Thanks for the suggestions. The views aren't that big and although a separate table is definitely a cleaner approach I don't feel it's enough of a benefit at the moment. Will definitely look into this if it does.

PR #1185, opened 2019-03-12
Add from-to params and batching to rake task

chao-xian:
This rake task has failed to run on integration. Adding from and to
parameters for the update to run across chunked ranges of items
by date should help, as well as allowing us to re-run the task at
specific points where we might she the failure. Also breaking
up the query to run in batches should help mitigate failures,
as rather than running one query to run against a very large
dataset (which the database would also need to store a snapshot
of changes to rollback to) this will make it a series of smaller
operations.


chao-xian:
[(3) User satisfaction chart have a line along zero rather than no line for dates before data exists](https://trello.com/c/r365wSrf/1178-3-user-satisfaction-chart-have-a-line-along-zero-rather-than-no-line-for-dates-before-data-exists)

PR #1190, opened 2019-03-19
Remove some unnecessary gems and initializers

cbaines:


Gemfile
```diff
@@ -5,7 +5,6 @@ ruby File.read('.ruby-version').chomp
 # GOV.UK gems and forks
 gem 'gds-api-adapters', '~> 57.4.2'
 gem 'gds-sso'
-gem 'govuk_admin_template', '~> 6.5'
```
pmanrubia: If I remember correctly these [two lines won't work](https://github.com/alphagov/content-performance-manager/blob/master/app/models/dimensions/date.rb#L9-L10) if this gem is removed. This could be the cause of the failing build.

cbaines: The failing build seems unrelated, something to do with the `Healthchecks::SearchAggregations` tests, which I can reproduce locally on master.

I'll have a look at those lines though, thanks :)

cbaines: I've copied the configuration from the `govuk_admin_template` gem now (in [1]), so this shouldn't be an issue.

1: https://github.com/alphagov/content-performance-manager/pull/1190/commits/d21bb7d2a9fdfbae1a7aded056b0e672f40d1bc9

PR #1197, opened 2019-03-27
Add missing attributes to transform_path helper

chao-xian:
We were not consolidating metrics from Google Analytics such as searches
when addressing duplicate paths from GA reports. This resulted in odd
numbers in our data.

### DIFF
--- a/app/domain/etl/ga/concerns/transform_path.rb
+++ b/app/domain/etl/ga/concerns/transform_path.rb
@@ -22,6 +22,11 @@ private
     if duplicate_event
       attributes[:pviews] = event.pviews + duplicate_event.pviews
       attributes[:upviews] = event.upviews + duplicate_event.upviews
+      attributes[:useful_no] = event.useful_no + duplicate_event.useful_no
+      attributes[:useful_yes] = event.useful_yes + duplicate_event.useful_yes
+      attributes[:searches] = event.searches + duplicate_event.searches
+      attributes[:entrances] = event.entrances + duplicate_event.entrances
+      attributes[:exits] = event.exits + duplicate_event.exits
 
       duplicate_event.destroy
     end
--- a/app/domain/etl/ga/concerns/transform_path.rb
+++ b/app/domain/etl/ga/concerns/transform_path.rb
@@ -20,13 +20,10 @@ private
     attributes = { page_path: sanitised_page_path }
 
     if duplicate_event
-      attributes[:pviews] = event.pviews + duplicate_event.pviews
-      attributes[:upviews] = event.upviews + duplicate_event.upviews
-      attributes[:useful_no] = event.useful_no + duplicate_event.useful_no
-      attributes[:useful_yes] = event.useful_yes + duplicate_event.useful_yes
-      attributes[:searches] = event.searches + duplicate_event.searches
-      attributes[:entrances] = event.entrances + duplicate_event.entrances
-      attributes[:exits] = event.exits + duplicate_event.exits
+      metric_names = Metric.ga_metrics.map(&:name)
+      metric_names.each do |metric|
+        attributes[metric.to_sym] = event.send(metric) + duplicate_event.send(metric)
+      end
 
       duplicate_event.destroy
     end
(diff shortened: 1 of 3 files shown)


chao-xian:
[(5) Bug - Number of searches data is lower than expected for some days for some pages](https://trello.com/c/RuDaeDzX/944-5-bug-number-of-searches-data-is-lower-than-expected-for-some-days-for-some-pages)

spec/domain/etl/ga/concerns/transform_path_spec.rb
```diff
@@ -18,10 +18,38 @@ class Dummy
...
+      create(:ga_event,
+        :with_views,
+        page_path: "/topics",
+        upviews: 100,
+        pviews: 200,
+        useful_no: 300,
+        useful_yes: 400,
+        searches: 500,
+        entrances: 600,
+        exits: 700)
```
cbaines: :+1: I like avoiding long lines :)

app/domain/etl/ga/concerns/transform_path.rb
```diff
@@ -20,13 +20,10 @@ def transform_event_attributes(event)
...
-      attributes[:pviews] = event.pviews + duplicate_event.pviews
-      attributes[:upviews] = event.upviews + duplicate_event.upviews
-      attributes[:useful_no] = event.useful_no + duplicate_event.useful_no
-      attributes[:useful_yes] = event.useful_yes + duplicate_event.useful_yes
-      attributes[:searches] = event.searches + duplicate_event.searches
-      attributes[:entrances] = event.entrances + duplicate_event.entrances
-      attributes[:exits] = event.exits + duplicate_event.exits
+      metric_names = (Metric.custom_metrics + Metric.ga_metrics).map(&:name)
+      metric_names.each do |metric|
+        attributes[metric.to_sym] = event.send(metric) + duplicate_event.send(metric) if event.respond_to?(metric)
```
cbaines: What does it mean is happening if `event.respond_to?(metric)` is false?

spec/domain/etl/ga/concerns/transform_path_spec.rb
```diff
@@ -23,27 +23,61 @@ class Dummy
...
+    it 'returns the correct number of metrics' do
+      event_attributes = event2.attribute_names
+        .reject { |name| name == 'id' }
+        .reject { |name| name == 'bounces' }
+        .reject { |name| name == 'date' }
+        .reject { |name| name == 'page_path' }
+        .reject { |name| name == 'updated_at' }
+        .reject { |name| name == 'created_at' }
+        .reject { |name| name == 'process_name' }
+        .reject { |name| name == 'page_time' }
```
cbaines: Maybe this could be rewritten, something like?
```ruby
names_to_exclude = %w[id bounces date ...]
event_attributes = event2.attribute_names
  .reject { |name| names_to_exclude.include? name }
```

app/domain/etl/ga/concerns/transform_path.rb
chao-xian: I know it's super unlikely, but I'd like to guard against `NoMethodError` here.

spec/domain/etl/ga/concerns/transform_path_spec.rb
chao-xian: Keep forgetting that's the better way to do things like this. Fixed in https://github.com/alphagov/content-performance-manager/pull/1197/commits/5646194a2d6e75947d5f83716199581850d887af - thanks for the feedback.

app/domain/etl/ga/concerns/transform_path.rb
cbaines: Wouldn't we want to know about an `NoMethodError` here, if it occurred? 

At the moment, attributes would just have a `nil` value I assume.

chao-xian: Ah it's actually because one of our 'Custom' metrics (satisfaction) isn't an attribute of Events::GA and I can't think of an elegant way to handle that.

cbaines: So if there's an exception, it would be more readable to be explicit about it.

Maybe put `next if metric == 'satisfaction'` if the block, along with a comment, or maybe do:
```ruby
some_meaningful_variable = %w[satisfaction]
(metric_names - some_meaningful_variable).each do |metric|
 ....
end
```

PR #1206, opened 2019-04-01
ADR to rename Content Performance Manager

pmanrubia:
The data warehouse is no longer an application with an UI. It is an API that is
consumed by other applications in order to get metrics through time for content
items of GOV.UK.


doc/arch/adr-016-rename-content-performance-manager.md
```diff
@@ -0,0 +1,25 @@
...
+
+The data warehouse is no longer an application with an UI. It is an API that is
+consumed by other applications in order to get metrics through time for content
+items of GOV.UK.
+
+The name of the application is not compliant with GOV.UK guidelines for naming.
+
+## Decision
+
+Rename Content Performance Manager to Content Data Admin
```
theseanything:
```suggestion
Rename Content Performance Manager to Content Data API
```

PR #1214, opened 2019-04-08
Redefine `latest` to `live` for editions

theseanything:
This PR is to fix the logic around how we process messages from the PublishingAPI. We have bug whereby content items that are unpublished ('gone' or 'vanish') prevent new content items published with the same base path from being processed.

For example:
- Content item `A` published for base path `/a`
- Content item `A` unpublished for base path `/a` - so edition: `latest: true` and `gone` 
- Content item `B` published for base path `/a` - **this fails to add new edition**

Editions for new content item `B` fail to be added due to the unique `base_path` and `latest` constraint for dimension editions. This is because the base path `/a`  is being used by the unpublished edition from the old content item `A`. Our logic doesn't "demote" the old edition to `latest: false` as it is from a different content item.

This PR aims to redefine `latest`. The attribute previously represented the most current edition of a content item. However, this means the unique constraint of `base_path` and `latest` is incorrect. To fix this `latest` should represent the editions which are currently "live" i.e. using a base_path on GOV.UK. This means `gone` and `vanish` editions are not "live" and allows new content items to use the base path. 

Changes include:
- Renaming the edition attribute `latest` to `live`
- Preventing the promotion of unpublished editions to `live`
- Fix finding previous editions of a content item when processing messages from PublishingAPI
- Add rake task to update existing unpublished editions

### DIFF
--- a/spec/integration/streams/single/publishing_editions_spec.rb
+++ b/spec/integration/streams/single/publishing_editions_spec.rb
@@ -179,7 +179,7 @@ RSpec.describe 'PublishingAPI message queue' do
   end
 
 
-  xit 'publishes redirect edition for content item' do
+  it 'publishes redirect edition for content item' do
     expect(GovukStatsd).to receive(:increment)
      .with('monitor.messages.major').twice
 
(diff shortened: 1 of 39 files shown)


theseanything:
[(5) Bug - Content item not found on index page](https://trello.com/c/D78u5I3O/1176-5-bug-content-item-not-found-on-index-page)

spec/integration/streams/single/publishing_editions_spec.rb
```diff
@@ -0,0 +1,285 @@
...
+    )
+
+    expect_messages_to_have_publishing_api_events([message, gone_message1])
+    expect(message).to be_acked
+    expect(gone_message1).to be_acked
+    expect(gone_message2).to be_acked
+  end
+
+
+  xit 'publishes redirect edition for content item' do
```
peteglondon: Should this spec be skipped?

theseanything: Enabled in newer commit

PR #1219, opened 2019-04-11
Improve ETL process error reporting

cbaines:
# Description

Propagate failures up through the ETL process, so that the Rake tasks exits at the end with code 1 when any part of the process has failed. This means that the Jenkins job will show an appropriate status, and the Icinga passive alert will show the correct status.

---
# Review Checklist
* ~~Changes in scope.~~
* ~~Added to Trello card.~~

### DIFF
--- a/app/domain/etl/feedex/processor.rb
+++ b/app/domain/etl/feedex/processor.rb
@@ -27,11 +27,6 @@ private
       Events::Feedex.import(events, batch_size: BATCH_SIZE)
       batch += 1
     end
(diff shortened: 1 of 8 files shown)


cbaines:
> Can we have tests for catching a failed processor or monitor?

Good point, I've changed the shared examples for the traps and logs stuff to test the return value, and improved the `master_processor_spec.rb` and `metrics_spec.rb` tests to have expectations about the return values.

PR #1231, opened 2019-04-23
Include the Sidekiq retry set size in the healthcheck

cbaines:
It's a good idea to keep an eye on this. I've set the warning
threshold very low as currently the application is running with an
empty retry set (which is great!).

---
# Review Checklist
* [x] Changes in scope.
* ~~Added/updated unit tests.~~
* ~~Added/updated feature tests.~~
* [x] Added to Trello card.

### DIFF
--- a/app/domain/healthchecks/sidekiq_retry_size.rb
+++ b/app/domain/healthchecks/sidekiq_retry_size.rb
@@ -1,0 +1,11 @@
+module Healthchecks
+  class SidekiqRetrySize < GovukHealthcheck::SidekiqRetrySizeCheck
+    def critical_threshold
+      100
+    end
+
+    def warning_threshold
+      1
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


app/domain/healthchecks/sidekiq_retry_size.rb
```diff
@@ -0,0 +1,11 @@
+module Healthchecks
+  class SidekiqRetrySize < GovukHealthcheck::SidekiqRetrySizeCheck
+    def critical_threshold
+      100
+    end
+
+    def warning_threshold
+      1
```
theseanything: Could this potentially be too low and become noise?

cbaines: Yep, I think we'll just have to see how it goes. Having an empty retry set is ideal, and it does seem to be the case at the moment.

PR #1233, opened 2019-04-25
Trap errors in our Sidekiq jobs

peteglondon:
The default behaviour for the Sidekiq jobs is
the write the whole exception including a full stack
trace into the logs.

This is causing problems when we have an error in our code
because the huge log files fill the hard disk, causing problems
for the other apps on the `backend` machines.

This PR traps the errors in our own code and:

* Writes only the message from the exception to the logs
* Sends the entire exception to sentry, where we have set up a Slack
  notification for recurring errors.

Trello: https://trello.com/c/bxmM4ujk/1355-configure-logging-to-exclude-stack-trace-for-sidekiq-jobs#

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.

### DIFF
--- a/app/domain/streams/message_processor_job.rb
+++ b/app/domain/streams/message_processor_job.rb
@@ -14,9 +14,20 @@ module Streams
 
       Monitor::Messages.run(routing_key)
 
-      ActiveRecord::Base.transaction do
-        message.handler.process
+      begin
+        ActiveRecord::Base.transaction do
+          message.handler.process
+        end
+      rescue StandardError => e
+        GovukError.notify(e)
+        logger.error(e.message)
       end
+    end
+
+  private
+
+    def logger
+      Logger.new(STDERR)
     end
   end
 end
--- a/spec/domain/streams/message_processor_job_spec.rb
+++ b/spec/domain/streams/message_processor_job_spec.rb
@@ -5,4 +5,25 @@ RSpec.describe Streams::MessageProcessorJob do
 
     subject.perform(message.payload, message.delivery_info.routing_key)
   end
+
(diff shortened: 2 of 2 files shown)


peteglondon:
[Configure logging to exclude stack trace for sidekiq jobs](https://trello.com/c/bxmM4ujk/1355-configure-logging-to-exclude-stack-trace-for-sidekiq-jobs)

app/domain/streams/message_processor_job.rb
```diff
@@ -14,9 +14,20 @@ def perform(payload, routing_key)
...
 
-      ActiveRecord::Base.transaction do
-        message.handler.process
+      begin
+        ActiveRecord::Base.transaction do
+          message.handler.process
+        end
+      rescue StandardError => e
+        GovukError.notify(e)
+        logger.error(e.message)
```
cbaines: How does this compare to changing the logger that Sidekiq uses? Either just changing the log level, or changing the formatter?

peteglondon: We are trying to get rid of the stack trace

cbaines: I get that, I'm asking about the approach.

One difference I think, is that this approach might not work with Sidekiq retrying? If the exception is being caught, does Sidekiq know the job has failed?

cbaines:
@peteglondon Looks like the move to RabbitMQ only is going ahead, so maybe this can be closed now? https://github.com/alphagov/content-performance-manager/pull/1240

theseanything:
Potential issues with prevent Sidekiq retries - closing for now.

PR #1236, opened 2019-05-01
Fix to prevent failing tests on beginning of month

theseanything:
We have failing test when the current day is beginning of the month and the previous month only had 30 days. This is due to the tests calculating `x` months ago from today instead of yesterday, like the
Postgres aggregation views.

---
# Review Checklist
* [x] Changes in scope.
* [x] Added/updated relevant documentation.
* [x] Added to Trello card.


theseanything:
[Fixed date issue causing failing tests](https://trello.com/c/UT6KQmwR/1416-fixed-date-issue-causing-failing-tests)

PR #1266, opened 2019-06-03
Rename `organisation_id` to `primary_organisation_id`

cbaines:
This is more reflective of what it actually means (the content_id of
the primary publishing organisation), and more consistent with the
other fields (like primary_organisation_title).

As the Content Data app expects organisation_id, these changes are
mostly internal, and the external changes should be compatible with
the Content Data Admin.

Once this has been deployed, and the Content Data Admin switched
across to use the primary_organisation_id, then the intermediate
changes can be removed.

---
# Review Checklist
* [x] Changes in scope.
* [x] Added to Trello card.

### DIFF
--- a/app/domain/finders/content.rb
+++ b/app/domain/finders/content.rb
@@ -1,9 +1,5 @@
 class Finders::Content
   def self.call(filter:)
-    raise ArgumentError unless filter.has_key?(:organisation_id) && filter.has_key?(:date_range)
-
-    filter.assert_valid_keys :search_term, :date_range, :organisation_id, :document_type, :page, :page_size, :sort_attribute, :sort_direction
-
     new(filter).call
   end
 
@@ -23,6 +19,14 @@ private
   NONE = 'none'.freeze
 
   def initialize(filter)
+    raise ArgumentError unless
+      filter.has_key?(:organisation_id) && filter.has_key?(:date_range)
+
+    filter.assert_valid_keys(
+      :search_term, :date_range, :organisation_id, :document_type, :page,
+      :page_size, :sort_attribute, :sort_direction
+    )
+
     @filter = filter
 
     @page = filter[:page] || 1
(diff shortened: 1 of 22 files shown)


theseanything:
So I don't think we should rename query string and change the external API, as we haven't got the need to distinguish the difference in the filters.

cbaines:
> So I don't think we should rename query string and change the external API, as we haven't got the need to distinguish the difference in the filters.

I get that, however, I think there's still a benefit in having the internal code and the external API be consistent. 

Additionally, I'm thinking of adding a `organisation_ids` query string parameter, which I don't want to be confused with `organisation_id`.

theseanything:
> I get that, however, I think there's still a benefit in having the internal code and the external API be consistent.

I'm not sure what these benefits would be for those consuming the API?

> Additionally, I'm thinking of adding a `organisation_ids` query string parameter, which I don't want to be confused with `organisation_id`.

So I think we want to get the API into a state where the `organisation_id` filter matches either `primary_organisation_id` OR `organisations_id`. Currently we have no need to distinguish these to when searching content items - and not sure what the benefits would be when searching.

cbaines:
> > I get that, however, I think there's still a benefit in having the internal code and the external API be consistent.
> 
> I'm not sure what these benefits would be for those consuming the API?

I wasn't thinking of consumers benefiting, just those working on the code and the API.

> > Additionally, I'm thinking of adding a `organisation_ids` query string parameter, which I don't want to be confused with `organisation_id`.
> 
> So I think we want to get the API into a state where the `organisation_id` filter matches either `primary_organisation_id` OR `organisations_ids` database columns. Currently we have no need to distinguish these to when searching content items.

Ah, right. I wasn't thinking at far ahead, but now that I am, yeah, I can't think of any feasible alternatives to represent the query that we at least want to initially support. I'll have a look at adjusting these changes to use `organisation_id` in the API still.

cbaines:
> Ah, right. I wasn't thinking at far ahead, but now that I am, yeah, I can't think of any feasible alternatives to represent the query that we at least want to initially support. I'll have a look at adjusting these changes to use `organisation_id` in the API still.

I've now done this and pushed some new commits.

PR #1269, opened 2019-06-04
Make ga metric imports case insensitive

peteglondon:
# What
Make the GA import processes case insensitive.

# Why

We are missing metrics from GA, this is because all data returned from GA has all lowercase 
`page_path` - data in content data has mixed case paths so they don't match.

Trello: https://trello.com/c/kCcLpI1m/967-3-bug-base-paths-with-uppercase-characters-get-no-metrics-from-google-analytics

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.

### DIFF
--- a/app/domain/etl/ga/internal_search_processor.rb
+++ b/app/domain/etl/ga/internal_search_processor.rb
@@ -50,7 +50,7 @@ private
         SELECT searches,
                dimensions_editions.id
         FROM events_gas, dimensions_editions
-        WHERE page_path = base_path
+        WHERE page_path = LOWER(base_path)
               AND events_gas.date = '#{date_to_s}'
       ) AS s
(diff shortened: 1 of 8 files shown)


peteglondon:
[(3) Bug - Base paths with uppercase characters get no metrics from Google Analytics](https://trello.com/c/kCcLpI1m/967-3-bug-base-paths-with-uppercase-characters-get-no-metrics-from-google-analytics)

cbaines:
Code and tests looks good :+1: 

I think the commit messages could do with having this bit in them:

> We are missing metrics from GA, this is because all data returned from GA has all lowercase page_path - data in content data has mixed case paths so they don't match.

But apart from that, it looks good to me.

PR #1277, opened 2019-06-14
Mark 4 coming soon editions as not live

cbaines:
These editions shouldn't be live, as they should have been replaced by
some actual content. However, due to the special behaviour in the
Publishing API for replacing coming_soon items, the Content Data API
wasn't informed.

The plan to rectify this is to manually mark the affected items as not
live (this migration), and then manually have the Publishing API send
the Content Data API the state of the affected pages.

---
# Review Checklist
* [x] Changes in scope.
* ~~Added/updated unit tests.~~
* ~~Added/updated feature tests.~~
* [ ] Added to Trello card.


db/migrate/20190614142901_mark4_coming_soon_pages_as_not_live.rb
```diff
@@ -0,0 +1,20 @@
+class Mark4ComingSoonPagesAsNotLive < ActiveRecord::Migration[5.2]
+  CONTENT_IDS = %w[
+    33a6001a-8947-477e-93e9-9134587801ae
+    81ebd48d-733c-4d54-8bc3-572abf6d02a5
+    9a8844b3-c4b8-4723-aecb-5d27b5cdfc88
+    9bc3c55b-1d0d-4c7c-92b4-0b56ff560ee3
+  ]
```
theseanything:
```suggestion
  Dimensions::Edition.where(live: true, document_type: 'coming_soon')
```
Instead of hardcoding values?

cbaines: Personally I like having the behaviour be explicit. It avoids the migration changing anything it shouldn't (until [1] is merged, it'll affect a lot more records than 4), and makes it clear what it should have done when you look back at it.

1: https://github.com/alphagov/whitehall/pull/4851

Do you think there's a benefit to using a query?

theseanything: I feel that hardcoding values into the code is introducing state. Thinking about it, I feel this should be a rake task instead of a migration because:
- This isn't modifying the db schema, only the data - so would be missed by `rake db:schema:load` or `rake db:reset`
- As rake task, we can run it when and as needed.

theseanything:
See: https://github.com/alphagov/content-data-api/pull/1279

PR #1313, opened 2019-07-22
Make etl more resilient

peteglondon:
# Description
Attempt to tidy up the Etl:Master process - includes:

* Allow Etl:Master to be re-run without raising an error.
* Rename `find_or_create` to `for_date` for `Dimensions:Date` and `Dimensions::Month`

Trello: https://trello.com/c/TYYrZ08d/653-content-data-api-etl-extract-transform-load

### DIFF
--- a/app/models/dimensions/date.rb
+++ b/app/models/dimensions/date.rb
@@ -30,12 +30,8 @@ class Dimensions::Date < ApplicationRecord
     date_dimension
   end
 
-  def self.find_or_create(date)
-    begin
-      find_by(date: date) || create_with(date)
-    rescue ActiveRecord::RecordNotUnique
-      find_by(date: date)
-    end
+  def self.for_date(date)
+    find_by(date: date) || create_with(date)
   end
 
   def self.exists?(date)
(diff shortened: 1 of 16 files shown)


app/models/dimensions/date.rb
```diff
@@ -30,12 +30,8 @@ def self.create_with(date)
     date_dimension
   end
 
-  def self.find_or_create(date)
-    begin
-      find_by(date: date) || create_with(date)
-    rescue ActiveRecord::RecordNotUnique
-      find_by(date: date)
-    end
+  def self.for_date(date)
```
theseanything: Again I think the method name here isn't descriptive enough.

app/models/dimensions/date.rb
```diff
@@ -30,12 +30,8 @@ def self.create_with(date)
     date_dimension
   end
 
-  def self.find_or_create(date)
-    begin
-      find_by(date: date) || create_with(date)
-    rescue ActiveRecord::RecordNotUnique
```
theseanything: We still have race conditions where concurrent executions both "find" at the same time, and then attempt to create.

PR #1442, opened 2020-03-10
Allow content data to be returned by month

edwardkerry:
This PR extends the content-data-api to receive a date_range param with a specified month and year, in addition to the preset date ranges.

In order to retrieve the correct data it also updates the monthly aggregations logic to present the "satisfaction" data as a ratio rather than a sum, in line with logic elsewhere in the api.

This change is for a specific use-case and is not being presented in the content-data-admin UI.

[dependent content-data-admin PR](https://github.com/alphagov/content-data-admin/pull/753)
[trello](https://trello.com/c/OKYeyYHe/1799-5-allow-content-data-api-and-content-data-admin-to-return-data-by-month)

### DIFF
--- a/app/domain/finders/content.rb
+++ b/app/domain/finders/content.rb
@@ -1,4 +1,6 @@
 class Finders::Content
+  include SpecificMonths
+
   def self.call(filter:)
     new(filter).call
   end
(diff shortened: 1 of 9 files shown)


app/domain/finders/content.rb
```diff
@@ -42,13 +44,25 @@ def view
...
+    scope = find_by_month(scope) if SpecificMonths::VALID_SPECIFIC_MONTHS.include?(@date_range)
     scope = find_by_search_term(scope) if @filter[:search_term].present?
     scope = find_by_document_type(scope) if @filter[:document_type].present?
     scope = find_by_organisation(scope)
 
     scope
   end
 
+  def find_by_month(scope)
+    month_and_year = @date_range.split("-")
```
theseanything: might be clearer and more concise to assign individual variables straight away
```suggestion
    month, year = @date_range.split("-")
```

app/domain/finders/content.rb
```diff
@@ -42,13 +44,25 @@ def view
...
+    month_num = Date::MONTHNAMES.index(month_and_year[0].capitalize)
+    year = month_and_year[1]
+    month_id = "#{year}-#{sprintf '%02d', month_num}"
```
theseanything:
```suggestion
    month_id = "%s-%02d" % [year, month_num]"
```

edwardkerry: 👍

PR #1509, opened 2020-07-14
Update Rails to 6.0.3.2

pudinha:


config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,30 @@
+# Be sure to restart your server when you modify this file.
```
cdccollins, line 1: I think govuk has a way of managing their security policies separately, so this file isn't necessary. See https://github.com/alphagov/government-frontend/pull/1811#discussion_r450932765

benbaumann95:
Just FYI for commit 4b13ecd take a look at the commit https://github.com/alphagov/service-manual-frontend/pull/749/commits/b85a150f4245367439d397bde104784d0e3e9448
I've used this message when upgrading the remaining dependencies

config/initializers/content_security_policy.rb
pudinha: :sparkles:

EGiataganas:
Hey @pudinha  this looks great, I've compared the pr changes with [railsdiff][1] and there are some updates that could also be done

1. Remove `bin/bundle`, `bin/update`
2. In the `config/spring.rb` the `%w[]` needs to be updated to `Spring.watch()`
```ruby
Spring.watch(
  ".ruby-version",
  ".rbenv-vars",
  "tmp/restart.txt",
  "tmp/caching-dev.txt"
)
```

[1]: http://railsdiff.org/5.2.4.3/6.0.3.2

pudinha:
@thomasleese one test is failing because of https://github.com/alphagov/govuk_schemas/pull/60 . Should merge that fix and release govuk_schemas 4.1.1 before merging content-data-api? Or ignore that test for now? What do you think?

thomasleese:
@pudinha I think we should merge and release https://github.com/alphagov/govuk_schemas/pull/60 first, thanks for checking!

PR #1557, opened 2020-12-08
Add Healthchecks for ActiveRecord and SidekiqRedis

edwardkerry:
These two healthchecks are prerequisites for enabling ContinousDeployment
for this app.

Co-authored-by: Rebecca Pearce <rebecca.pearce@digital.cabinet-office.gov.uk>


thomasleese (commented):
I reckon we can get rid of this custom check now: https://github.com/alphagov/content-data-api/blob/master/app/domain/healthchecks/database_connection.rb

edwardkerry:
> I reckon we can get rid of this custom check now: https://github.com/alphagov/content-data-api/blob/master/app/domain/healthchecks/database_connection.rb

good idea, thank you!

thomasleese (approved):

PR #1582, opened 2021-03-08
Rename ETL master => main

MuriloDalRi:
https://trello.com/c/Ts2kFTYc/216-rename-etl-master-task-to-etl-main
https://trello.com/c/1DGoJ3oZ/630-rename-etl-master-task-to-etl-main-task

### DIFF
--- a/lib/tasks/etl.rake
+++ b/lib/tasks/etl.rake
@@ -1,10 +1,13 @@
 namespace :etl do
-  desc "Run ETL master process for yesterday"
-  task master: :environment do
-    unless Etl::Master::MasterProcessor.process
-      abort("Etl::Master::MasterProcessor failed")
+  desc "Run ETL main process for yesterday"
+  task main: :environment do
+    unless Etl::Main::MainProcessor.process
+      abort("Etl::Main::MainProcessor failed")
     end
   end
+
+  desc "delete after main task is deployed"
+  task master: :main # delete after main task is deployed
 
   desc "Run Etl::Aggregations::Monthly for range of dates"
   task :repopulate_aggregations_month, %i[from to] => [:environment] do |_t, args|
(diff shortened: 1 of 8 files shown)


lib/tasks/etl.rake
```diff
@@ -1,8 +1,8 @@
 namespace :etl do
-  desc "Run ETL master process for yesterday"
-  task master: :environment do
```
thomasleese, line 3: I think we should keep the old `master` task around until the Puppet PR has been deployed so we don't end up with a period of time where the Jenkins job doesn't work. I think we can do something like this `task :master => :main`.

MuriloDalRi: I've made this change. Does it look right to you?

thomasleese: I think it'll need to be two tasks, so something like this:

```rb
desc "Run ETL main process for yesterday"
task main: :environment do
  unless Etl::Master::MasterProcessor.process
    abort("Etl::Master::MasterProcessor failed")
  end
end

task master: :main  # master depends on main, but does nothing itself
```

MuriloDalRi: Oh right, yeah! I'll change that

thomasleese (approved):
Code looks fine, looks like there are some unrelated security fixes... if you want to have a go at fixing those that would be great, or we've got https://trello.com/c/6OOK52R2/2149-fix-content-data-api-brakeman-errors.

MuriloDalRi:
> Code looks fine, looks like there are some unrelated security fixes... if you want to have a go at fixing those that would be great, or we've got https://trello.com/c/6OOK52R2/2149-fix-content-data-api-brakeman-errors.

Thanks, I was going to add a card about that! I think it should be looked at separately as it's not blocking deployment for this PR though.

PR #1590, opened 2021-03-25
Add govuk_test

thomasleese:
This is a library that we should be using in all our applications to provide a consistent test environment. In this particular case, it brings in a version of Brakeman that won't be automatically updated to the latest version and thus avoids problems of PRs failing due to an external factor.

We should eventually upgrade to a more recent version of Brakeman, however this should be done in coordination with the govuk_test Gem so it's something we do across all our apps at the same time.

See https://github.com/alphagov/govuk_test/pull/28 for more background on why we're doing this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,6 +38,7 @@ group :development, :test do
   gem "database_cleaner"
   gem "factory_bot_rails"
   gem "govuk_schemas"
+  gem "govuk_test"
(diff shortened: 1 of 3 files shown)


brucebolt (approved):

ChrisBAshton (approved):
Good spot 👍

PR #1595, opened 2021-03-29
Add search and metrics healthcheck endpoints

brucebolt:
This creates a separate endpoint for the search and metrics healthchecks.

At the moment, a failure in any of these will report the application instance as being `critical` in Icinga, once per node.  This should be measured on per application (rather than per node) basis.

In a later commit, we will remove these metrics from the main application healthcheck. This means we can have a separate alert in Icinga for the search (once, rather than per instance) rather than marking each instance as critical.

Trello card: https://trello.com/c/X2AKAUKn and https://trello.com/c/aDoOuYf5

### DIFF
--- a/spec/requests/healthcheck_metrics_spec.rb
+++ b/spec/requests/healthcheck_metrics_spec.rb
@@ -1,0 +1,19 @@
+RSpec.describe "/healthcheck/metrics" do
+  before do
+    get "/healthcheck/metrics"
+  end
+
+  it "returns distinct organisations ordered by title" do
+    json = JSON.parse(response.body)
+
+    expect(json["checks"]).to include("daily_metrics")
+    .and(include("etl_metric_values_pviews"))
+    .and(include("etl_metric_values_searches"))
+    .and(include("etl_metric_values_upviews"))
+    .and(include("etl_metric_values_feedex"))
+  end
+
+  it "is not cacheable" do
+    expect(response.headers["Cache-Control"]).to eq "max-age=0, private, must-revalidate"
+  end
+end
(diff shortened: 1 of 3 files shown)


spec/requests/healthcheck_metrics_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    expect(json["checks"]).to include("daily_metrics")
+    .and(include("etl_metric_values_pviews"))
+    .and(include("etl_metric_values_searches"))
+    .and(include("etl_metric_values_upviews"))
+    .and(include("etl_metric_values_feedex"))
+  end
+
+  it "is not cacheable" do
+    get "/healthcheck"
```
thomasleese, line 14:
```suggestion
    get "/healthcheck/metrics"
```

But I wonder if it's worth having a `before { get "/healthcheck/metrics" }` in the `describe` block? It might need refactoring later if we want a custom request, but for these tests it should work.

spec/requests/healthcheck_search_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    expect(json["checks"]).to include("aggregations")
+      .and(include("search_last_month"))
+      .and(include("search_last_six_months"))
+      .and(include("search_last_thirty_days"))
+      .and(include("search_last_three_months"))
+      .and(include("search_last_twelve_months"))
+  end
+
+  it "is not cacheable" do
+    get "/healthcheck"
```
thomasleese, line 15:
```suggestion
    get "/healthcheck/search"
```

And similar to my comment above about a `before` block.

spec/requests/healthcheck_metrics_spec.rb
brucebolt: Good idea.  I've moved these into a `before`.

thomasleese (approved):

PR #1644, opened 2021-09-09
Bugfix: limit path lengths to 1500

bilbof:
Until now we have limited paths to 1500 chars only if they include a query string. However, some paths are very long, so we need to reject those paths (regardless of whether they have a query string) due to a DB limit: index row size maximum is 2712 chars.

We saw some [PG::ProgramLimitExceeded](https://sentry.io/organizations/govuk/issues/2636165771) errors which has motivated this change.

---
# Review Checklist
* [ ] Changes in scope.
* [ ] Added to Trello card.


ChrisBAshton (approved):
🎉

PR #1658, opened 2021-12-07
Update aggregation spec floating point precision assertion

edwardkerry:
This commit replaces a direct assertion on a 15 decimal place
floating point number as the extreme precision can cause tests to fail.

Instead we are now asserting against a (still very precise) range.


hannako (approved):
👍

PR #1659, opened 2021-12-07
Use Postgres 13 in CI

edwardkerry:
Uses Postgres13 for CI by setting the TEST_DATABASE_URL environment variable.

[find and view trello](https://trello.com/c/zhH82ATS/955-upgrade-content-data-api-from-postgres-96-to-postgres-13)
[database upgrade trello](https://trello.com/c/hHuLcRMq/3-content-data-api)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,9 +2,12 @@
 
 library("govuk")
 
-node("postgresql-9.6") {
+node() {
   govuk.buildProject(
     beforeTest: {
+      // Run against the Postgres 13 Docker instance on GOV.UK CI
+      govuk.setEnvar("TEST_DATABASE_URL", "postgresql://postgres@127.0.0.1:54313/content-data-api-test")
+
       govuk.setEnvar("TEST_COVERAGE", "true")
     }
   )


Jenkinsfile
```diff
@@ -5,6 +5,9 @@ library("govuk")
 node("postgresql-9.6") {
```
1pretz1, line 5: Should this node (name?) be updated too?

edwardkerry: i'm just checking that...I wonder if we can remove the argument altogether as i've not seen it in other apps...

1pretz1: Yeah, maybe. I've just realised it's also present in Content Data Admin (https://github.com/alphagov/content-data-admin/blob/main/Jenkinsfile#L5).

1pretz1: I've removed it from my PR (https://github.com/alphagov/content-data-admin/pull/993) - lets see if the build passes.

1pretz1: Build passed - https://ci.integration.publishing.service.gov.uk/job/content-data-admin/job/upgrade-to-postgres-13/3/

edwardkerry: I've found renaming it fails - the name is not arbitrary. Removing it entirely passes so I assume that is the best route to take at the moment

1pretz1 (approved):
Looks good to me!

PR #1691, opened 2022-03-15
Upgrade rails 7

catalinailie:
Trello card: https://trello.com/c/Fj7DmBh0/2851-upgrade-our-applications-to-rails-7-5


BhupsAL1 (commented):
LGTM

config/environments/test.rb
```diff
@@ -45,8 +45,6 @@
   # Raises error for missing translations.
   # config.i18n.raise_on_missing_translations = true
 
-  config.active_job.queue_adapter = :inline
```
ChrisBAshton, line 48: Looks like this was set deliberately in b7ae780e58dbcd4fcd5ca56b00c790c155380c1f - we should have a think about whether to leave this alone.

Gemfile.lock
```diff
@@ -442,38 +465,39 @@ GEM
...
+    simplecov_json_formatter (0.1.4)
+    sorted_set (1.0.3)
+      rbtree
+      set (~> 1.0)
+    spring (2.1.1)
     spring-commands-rspec (1.0.4)
       spring (>= 0.9.1)
     spring-watcher-listen (2.0.1)
       listen (>= 2.7, < 4.0)
       spring (>= 1.2, < 3.0)
```
ChrisBAshton, line 487: As per https://github.com/alphagov/support-api/pull/628#discussion_r832031153, we need to set this to at least 3.0.0, and keep `cache_classes` as `false`.

ChrisBAshton (commented):
Thanks @catalinailie - this looks close. A couple of comments below 👀

PR #1694, opened 2022-03-25
Remove cookie rotation following Rails 7 upgrade

catalinailie:
A cookie rotation was used to upgrade encrypted cookies from SHA1 to SHA256 digests to avoid active sessions being lost during the Rails 7 upgrade.

It's now safe to remove this, on the assumption that active sessions have now been upgraded and inactive sessions would have expired by now anyway.


sihugh (approved):
🍪

PR #1753, opened 2022-09-06
Patch time-based spec

MuriloDalRi:
This seems to fail when it's over a year old. Perhaps that's a feature of
content-data?

It's blocking a release of govuk-content-schemas. I intend to return here
when I have more time to fix properly.


ChrisBAshton (approved):
Weird 🤔 but seems a harmless change if it fixes the tests.

PR #1813, opened 2023-01-10
[WIP] Add workflow that combines Dependabot PRs

MuriloDalRi:
This workflow will run daily and combine Dependabot PRs if the tests pass and if they're not major updates.

https://trello.com/c/haWz0i9M/3064-explore-consolidating-dependabot-prs-into-one-pr-5

### DIFF
--- a/.github/workflows/combine-dependabot-prs.yml
+++ b/.github/workflows/combine-dependabot-prs.yml
@@ -1,0 +1,16 @@
+name: 'Combine Dependabot PRs'
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  combine-prs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: alphagov/github-action-combine-dependabot-prs/@v1
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}


.github/workflows/combine-dependabot-prs.yml
```diff
@@ -0,0 +1,218 @@
...
+
+jobs:
+  combine-prs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v6
+        id: create-combined-pr
+        name: Create Combined PR
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
```
robinjam, line 18: Might be a dumb question, but where's this secret defined? (And why do we need a custom token instead of the default generated one?)

robinjam (approved):
Nice one Murilo! Looking forward to seeing it in action.

Assuming it works, is the intention to split out the JS code from this action into a separate repo? (Would allow us to add some unit tests)

ChrisBAshton (commented):
Promising work, @MuriloDalRi 👏 

+1 to James' suggestion though - in the absence of an existing action we can utilise, I think we'll need to create our own action repo (`alphagov/github-action-combine-prs` or something). Then:

1) Test 'calling' that external action from content-data-api
2) Split the JS out into a separate file in that repo, and repeat step 1)
3) Rewrite the JS to be less procedural, and to add unit tests, and repeat step 1)

Ultimately it would be good to get to a point where we can reduce `combine-dependabot-prs.yml` here to just a few lines, e.g.:

```yaml
name: 'Combine Dependabot PRs'
on:
  schedule:
  - cron: "0 0 * * *"

permissions:
  contents: write
  pull-requests: write

jobs:
  combine-prs:
    runs-on: ubuntu-latest
    steps:
      - uses: alphagov/github-action-combine-prs@v1
        with:
          author: 'dependabot[bot]'
```

.github/workflows/combine-dependabot-prs.yml
MuriloDalRi: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret

robinjam: Oh right, it _is_ the default one

You can remove this line then, the `github-script` action uses it by default

PR #1819, opened 2023-01-17
Remove pointless test

brucebolt:
This test checked that `GovukSchemas::Schema.schema_names` included all the known schema names.

When adding a new schema to Content Schemas (in the `publishing-api` repo), this meant we had to add it to the list in this test without making any changes to the application code.

This repo is not an appropriate place for that test, as it solely checks the behaviour of a gem which has its own test for the method.

Removing the test will mean `content-data-api` won't need to be updated each time we add a new schema type.

[Trello card](https://trello.com/c/xUout0dR)

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.


BeckaL (approved):
Good shout, this test always seemed off to me

PR #1828, opened 2023-01-24
Revert "Bump govuk_sidekiq from 5.0.0 to 6.0.0"

Rosa-Fox:
Reverts alphagov/content-data-api#1810

Due to warnings in Icinga:

```
No processes found for content-data-api-bulk-import-publishing-api-consumer-procfile-worker
No processes found for content-data-api-publishing-api-consumer-procfile-worker
```

It is suspected that bumping Sidekiq has caused these warnings. Temporarily downgrading whilst we figure out the issue.


gclssvglx (approved):
LGTM

PR #1836, opened 2023-02-03
Ignore Ruby 2.7.6 EOL Brakeman warning

theseanything:
It's not yet at EOL and work is currently on-going to upgrade the Ruby version.


ChrisBAshton (approved):
Approving, will revert next week when we upgrade the ruby version 👍

PR #1844, opened 2023-02-25
Clean up redundant repo ref in RSpec action.

sengi:
The checkout action checks out the current repo by default, as one would expect. Omitting the repo name therefore slightly helps readability and consistency between similar jobs across our repos.

This is a bulk change consisting of 20 PRs.


martynipratt (approved):
LGTM

PR #1877, opened 2023-05-10
Fix for Travel Advice pages with no summary

robinjam:
A `summary` is no longer required for Travel Advice pages[^1], but if it's not present we currently crash with ```undefined method `find' for nil:NilClass (NoMethodError)```.

[^1]: https://github.com/alphagov/publishing-api/commit/21c0711c0a894182ffcd1ed556ac2194e7365c2b

### DIFF
--- a/app/domain/streams/messages/multipart_message.rb
+++ b/app/domain/streams/messages/multipart_message.rb
@@ -36,11 +36,14 @@ module Streams
     def parts
       message_parts = @payload.dig("details", "parts").dup
       if doc_type == "travel_advice"
-        message_parts.prepend(
-          "slug" => base_path,
-          "title" => "Summary",
-          "body" => [@payload.dig("details", "summary").find { |x| x["content_type"] == "text/html" }],
-        )
+        summary = @payload.dig("details", "summary")&.find { |x| x["content_type"] == "text/html" }
+        unless summary.nil?
+          message_parts.prepend(
+            "slug" => base_path,
+            "title" => "Summary",
+            "body" => [summary],
+          )
+        end
       end
       message_parts
     end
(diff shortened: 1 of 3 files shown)


ChrisBAshton (approved):
Looking at the commit message of the [change in publishing api](https://github.com/alphagov/publishing-api/commit/21c0711c0a894182ffcd1ed556ac2194e7365c2b):

> This is part of a piece of work that will enable FCDO to rename the ‘Summary’ field (or otherwise) for the Travel Advice pages. The plan is to move summary into the parts section and ensure the government-frontend renders either summary if provided or the first part.

We could carry this over to Content Data by using `result["details"]["parts"][0]` if summary is omitted?
But I don’t know enough about where the summary is used in Content Data to know if that’s a good idea.

Your fix looks fine to me though.

sihugh (approved):
This looks good as a fix 👍
As Chris mentioned, we don't know enough about the upcoming changes to make what might perhaps be the _best_ change. Maybe it's worth advertising this fix in the channel of the team working on that stuff, so they might consider the downstream effects?

Also, it's a bit annoying that the "Test Content Schemas" CI step in their PR didn't pick this up in the [PR changing the summary to optional](https://github.com/alphagov/publishing-api/pull/2342). I wonder if there's something there to carry forward 🤔

robinjam:
Had a quick chat with @rylangooch about this, this fix seems OK and moreover it needs to go live in advance of 10am tomorrow, when the summaries are going to be removed from every travel advice page on production.

The `summary` section might be removed entirely in the future, which _shouldn't_ cause another breakage, but would allow us to remove this special case from our code entirely.

PR #1894, opened 2023-06-01
Fixes for govuk-docker

MuriloDalRi:
Adds sass-rails gem and removes Sentry exception that isn't needed anymore.

https://trello.com/c/915tHRoo/3206-fix-content-data-api-in-govuk-docker

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,7 @@ gem "pg"
 gem "rack-proxy"
 gem "rails-erd"
 gem "ruby-progressbar"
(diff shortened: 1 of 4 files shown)


robinjam:
Any idea where the changes to `schema.rb` came from and what they mean? 🤔
(I'd be inclined to just roll them back)

robinjam (approved):
Approving so you can merge, but I'd roll back the changes to `schema.rb` first, personally

PR #2027, opened 2024-02-09
Upgrade ruby to 3.3

MuriloDalRi:
The #govuk-platform-security-reliability-team expects teams to review, test and deploy this following the steps in the [docs](https://docs.publishing.service.gov.uk/manual/ruby.html#updating-individual-applications)


deborahchua (approved):
I've tested this by running some of the ETL rake tasks, and deployed to Integration where the scheduled task ran with no issues. I've also tested through Content Data's UI and all seems well.

PR #2042, opened 2024-03-01
Delete unused API docs

MuriloDalRi:
These have never been published or updated and I don't believe anyone even knows about them.


robinjam:
I think(?) these docs are live at https://content-api.publishing.service.gov.uk but it looks like the GitHub Pages integration was switched off at some point so they're not up to date with this repo.

To be fair though, the API itself doesn't look like it's changed since May 2020, so the API docs are probably still current 😅

robinjam (approved):

PR #2125, opened 2024-07-16
Update dependabot.yml

MahmudH:
Dependabot is looking for a package.json file. This PR removes the check.

Alert from: https://github.com/alphagov/content-data-api/security/dependabot/99

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: /
     schedule:
       interval: daily
+  - package-ecosystem: npm
+    directory: /doc/api
+    schedule: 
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,3 +4,7 @@ updates:
     directory: /
     schedule:
       interval: daily
+  - package-ecosystem: npm
+    directory: /doc/api
+    schedule: 
+      internal: none
```
MuriloDalRi, line 10: Was this meant to be `interval`?

MuriloDalRi:
@MahmudH I don't think this PR is needed anymore

PR #2243, opened 2025-03-03
Add `link_collection` schema to `NoContent` parser

brucebolt:
A schema for `link_collection` was added in https://github.com/alphagov/publishing-api/commit/e3cbefbb6a5adf60be555912f67431f207dad112 but a parser was not defined in this application, resulting in subsequent CI builds failing.

Adding this to the `NoContent` schema as this not a page containing any content, but rather something that is included in other pages to show a list of links.

[Trello card](https://trello.com/c/4BB49LSt)

### DIFF
--- a/app/domain/etl/edition/content/parsers/no_content.rb
+++ b/app/domain/etl/edition/content/parsers/no_content.rb
@@ -22,6 +22,7 @@ class Etl::Edition::Content::Parsers::NoContent
       how_government_works
       knowledge_alpha
       landing_page
+      link_collection
       ministers_index
       organisations_homepage
       person
--- a/spec/domain/etl/edition/content/no_content_spec.rb
+++ b/spec/domain/etl/edition/content/no_content_spec.rb
@@ -24,6 +24,7 @@ RSpec.describe Etl::Edition::Content::Parser do
       how_government_works
       knowledge_alpha
       landing_page
+      link_collection
       ministers_index
       organisations_homepage
       person


spec/domain/etl/edition/content/link_collection_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  it "returns nil if json does not have 'link_items' key" do
+    json = { schema_name: "link_collection",
+             details: {} }
+    expect(subject.extract_content(json.deep_stringify_keys)).to eq(nil)
+  end
+
+  it "returns content json if schema_name is 'link_collection'" do
+    json = { schema_name: "link_collection",
+             details: { link_items: [
+               { title: "Check your Council Tax band" },
```
yndajas, line 13: **question:** should there be a URL property too, or do these tests not need to adhere too strictly to the required fields in the content schema?

brucebolt: Looking at the other parsers, we only seem to deal with the title (or other non-URL) data for the link.

I'm not really sure why this data is needed or where it gets rendered to users in the UI.

brucebolt: Thinking about this more, I think `link_collection` would probably be better off in the `NoContent` parser as it's not a rendered content item with a URL, as far as I can see.

yndajas (approved):
Seems sensible to me. The commit message mentions "`NoContent` schema" - maybe it should be "parser" instead?

PR #2244, opened 2025-03-03
Fix flakey test

brucebolt:
The `./spec/integration/streams/all_schemas_spec.rb` tests are flakey as the payload sent to the message queue does not always contain a base path.

This means the test passes on occasion when it shouldn't,  because the application has been designed to silently ignore invalid payloads.

Explicitly adding a base path for these tests, to remove the flakiness.

Depends on https://github.com/alphagov/content-data-api/pull/2243 (to fix a failing test which passes ~25% of the time before this change).

[Trello card](https://trello.com/c/4BB49LSt)

### DIFF
--- a/spec/support/schemas_iterator.rb
+++ b/spec/support/schemas_iterator.rb
@@ -7,7 +7,9 @@ module SchemasIterator
   end
 
   def self.payload_for(schema_name, schema)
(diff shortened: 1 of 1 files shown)


brucebolt:
Closing as the original behaviour is correct: the schema has an optional base path defined.

PR #2492, opened 2026-04-10
Update rails from 7.2.2.2 to 8

graycodes:
# Description

Dependabot opened a PR here https://github.com/alphagov/content-data-api/pull/2480 to update Rails from 7.2.2.2 to 8.1.3. The guidance in the upgrade-guide suggests only migrating one minor version at a time so this is the first part of the migration, simply moving from 7.2 to 8.0. We can follow this up with another one of these to move from 8.0 to 8.1.

Most config changes are just accepting the suggestions made by Rails. The only tweaks I made here were:

- Ignoring the changes to the Puma config as that's handled by
`GovukPuma`
- Ignoring the changes to ssl as these have been known to cause issues
in other apps
- Ignoring newly generated public asset files as these are discouraged
in the govuk rails upgrade guidance
- A couple of config changes for config.hosts and factorybot that
rails wanted to remove by default, but have been intentionally added
by the maintainers of this repo.

# Links

https://rubylearning.com/blog/2026/04/04/rails-8-upgrade-guide/
https://guides.rubyonrails.org/upgrading_ruby_on_rails.html
https://docs.publishing.service.gov.uk/manual/how-to-upgrade-rails.html

Many decisions influenced by https://github.com/alphagov/authenticating-proxy/commit/13bc57c266d4f296d7e585e609f7fd9c94076664#diff-d2aa36f1b3f7d7185a9513eb6d4e9413bbb5df406e865f19ead21e32a423308b


graycodes:
I've deployed this to integration and it seems to be behaving the same as before, as far as I can tell.

config/initializers/new_framework_defaults_8_0.rb
```diff
@@ -0,0 +1,30 @@
+# Be sure to restart your server when you modify this file.
```
pezholio: We probably don't need this file as its all comments now

db/schema.rb
```diff
@@ -261,53 +260,7 @@
   add_index "aggregations_search_last_months", ["upviews", "warehouse_item_id"], name: "search_last_month_upviews", order: { upviews: "DESC NULLS LAST", warehouse_item_id: :desc }
   add_index "aggregations_search_last_months", ["warehouse_item_id"], name: "aggregations_search_last_months_pk", unique: true
 
-  create_view "aggregations_search_last_thirty_days", materialized: true, sql_definition: <<-SQL
```
pezholio, line 264: Frustrating how the order of these views has been tweaked. Is there a way we can ensure the order is maintained? It's a bit difficult to see what (if anything) has changed, and I'm worried we might get some gnarly results if running `db:schema:dump` has done anything unexpected (I'm sure it hasn't, but better to be safe than sorry!)

pezholio (commented):
Just on minor thing, but I am a little worried about shipping this without the diff of the schema being a bit clearer. I've not dealt with database views in Rails before, so I'm a little nervous about unintended consequences!

db/schema.rb
pezholio: Looks like this was sorted in Scenic `1.9.0`, but our views are still in the "wrong" order - https://github.com/alphagov/content-data-api/pull/2342

graycodes: Right I've done as you suggested and split this into 2 commits where the first keeps the original ordering and just has the code changes, and the second just has the re-ordering with no code changes.

I've taken a look and the only real code changes are a different-but-also-correct way of expressing `interval`s so I think it's all good (but as you say, that was hard to know before!)

pezholio: Amazing! Thanks!

pezholio (approved):
LGTM 👍

PR #2502, opened 2026-04-21
Improve docs for 'Reimport' support task

edavey:
This adds guidance:

- that the task is likely to be run in the `production` env
- on the expected timescale of the task
- on how to monitor progess on the task and what to expect in the output


pezholio (approved):
Nice additions 👍

PR #2546, opened 2026-07-22
CR-53 MHCLG missing from list of organisation seen in admin UI

edavey:
In this PR we fix two bugs which have caused organisations including MHCLG (Ministry of Housing, Communities & Local Government) to be missing from the Content Data Admin interface. 

See [CR-53](https://gov-uk.atlassian.net/browse/CR-53) which was created from [Zendesk ticket 6611295](https://govuk.zendesk.com/agent/tickets/6611295 "https://govuk.zendesk.com/agent/tickets/6611295"). That ticket describes a long standing issue:

> Hi
> 
> The Ministry of Housing, Communities and Local Government (MHCLG) is missing from the Organisation dropdown list. Can you add this in?
> 
> It currently only lists the old organisation: MHCLG (2018 to 2021).
> 
> We have raised a couple of other times, but both have been marked solved and we haven't seen added as a new org.
> 
> PREVIOUS TICKETS:
> 
> [MHCLG missing from Content Data - Govt Agency General Issue – GOV.UK](https://govuk.zendesk.com/hc/en-us/requests/5953490 "https://govuk.zendesk.com/hc/en-us/requests/5953490")
> 
> [https://govuk.zendesk.com/hc/requests/5953321](https://govuk.zendesk.com/hc/requests/5953321 "https://govuk.zendesk.com/hc/requests/5953321")
> 
> Can you review and add us please?

[Issue 2169 ](https://github.com/alphagov/content-data-api/issues/2169) also describes the MHCLG problem and notes that it
> is because Content Data considers it's document type as "substitute" 

## Summary

At present there are 2 editions of the MHCLG competing to be "live":

1. the **old** with `content_id` of `33479fe6-d274-4ab6-bf37-053ebd7fd162`
2. the **new** with `content_id` of `1390d07f-12b3-4f55-9dd5-fec5fd9e3649`

Both wish to use the same `base_path` of `/government/organisations/ministry-of-housing-communities-and-local-government`

The new record can't get successfully "promoted" to the live state as it's unable to demote the incumbent.

## Bugs

On investigation this turns out to be a long standing problem resulting from two problems fixed here:

### 1. "substitute" was not recognised as unpublished

An "unpublished" event of type "substitute" should never be made live. In Content Data API the definition of "unpublished" did not include "substitute". This led to the "old" edition of MHCLG being set as "live".

### 2. demotion did not include editions with matching base_path

When attempting to "promote" a new edition, we were "demoting" any live edition with the same `content_id`. However, we were not also demoting editions with a different `content_id` but the same `base_path`. We have unique indexes on both:

- `content_id` + `live == true` and 
- `base_path` + `live == true`

Since July 2024 we've had 3 editions for the "old" MHCLG record (with `content_id`  of `33479fe6-d274-4ab6-bf37-053ebd7fd162`):

| #   | ID       | created_at (BST)    | document_type | live     | event_id |
| --- | -------- | ------------------- | ------------- | -------- | -------- |
| 1   | 19553926 | 2024-07-09 14:04:03 | organisation  | false    | 4496248  |
| 2   | 19554005 | 2024-07-09 14:09:51 | gone          | false    | 4496324  |
| 3   | 19578717 | 2024-07-11 09:53:50 | substitute    | **true** | 4520698  |

Every attempt since to promote the `content_id` / `warehouse_id` for the new MHCLG record (`content_id`: `1390d07f-12b3-4f55-9dd5-fec5fd9e3649`) has failed due to this db constraint. 

We can see numerous recent events in Publishing API which will have emitted messages of this type:

```rb
Event.where(
  content_id: "1390d07f-12b3-4f55-9dd5-fec5fd9e3649", 
  action: ["Publish", "PatchLinkSet"]
  )
  .order(created_at: :desc)
  .limit(25)
  .pluck(:action, :created_at)
=>
[["Publish", 2026-07-16 12:57:17.214978000 UTC +00:00],
 ["PatchLinkSet", 2026-07-16 12:57:17.151597000 UTC +00:00],
 ["PatchLinkSet", 2026-07-16 12:57:17.000083000 UTC +00:00],
 ["PatchLinkSet", 2026-07-16 12:57:16.978320000 UTC +00:00],
 ["Publish", 2026-07-16 12:50:04.095716000 UTC +00:00],
 ...
 ["PatchLinkSet", 2026-07-15 15:35:29.353119000 UTC +00:00],
 ["PatchLinkSet", 2026-07-15 15:35:29.140004000 UTC +00:00],
 ["Publish", 2026-07-15 15:35:06.420799000 UTC +00:00],
 ["PatchLinkSet", 2026-07-15 15:35:06.270862000 UTC +00:00],
 ["Publish", 2026-07-15 15:33:06.735484000 UTC +00:00],
 ["PatchLinkSet", 2026-07-15 15:33:06.518405000 UTC +00:00],
 ["Publish", 2026-07-13 13:02:27.263596000 UTC +00:00]]
```

## Remediation

Once these fixes are deployed we'll be able to manually trigger messages from Publishing API to get the 2 missing organisations correctly promoted in Content Data API:

```rb
%w[
  1390d07f-12b3-4f55-9dd5-fec5fd9e3649
  63953ed9-98ac-4d08-a1b9-8ea38dbd0832
].each do |content_id|
  Commands::V2::RepresentDownstream.new.call(content_id)
end
```
## Broader implications

These bugs will have affected all document types, not just organisations. It would be a big job to perform reconciliations for every document types between Content Data API and Publishing API, but as documents are republished, promotions which were previously blocked will take place. 

Republication, including periodic bulk republication for maintenance, is common, so this gradual approach seems adequate. As Content Data API is an analytics data warehouse, rather than a critical operational system, this form of event-driven self-healing over time seems like an appropriate overall remediation strategy.

[CR-53]: https://gov-uk.atlassian.net/browse/CR-53?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/dimensions/edition.rb
+++ b/app/models/dimensions/edition.rb
@@ -47,7 +47,7 @@ class Dimensions::Edition < ApplicationRecord
   end
 
   def unpublished?
-    %w[gone vanish redirect].include?(document_type)
+    %w[gone vanish redirect substitute].include?(document_type)
   end
 
   def change_from?(attributes)
--- a/app/models/dimensions/edition.rb
+++ b/app/models/dimensions/edition.rb
@@ -40,7 +40,11 @@ class Dimensions::Edition < ApplicationRecord
 
   def promote!(old_edition)
     old_edition.update!(live: false) if old_edition
-    update!(live: true) unless unpublished?
+
+    return if unpublished?
+
+    demote_any_live_editions_on_same_base_path
+    update!(live: true)
   rescue ActiveRecord::RecordNotUnique
     logger.info("Duplicate live edition detected for base_path: #{base_path}, skipping promotion.")
     raise ActiveRecord::Rollback
@@ -83,5 +87,14 @@ class Dimensions::Edition < ApplicationRecord
     return child_count unless child_count.zero?
 
     parent.try(:related_content_count) || 0
+  end
+
+private
+
+  def demote_any_live_editions_on_same_base_path
+    self.class
+        .where(base_path: base_path, live: true)
+        .where.not(id: id)
+        .update_all(live: false)
   end
 end
--- a/app/models/dimensions/edition.rb
+++ b/app/models/dimensions/edition.rb
@@ -45,9 +45,6 @@ class Dimensions::Edition < ApplicationRecord
 
     demote_any_live_editions_on_same_base_path
     update!(live: true)
-  rescue ActiveRecord::RecordNotUnique
-    logger.info("Duplicate live edition detected for base_path: #{base_path}, skipping promotion.")
-    raise ActiveRecord::Rollback
   end
 
   def unpublished?
--- a/lib/tasks/organisations.rake
+++ b/lib/tasks/organisations.rake
@@ -1,0 +1,40 @@
+namespace :organisations do
+  desc "Identify organisations missing from CDA and provide replay instructions"
+  task check_missing: :environment do
+    known_missing = %w[
+      1390d07f-12b3-4f55-9dd5-fec5fd9e3649
+      63953ed9-98ac-4d08-a1b9-8ea38dbd0832
+      99a5828c-79dd-4533-ab18-18069be289cf
+    ]
+
+    puts "Checking for missing organisations..."
+    puts
+
+    still_missing = known_missing.reject do |content_id|
+      Dimensions::Edition.exists?(content_id:, live: true)
+    end
+
+    no_op_message = "All organisations are present and live. Nothing to do."
+    next puts no_op_message if still_missing.empty?
+
+    puts "#{still_missing.size} organisation(s) still missing from CDA:"
+    puts
+    still_missing.each do |content_id|
+      edition = Dimensions::Edition.find_by(content_id:)
+      status = if edition
+                 "exists but live=#{edition.live}"
+               else
+                 "no record"
+               end
+      puts "  #{content_id} (#{status})"
+    end
+    puts
+    puts "To fix, run the following on Publishing API console:"
+    puts
+    puts "  %w[#{still_missing.join(' ')}].each do |content_id|"
+    puts "    Commands::V2::RepresentDownstream.new.call(content_id)"
+    puts "  end"
+    puts
+    puts "Then re-run this task to verify."
+  end
+end
--- a/spec/models/dimensions/edition_spec.rb
+++ b/spec/models/dimensions/edition_spec.rb
@@ -170,6 +170,18 @@ RSpec.describe Dimensions::Edition, type: :model do
         expect(new_edition.reload.live).to be true
       end
     end
+
+    context "when RecordNotUnique is raised unexpectedly" do
+      let(:edition) { create :edition, live: false }
+
+      it "lets the exception propagate for job-level retry" do
+        allow(edition).to receive(:update!)
+          .and_raise(ActiveRecord::RecordNotUnique)
+
+        expect { edition.promote!(nil) }
+          .to raise_error(ActiveRecord::RecordNotUnique)
+      end
+    end
   end
 
   describe "#change_from?" do
--- a/spec/models/dimensions/edition_spec.rb
+++ b/spec/models/dimensions/edition_spec.rb
@@ -140,6 +140,36 @@ RSpec.describe Dimensions::Edition, type: :model do
         expect(old_edition.live).to be false
       end
     end
+
+    context "when a different content_id occupies the same base_path" do
+      let(:base_path) { "/government/organisations/example-org" }
+
+      let!(:blocking_edition) do
+        create :edition,
+               base_path: base_path,
+               content_id: SecureRandom.uuid,
+               live: true
+      end
+
+      let(:new_edition) do
+        create :edition,
+               base_path: base_path,
+               content_id: SecureRandom.uuid,
+               live: false
+      end
+
+      it "demotes the blocking edition at the same base_path" do
+        new_edition.promote!(nil)
+
+        expect(blocking_edition.reload.live).to be false
+      end
+
+      it "promotes the new edition to live" do
+        new_edition.promote!(nil)
+
+        expect(new_edition.reload.live).to be true
+      end
+    end
   end
 
   describe "#change_from?" do
--- a/spec/models/dimensions/edition_spec.rb
+++ b/spec/models/dimensions/edition_spec.rb
@@ -124,6 +124,22 @@ RSpec.describe Dimensions::Edition, type: :model do
         expect(old_edition.live).to be false
       end
     end
+
+    context "for substitute edition" do
+      let(:edition) { build :edition, live: false, document_type: "substitute" }
+      let(:warehouse_item_id) { "warehouse-item-id" }
+      let(:old_edition) { build :edition, live: true, warehouse_item_id: }
+
+      it "does not set the live attribute to true" do
+        edition.promote!(old_edition)
+        expect(edition.live).to be false
+      end
+
+      it "sets the live attribute to false for the old version" do
+        edition.promote!(old_edition)
+        expect(old_edition.live).to be false
+      end
+    end
   end
 
   describe "#change_from?" do
--- a/spec/models/dimensions/edition_spec.rb
+++ b/spec/models/dimensions/edition_spec.rb
@@ -100,12 +100,12 @@ RSpec.describe Dimensions::Edition, type: :model do
 
       it "sets the live attribute to true" do
         edition.promote!(old_edition)
-        expect(edition.live).to be true
+        expect(edition.reload.live).to be true
       end
 
       it "sets the live attribute to false for the old version" do
         edition.promote!(old_edition)
-        expect(old_edition.live).to be false
+        expect(old_edition.reload.live).to be false
       end
     end
 
@@ -121,7 +121,7 @@ RSpec.describe Dimensions::Edition, type: :model do
 
       it "sets the live attribute to false for the old version" do
         edition.promote!(old_edition)
-        expect(old_edition.live).to be false
+        expect(old_edition.reload.live).to be false
       end
     end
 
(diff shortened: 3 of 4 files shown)


edavey:
@ChrisBAshton @callumknights Would you please sense-check this for us? (`git` is telling me you know a bit about `Edition#promote!`)

We discovered a lot about both Content Data and about Publishing API in the course of this investigation . We've tried to capture info along the way (in the commit messages) to help others in the future who are new to supporting this app and find themselves working in this area of "promotion/demotion". 

But maybe we've got things wrong, either on the macro or micro front!

spec/models/dimensions/edition_spec.rb
```diff
@@ -140,6 +140,36 @@
...
+        create :edition,
+               base_path: base_path,
+               content_id: SecureRandom.uuid,
+               live: false
+      end
+
+      it "demotes the blocking edition at the same base_path" do
+        new_edition.promote!(nil)
+
+        expect(blocking_edition.reload.live).to be false
```
graycodes, line 164: What is 'reload' doing for us here? Does it imply that other tests here that aren't using reload aren't checking valid values?

lib/tasks/organisations.rake
```diff
@@ -0,0 +1,41 @@
...
+
+    puts "Checking for missing organisations..."
+    puts
+
+    still_missing = known_missing.reject do |content_id|
+      Dimensions::Edition.exists?(content_id:, live: true)
+    end
+
+    if still_missing.empty?
+      puts "All organisations are present and live. Nothing to do."
```
graycodes: minor: Is it worth returning here so we can avoid having this following chunk of code all in an else?

graycodes (dismissed):

graycodes:
Just spotted that this issue is also recorded in https://github.com/alphagov/content-data-api/issues/2169 which may be worth a reference?

spec/models/dimensions/edition_spec.rb
edavey: [reload](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-reload) is reloading the record from the DB so we know that any changes made to the record since the instance variable `blocking_edition` was set are detected. Without this `reload` the test will fail, incorrectly, as the fact that the blocking edition has been demoted won't be visible.

I've added another commit which adds `reload` to other examples of the `#promote!` behaviour where we're asserting whether they've been promoted or demoted.

lib/tasks/organisations.rake
edavey: we can't `return` early or `break` from this block but we can `abort` e.g. with

```rb
no_op_message = "All organisations are present and live. Nothing to do."
abort no_op_message if still_missing.empty?
```

edavey: I've tweaked to use `abort` in this way.

callumknights (dismissed):
This all looks really good to me @edavey! I can only apologise for my past self not considering the content-data-api implications of creating the `substitute` unpublishing type, as it appears that is what's happened here. 🤦 Thanks for the fix!

lib/tasks/organisations.rake
edavey: sorry, this is wrong: `next` is right keyword here to bail out. If we use `abort` we exit with status 1 and the test suite fails...

app/models/dimensions/edition.rb
```diff
@@ -47,7 +47,7 @@ def promote!(old_edition)
   end
 
   def unpublished?
-    %w[gone vanish redirect].include?(document_type)
+    %w[gone vanish redirect substitute].include?(document_type)
```
patrickjfl, line 51: quite confusing that this state is called "substitute" when it refers to something that has _been substituted_

spec/models/dimensions/edition_spec.rb
```diff
@@ -124,6 +124,22 @@
         expect(old_edition.live).to be false
       end
     end
+
+    context "for substitute edition" do
+      let(:edition) { build :edition, live: false, document_type: "substitute" }
+      let(:warehouse_item_id) { "warehouse-item-id" }
+      let(:old_edition) { build :edition, warehouse_item_id: }
```
patrickjfl: i should probably know this but are we confident that `live` is not set here, because it's not being explicitly set?

spec/models/dimensions/edition_spec.rb
```diff
@@ -140,6 +140,36 @@
...
+      end
+
+      it "demotes the blocking edition at the same base_path" do
+        new_edition.promote!(nil)
+
+        expect(blocking_edition.reload.live).to be false
+      end
+
+      it "promotes the new edition to live" do
+        new_edition.promote!(nil)
```
patrickjfl, line 168: is it realistic to have the line `new_edition.promote!(nil)` in two separate test blocks? Maybe Rails is resetting the object between tests, but if not then it seems to me that we shouldn't be calling the method twice on the same object, when technically speaking it shouldn't be promoting the new edition to live the second time (because it already did the first time).

spec/models/dimensions/edition_spec.rb
```diff
@@ -100,12 +100,12 @@
 
       it "sets the live attribute to true" do
         edition.promote!(old_edition)
-        expect(edition.live).to be true
+        expect(edition.reload.live).to be true
```
patrickjfl, line 103: how does this work?

edavey:
> This all looks really good to me @edavey! I can only apologise for my past self not considering the content-data-api implications of creating the `substitute` unpublishing type, as it appears that is what's happened here. 🤦 Thanks for the fix!

thanks for the sanity-check @callumknights -- Once I saw the different definition of `unpublished` in Publishing API I figured the problem was probably all rooted in this "substitute" type... Hard to keep these systems in harmony over time!

app/models/dimensions/edition.rb
edavey: yes, my understanding is that there are five type of "unpublishing" event and that this "substitute" document type "is automatically created when an edition is substituted". 

When we look at the 5 types listed in https://github.com/alphagov/publishing-api/blob/7cea9433c32ff5a1de3fa6e4ee3b0744471972ba/docs/model.md?plain=1#L143-L170

- withdrawal (noun)
- redirect (imperative)
- gone (past participle)
- vanish (imperative)
- substitute (noun)

we see a mix of parts of speech. As these are "document types" perhaps a list of nouns would be clearest but that's not so easy...

- withdrawal
- redirection
- going
- vanishing
- substitution

or maybe the past participle is would be better?

- withdrawn
- redirected
- gone
- vanished
- substituted 

However, this is all a bit academic as these document types are defined in Publishing API and there's no mapping between document types received and represented internally. It's by design that the types received from Publishing API are the types used in the warehouse: https://github.com/alphagov/content-data-api/blob/7e179634bb8ee10ee9277e60fc87b8af1d8728a8/app/domain/streams/messages/base_message.rb#L20

spec/models/dimensions/edition_spec.rb
edavey: I'm not sure I understand the question exactly, but in these examples:

```rb
context "for substitute edition" do
  it "does not set the live attribute to true" do
  it "sets the live attribute to false for the old version" do
```
we start with 2 fixtures:

```rb
let(:edition) { build :edition, live: false, document_type: "substitute" }
let(:old_edition) { build :edition, warehouse_item_id: }
```
we describe how the `#promote!` method will:

- not promote a substitute
- demote a live incumbent

Perhaps you are asking about whether we can be confident in the starting state of `old_edition.live`? That's a fair point. We can see in the factory that `#live` defaults to true, but the test would be clearer with either an explicit setup:

```rb
let(:old_edition) { build :edition, live:true, warehouse_item_id: }
```
or a preliminary assertion before the action e.g:

```rb
it "sets the live attribute to false for the old version" do
  expect(old_edition.live).to be true
  edition.promote!(old_edition)
  expect(old_edition.reload.live).to be false
end
```

I'll make that change.

spec/models/dimensions/edition_spec.rb
edavey: I don't think this is an issue. Our context is reset between examples. These are different objects. You can use `Object#object_id` to see whether an object is different or not e.g. I see:

```
Dimensions::Edition
  #promote!
    when a different content_id occupies the same base_path
1st 9088
      demotes the blocking edition at the same base_path
2nd 9152
      promotes the new edition to live
```

spec/models/dimensions/edition_spec.rb
edavey: are you asking i) how `reload` works or ii) how `#live` comes to be `true`?

For i) see https://github.com/alphagov/content-data-api/pull/2546#discussion_r3656312961

For ii) this is what `#promote!(old_edition)` does:

```rb
  def promote!(old_edition)
    old_edition.update!(live: false) if old_edition

    return if unpublished?

    demote_any_live_editions_on_same_base_path
    update!(live: true)
  end
```
on the final line the caller (`edition` in the test) sets itself live, provided it isn't an "unpublished" type (and after having demoted any live editions occupying its base path)

edavey:
@patrickjfl I think I've responded to your questions / comments. Could I please get a fresh review/approval?

spec/models/dimensions/edition_spec.rb
patrickjfl: sorry i should have been clearer, i was asking how `reload` works, thanks. so basically it's making sure the change made it all the way to the DB value?

spec/models/dimensions/edition_spec.rb
patrickjfl: i see, thanks for explaining that. In JS this would be more complicated!

app/models/dimensions/edition.rb
patrickjfl: yes, i think the approach is fine, just a musing really on what Publishing API is serving up...

patrickjfl (approved):
Great work Ed.

